PR #106, opened 2026-04-16
fix(api): standardize remaining error response codes

march26bot:
## Summary\n- standardize the remaining plain API error responses so they always include structured  and  fields\n- assign explicit codes for bad JSON/request-size/multipart/missing-file/no-IP error paths\n- extend API tests to cover the new error-code behavior and update the roadmap note for item 1.2.5\n\n## Testing\n- Could not run locally in this workspace because / are not installed ( and  returned nothing)\n- CI should validate  once the PR is opened\n


chinmay28 (approved):
CI passes, clean change. This does a thorough job standardizing the remaining plain `writeError` calls into structured `code` + `message` responses.

Good decisions:
- **Backwards compatible**: keeps the existing `error` field while adding `code` and `message`, so existing clients don't break
- **`defaultErrorCode` fallback**: any `writeError` callers not yet migrated to `writeErrorCode` still get a sensible code derived from the HTTP status
- **Test coverage**: adds `assertAPIErrorCode` calls to all the bad-JSON/body-too-large paths, plus a new `TestAddPort_VMIPUnavailable` test for the no-IP edge case
- **Explicit codes** are well-chosen: `invalid_request_body`, `request_too_large`, `missing_file`, `vm_ip_unavailable`, `invalid_multipart_form`, etc.

LGTM.

---

PR #126, opened 2026-04-23
test(web): cover VM clone modal flow

march26bot:
## Summary
- add stable test IDs for the VM clone action and modal controls
- reset clone modal errors on close/reopen so stale failures do not persist
- extend the mock web server and Playwright GUI coverage to exercise the clone flow end-to-end

## Validation
- TAP version 13
# Subtest: request adds Authorization header when API key is set
ok 1 - request adds Authorization header when API key is set
  ---
  duration_ms: 51.849718
  type: 'test'
  ...
# Subtest: 401 without token marks auth as required
ok 2 - 401 without token marks auth as required
  ---
  duration_ms: 1.0523
  type: 'test'
  ...
# Subtest: 401 with invalid token clears stored token and keeps auth prompt open
ok 3 - 401 with invalid token clears stored token and keeps auth prompt open
  ---
  duration_ms: 0.770267
  type: 'test'
  ...
# Subtest: image upload reports progress and sends auth header
ok 4 - image upload reports progress and sends auth header
  ---
  duration_ms: 4.005702
  type: 'test'
  ...
# Subtest: vms.list sends pagination params and returns total count metadata
ok 5 - vms.list sends pagination params and returns total count metadata
  ---
  duration_ms: 43.04527
  type: 'test'
  ...
# Subtest: images.list sends pagination params and returns total count metadata
ok 6 - images.list sends pagination params and returns total count metadata
  ---
  duration_ms: 0.465752
  type: 'test'
  ...
# Subtest: logs.list sends pagination params and returns total count metadata
ok 7 - logs.list sends pagination params and returns total count metadata
  ---
  duration_ms: 0.644102
  type: 'test'
  ...
# Subtest: vms.clone posts the requested clone name
ok 8 - vms.clone posts the requested clone name
  ---
  duration_ms: 0.373865
  type: 'test'
  ...
1..8
# tests 8
# suites 0
# pass 8
# fail 0
# cancelled 0
# skipped 0
# todo 0
# duration_ms 164.930537
-  (from )
- GUI test runner requires , which is not installed in this maintenance clone, so  is blocked locally despite the new mock-server coverage being included

PR #141, opened 2026-04-28
fix(web): keep log viewer pinned where the user scrolled

chinmay28:
## Summary

The Logs tab snapped to the bottom on every 3-second poll, even when the user had scrolled up. Two issues:

- `bottomRef.current.scrollIntoView({ behavior: 'smooth' })` fired intermediate scroll events during the animation, which the `onScroll` handler interpreted, racing with the user's input. `scrollIntoView` also scrolls every scrollable ancestor, not just the log container.
- Polling reassigned `entries` to a new array every cycle, retriggering the auto-scroll effect even when the user had scrolled away.

Fix in `web/src/pages/LogViewer.jsx`:

- Replace `scrollIntoView` with direct `listRef.current.scrollTop = listRef.current.scrollHeight` — instant, single scroll event, scoped to the log container.
- `handleScroll` only calls `setAutoScroll` when the bottom threshold is actually crossed, reading the current value via a ref to avoid stale-closure flips.
- The "Latest" button now uses the same direct scrollTop approach.

Net effect: when you scroll up, autoScroll flips off on that one event and stays off through subsequent polls until you scroll back to within 40px of the bottom.

## Test plan

- [x] `npx vite build` passes locally
- [ ] Manual: open the Logs tab, scroll up — the view stays put through poll cycles; click "Latest" or scroll near the bottom and the view re-pins
- [ ] CI: backend lint/build/tests, frontend build + mock GUI tests

https://claude.ai/code/session_01VeBiUDvZ8dcdWFpBQu6jkX

---
_Generated by [Claude Code](https://claude.ai/code/session_01VeBiUDvZ8dcdWFpBQu6jkX)_

### DIFF
--- a/web/src/pages/LogViewer.jsx
+++ b/web/src/pages/LogViewer.jsx
@@ -35,8 +35,12 @@ export default function LogViewer() {
   const [perPage, setPerPage] = useState(DEFAULT_PER_PAGE);
   const [autoScroll, setAutoScroll] = useState(true);
   const [paused, setPaused] = useState(false);
-  const bottomRef = useRef(null);
   const listRef = useRef(null);
+  const autoScrollRef = useRef(true);
+
+  useEffect(() => {
+    autoScrollRef.current = autoScroll;
+  }, [autoScroll]);
 
   const fetchLogs = useCallback(async () => {
     if (paused) return;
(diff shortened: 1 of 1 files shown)


chinmay28 (commented):
Reviewed locally — looks good. (Can't APPROVE because I authored the PR; leaving a COMMENT review instead.)

Verified:

- The branch is behind main (`#126` added Log Viewer testids + memos after this branch was cut), but a 3-way merge is clean — `useMemo`, `firstSourceIndex/firstLevelIndex`, and all `data-testid` attributes survive on the merge result. No conflicts.
- After merging main into the branch, `npx vite build` succeeds.
- Web GUI tests (`run-gui-tests.js` 17/17 + `run-built-web-regression.js` 4/4) all pass against the rebuilt bundle.

The fix itself is the right shape:

- Replacing `scrollIntoView({behavior:'smooth'})` with a direct `el.scrollTop = el.scrollHeight` keeps the auto-scroll scoped to the log container and avoids the intermediate scroll events that were racing with the user.
- Guarding `setAutoScroll(atBottom)` with the `autoScrollRef` ref means state only flips when the threshold is actually crossed, eliminating the stale-closure flip-flop on every poll.
- The "Latest" button uses the same direct path for consistency.

Recommend rebasing/merging main before merging so CI runs against the combined tree.

---

PR #140, opened 2026-04-28
feat(daemon): add automatic TLS certificates

march26bot:
## Summary
- add `daemon.tls.auto_cert` config for built-in Let's Encrypt certificate management
- serve HTTPS with `autocert` TLS-ALPN, host allowlisting, cache/email settings, and daemon tests
- document direct `:443` built-in TLS usage in the README, example config, and deployment roadmap

## Testing
- not run locally in this environment (Go toolchain unavailable)

PR #143, opened 2026-04-28
feat(api): serve Swagger UI docs

march26bot:
## Summary
- serve interactive Swagger UI at `/api/docs`
- serve the embedded OpenAPI schema at `/api/openapi.yaml`
- add API tests and docs references for the new docs routes

## Testing
- not run locally in this environment because the Go toolchain is unavailable (`go: not found`)
- reviewed the route wiring and added handler-level tests for `/api/docs` and `/api/openapi.yaml`


chinmay28 (changes requested):
## Review summary

Requesting changes — the implementation works and tests pass, but it diverges from the roadmap intent in a way that has security and operational consequences.

### Verified locally

- `go build ./...` succeeds.
- `go vet ./...` clean.
- New tests pass:
  ```
  === RUN   TestSwaggerUIRouteServesHTML
  --- PASS: TestSwaggerUIRouteServesHTML (0.01s)
  === RUN   TestOpenAPISpecRouteServesYAML
  --- PASS: TestOpenAPISpecRouteServesYAML (0.01s)
  ```
- The two unrelated unit-test failures (`TestUploadImage_MissingName`, `TestCLI_TemplateCreate_RejectsInvalidInput/duplicate_name`) reproduce on `main` and are not introduced by this PR.

### Blocker: roadmap says "Embed swagger-ui dist" but this PR pulls from a public CDN

`docs/ROADMAP.md` 4.3.2 reads:

> Serve Swagger UI at `/api/docs` | S | Embed swagger-ui dist

The implementation in `docs/apidocs.go` instead loads `swagger-ui.css` and `swagger-ui-bundle.js` from `cdn.jsdelivr.net`. That choice has real consequences for a self-hosted VM management daemon:

1. **Air-gapped / offline deployments.** VMSmith is shipped as `.deb`/`.rpm`/static binary (`make dist`, `make rpm`, `make deb`). It needs to work in environments without outbound internet. Today, opening `/api/docs` on such a host produces a blank/broken page.
2. **Supply-chain risk.** No SRI (`integrity=`) attributes are set on the `<link>` or `<script>` tags. Any compromise or hijack of jsdelivr causes arbitrary JavaScript to run inside an authenticated admin's browser context, with `localStorage` access to `vmsmith.apiKey`. Given that this UI is intended to be hit by operators who likely have their API key in localStorage, this is not a hypothetical risk.
3. **Privacy / network egress.** Every operator hit on `/api/docs` makes a third-party request, which is a surprise for a self-hosted tool.

The fix matches the original task: bundle the swagger-ui-dist `swagger-ui.css` + `swagger-ui-bundle.js` (+ `swagger-ui-standalone-preset.js` if you use it) under `docs/swagger-ui/` and serve them via the same `embed.FS`. Reference them with relative paths from the inline HTML. swagger-ui-dist is ~1.5 MB which is acceptable for the binary; if it's a concern, `go:embed` only the two files you actually need.

### Smaller things

1. **Package location.** Putting Go code at `docs/apidocs.go` is unusual — `docs/` has been documentation-only and `CLAUDE.md` still describes it that way. The `go:embed openapi.yaml` constraint forces co-location, but consider moving the package to `internal/apidocs/` and either embedding `../../docs/openapi.yaml` via a thin asset shim, or moving the spec under the package. Either is cleaner than introducing executable code in `docs/`.
2. **`html/template` is overkill.** `docsHTML` has no template fields. A `[]byte` constant served via `w.Write` is simpler and avoids the silently-discarded `_ = docsHTML.Execute(w, nil)` error.
3. **Auth/rate-limit bypass.** `/api/docs` and `/api/openapi.yaml` are registered on the root chi router, outside `/api/v1`, so they bypass `apiKeyAuth` and the per-IP rate limiter. Public docs are fine, but exposing the OpenAPI spec on an authenticated daemon hands attackers a complete API map for free. Consider gating both routes behind `apiKeyAuth` when `auth.enabled` is true (the SwaggerUI fetch can include the bearer token from localStorage).
4. **`http.ServeFileFS` content type.** You set `Content-Type: application/yaml` and then call `http.ServeFileFS`, which preserves the explicit header — fine, but worth a one-line comment to head off "why is this set twice?" questions later.
5. **README/CLAUDE.md.** README is updated, but `CLAUDE.md`'s REST API quick reference doesn't mention `/api/docs` or `/api/openapi.yaml`. Worth a small addition there for consistency with how other endpoints are documented.

Happy to re-review once the swagger-ui assets are embedded.


---

chinmay28 (approved):
Roadmap 4.3.2 done cleanly. Verified locally:

- `CGO_ENABLED=1 go build ./...` succeeds
- `go test ./internal/api/... -run "TestSwaggerUI|TestOpenAPISpec"` — all 3 new tests pass (`TestSwaggerUIRouteServesHTML`, `TestSwaggerUIAssetRouteServesEmbeddedBundle`, `TestOpenAPISpecRouteServesYAML`). The `/api/docs` HTML correctly references the embedded asset URLs and `/api/openapi.yaml` returns the OpenAPI 3 spec with the right content type.
- The `docs/swagger-ui` bundle is checked in and embedded (1.5MB JS + 179KB CSS) — that's the right call vs. a CDN, since the daemon should remain self-contained for air-gapped deployments.
- Routes are registered outside the `/api/v1` subrouter so they (correctly) bypass the bearer-auth middleware — Swagger browsability without an API key is the expected UX.
- The `AssetHandler` traversal guard (`strings.Contains(name, "..")` + `path.Clean`) is defensive enough for serving a fixed embed.FS.

> Heads up: I noticed `TestUploadImage_MissingName` fails on this branch, but it also fails on `origin/main` — pre-existing, not introduced here. Worth a separate fix but doesn't block this PR.

## Minor observations (non-blocking)

1. **Package location.** `docs/apidocs.go` puts Go source under a top-level `docs/` dir, which conflicts with the CLAUDE.md guideline that says "`internal/` — all implementation code". The reason is that `//go:embed` can't reach across directories with `..`, so co-locating the embed source with the spec is the practical choice. Consider either (a) moving the package to `internal/apidocs/` and copying `openapi.yaml` + `swagger-ui/` next to it as part of `make build`, or (b) leaving it as-is and adding a one-liner to CLAUDE.md noting that `docs/` also hosts an embedded Go package for the API explorer. Either is fine — just want the convention to be explicit.

2. **PR body says "not run locally — Go toolchain unavailable."** This is the second PR I've reviewed today with that note. If CI runs on these PRs, `go build` would catch issues like the `acme` import bug in #140 before review; if it doesn't, that's the gap to close. For this PR specifically, build + tests pass — but please confirm CI is green before merging.

3. **`docs/openapi.yaml` drift risk.** The spec is hand-written and will silently drift from real handler behavior. Worth adding a contract test (later, not in this PR) that boots the test server and checks that every documented path has a route registered, and vice versa. Tracking this in 4.3 follow-ups would be a good idea.

Approving — ship it once you're satisfied with CI.

---

PR #145, opened 2026-04-29
Add VM metrics collection and event bus infrastructure

chinmay28:
## Summary
This PR introduces comprehensive VM resource metrics collection and an event bus for system-wide event distribution. The metrics system polls libvirt's bulk stats API at configurable intervals and stores samples in per-VM ring buffers, while the event bus provides a fan-out mechanism for VM lifecycle and system events with persistent storage.

### Metrics Collection (`internal/metrics/`)
- **`metrics.go`**: Core `Manager` interface and `MockMetricsManager` for testing
- **`collector.go`**: `LibvirtMetricsManager` implementation using libvirt's bulk stats API
  - Polls CPU, memory, disk I/O, and network metrics at configurable intervals
  - Computes per-second rates from cumulative libvirt counters
  - Maintains per-VM in-memory ring buffers with configurable history size
  - Handles counter resets and missing data gracefully
- **`ring.go`**: Fixed-size circular buffer for storing metric samples
- **`metrics_test.go`**: Comprehensive test suite covering ring buffer behavior, mock manager, and API integration

### Event Bus (`internal/events/`)
- **`bus.go`**: `EventBus` implementation with monotonic ID assignment
  - Single background goroutine serializes ID assignment and persistence
  - Fan-out to multiple subscribers with per-subscriber buffering
  - Non-blocking publish with drop warnings for slow subscribers
  - Integrates with store for persistent event storage
- **`bus_test.go`**: Tests for publish/subscribe, multiple subscribers, and ID monotonicity

### API Enhancements (`internal/api/`)
- **`handlers_metrics.go`**: New `GET /api/v1/vms/{vmID}/stats` endpoint
  - Returns current metrics snapshot plus historical samples
  - Supports `?since=<rfc3339>` filtering for time-range queries
  - Supports `?fields=cpu,mem,disk,net` projection to reduce payload
  - Returns 503 when metrics disabled, 404 for unknown VMs
- **`handlers_events.go`**: Enhanced `GET /api/v1/events` with filtering
  - Added filters: `type`, `source`, `severity`, `until` (seq ID)
  - Pagination support via `page` and `per_page` parameters
  - Sorts events newest-first by `occurred_at` timestamp
- **`sse.go`**: Server-Sent Events writer for streaming responses
- **`router.go`**: Updated to accept `MetricsManager` interface (avoids transitive libvirt CGO dependency)

### CLI Enhancements (`internal/cli/`)
- **`vm.go`**: New `vmsmith vm stats <id>` command
  - Displays current and 5-minute average metrics in tabular format
  - `--watch` flag for continuous polling with screen refresh
  - `--fields` flag to filter metric output
  - Calls daemon REST API rather than direct VM manager access

### Storage & Persistence (`internal/store/`)
- **`bolt.go`**: New event persistence methods
  - `AppendEvent()`: Assigns monotonic uint64 sequence numbers via bbolt's `NextSequence()`
  - `ListEventsFiltered()`: Efficient reverse-order cursor scan with filtering and pagination
  - Supports both legacy string-keyed events and new sequence-keyed events

### Configuration (`internal/config/`)
- Added `MetricsConfig` with fields:
  - `enabled`: Toggle metrics collection
  - `sample_interval`: Polling interval in seconds (default 10)
  - `history_size`: Ring buffer capacity (default 360 = 1 hour at 10s intervals)

### Daemon Integration (`internal/daemon/`)
- Opens dedicated libvirt connection for metrics sampler (avoids contention with VM manager)
- Initializes `LibvirtMetricsManager` when metrics enabled
- Initializes `EventBus` and wires it into API server
- Gracefully handles libvirt connection failures

### Type Definitions (`pkg/types/`)
- **`metrics.go`**: New `MetricSample` and `VMStatsSnapshot` types
  - `MetricSample`: CPU%, memory,

https://claude.ai/code/session_0183q1dYyR5R6Wzz1tdAraYz


chinmay28 (commented):
Big PR — landing 4.1 (metrics) and a real chunk of 4.2 (event bus + persistence + SSE) together is ambitious, and the architecture mostly mirrors the roadmap (single-goroutine bus, ring buffer per VM, big-endian uint64 keys, drop-not-block fan-out). I built the branch (`CGO_ENABLED=1 go build ./...` clean) and ran the package tests:

- `go test ./internal/metrics/...` — pass
- `go test ./internal/events/...` — pass
- `go test ./internal/api/...` — passes overall, except `TestListVMs_PaginationSetsTotalCount` is **flaky** (fails ~30% of runs with `page 2 VM = "gamma", want beta`). I confirmed this is **pre-existing on `origin/main`**, not introduced here, but it should be fixed independently.

Submitting this as a comment review (REQUEST_CHANGES is blocked because GitHub doesn't allow it on your own PR), but treat the four blockers below as must-fix before merge.

## Blocker 1 — `LibvirtMetricsManager` rings are keyed by VM **name**; the API/CLI query by VM **ID**

`internal/metrics/collector.go` (the body of `sample()`): the sampler stores rings under `domName` (libvirt domain name), and `vmsmith` always sets the libvirt domain name to `spec.Name`, not the VM ID (`internal/vm/lifecycle.go:355` — `m.conn.LookupDomainByName(vm.Name)`).

```go
// in sample()
domName, _ := ds.Domain.GetName()           // returns "my-vm"
m.rings[domName] = newRing(m.histSize)      // keyed by "my-vm"
```

But the API and CLI both pass `vmID` (`vm-1741234567890123`):

```go
// handlers_metrics.go
snap, err := s.metricsManager.Snapshot(vmID)   // looks up rings["vm-174..."]

// internal/cli/vm.go
url := fmt.Sprintf("%s/api/v1/vms/%s/stats", ..., vmID)
```

So `m.rings[vmID]` always returns `(nil, false)`. The handler hides this by synthesizing an empty `VMStatsSnapshot` from the VM manager, so users get HTTP 200 with `current: null` and `history: []` regardless of how long the daemon has been running. The CLI's `--watch` mode will redraw an empty table forever. Same for the Prometheus exporter — `s.metricsManager.Snapshot(v.ID)` always returns nil, so `/metrics` will only ever emit the HELP/TYPE lines and no actual values.

The collector has a `vmNames map[string]string  // libvirt domain name -> vmsmith VM ID` field declared (and deleted in pruning), but it is **never written to**. Two ways to fix:

1. **Translate domName → vmID at sample time** (preferred). Inject a small lookup interface (`VMNameToID(name string) (string, bool)` backed by `store.ListVMs` cached for ~30s) and key rings by VM ID. Populate `vmNames` so the prune loop continues to work.
2. **Translate vmID → name at query time.** Have `Snapshot(vmID)` look up the VM's name via the store, then index by name. Slightly racier on rename, but smaller diff.

Either fix needs unit coverage that exercises the `LibvirtMetricsManager` itself (see Blocker 4).

## Blocker 2 — `EventBus.Subscribe` silently overwrites duplicate names; SSE handler picks a non-unique name

`internal/api/handlers_events.go:184`:
```go
ch, cancel := s.eventBus.Subscribe("sse-" + r.RemoteAddr)
```

…and `internal/events/bus.go:100`:
```go
b.subscribers[name] = sub  // overwrites whatever was there, leaking the old channel
```

Two browser tabs behind the same NAT/proxy IP will collide on `r.RemoteAddr`. The first tab's channel is silently leaked (still receiving events nobody reads, eventually filling and triggering `subscriber_lagged`), and the first tab stops receiving events. Either:
- Make `Subscribe` return an error on duplicate name (callers must pick unique names — UUID), or
- Always append a UUID suffix in `Subscribe`, ignoring caller-supplied collisions.

I'd go with the second; the caller never cares about the exact subscriber name.

## Blocker 3 — `ListEvents` ignores the new `ListEventsFiltered`

The roadmap (4.2.5) added `Store.ListEventsFiltered` precisely so we'd do reverse-cursor scans with filter pushdown instead of loading everything into memory and slicing in Go. But `internal/api/handlers_events.go:57` still calls `s.store.ListEvents()` (the legacy method), then sorts and filters in-process:

```go
allEvents, err := s.store.ListEvents()   // loads everything
...
filtered := make([]*types.Event, 0, len(allEvents))
for _, e := range allEvents { ... }       // O(N) in Go
sort.Slice(filtered, ...)                 // re-sort
```

Once `daemon.events.max_records` defaults to 50,000, every request to `/api/v1/events?vm_id=foo` allocates 50k+ pointers and a sort. Switch to `ListEventsFiltered(EventFilter{...})` and let Bolt do the work. The added benefit is that `until=<seq>` and `since=<seq>` then map directly to cursor bounds instead of post-filtering by string ID.

## Blocker 4 — No real-collector unit tests for rate math / counter-reset / stale prune

`internal/metrics/metrics_test.go` only covers the ring buffer and the mock manager. The roadmap's 4.1.2 explicitly called for "Mock provider feeds synthetic counter sequences (normal, reset, jump, stale) to validate rate math". `LibvirtMetricsManager.sample()` has the most error-prone code in this PR — counter-reset detection, partial-data branches, the "stamp `prev.sampleTime` after deltas" ordering — and it has zero test coverage. Combined with Blocker 1, it's exactly why the keying bug slipped through.

Please add a small fake `domainStatsProvider` interface (just enough to feed synthetic `libvirt.DomainStats` slices) so the rate-math, reset, and prune behavior can be unit-tested without a real libvirt connection.

## Smaller issues (not blocking)

5. **SSE auth gap.** `/api/v1/events/stream` lives inside the `apiKeyAuth` middleware group, but browser `EventSource` cannot send custom `Authorization` headers — the roadmap (4.2.10) called for an `?api_key=` query fallback. As-is, the embedded GUI's planned `useEventStream` hook will get 401 the moment auth is enabled. Either add the `?api_key=` shortcut now or scope it as a follow-up — but call it out in the PR description so it's not forgotten.

6. **Prometheus metric labels — escape `vm.Name`.** `handlers_metrics.go:185` uses `fmt.Sprintf("vm_id=%q,vm_name=%q", v.ID, v.Name)` — Go's `%q` happens to match Prometheus label syntax for typical names, but a name containing a backslash or non-ASCII character can produce a value the Prometheus text format won't accept. VM names today are validated at create time per the roadmap, so this is mostly defensive; either rely on that and drop the concern, or implement Prometheus's escape rules (`\\`, `\"`, `\n`) to be safe against future relaxation.

7. **Unused `vmNames` map.** Once Blocker 1 is fixed, populate it; if not used, remove it. Don't ship dead state.

8. **`TestListVMs_PaginationSetsTotalCount` flake** (pre-existing, not from this PR, just flagging again). The MockManager apparently returns VMs in non-deterministic order; pagination tests need a stable sort.

Once Blockers 1–4 are addressed and 5–6 are at least decided on, this is approve-worthy. The architecture is sound; it just needs the wiring to match.

---

chinmay28:
Pushed `5e2fe37` addressing the review.

**Blockers**

| # | Status | Notes |
|---|---|---|
| 1 — ring keying by name vs ID | ✅ | Added `metrics.NameToIDFunc` resolver; daemon supplies a store-backed impl (`internal/daemon/metrics_resolver.go`) with a 30s name → ID cache. Rings are now keyed by VM ID; `vmNames` is populated and used in stale-prune. |
| 2 — SSE Subscribe duplicate name | ✅ | `EventBus.Subscribe` now always appends a monotonic `#N` suffix internally. Two callers passing the same supplied name no longer collide. New regression test: `TestBusSubscribeDuplicateNamesDoNotCollide`. |
| 3 — `ListEvents` ignores `ListEventsFiltered` | ✅ | Switched the handler to `store.ListEventsFiltered` (cursor scan + filter pushdown). Also fixed the dead-code `time.Time Since` branch in `ListEventsFiltered`. |
| 4 — no real-collector unit tests | ✅ | Refactored the sampler behind a `domainStatsProvider` interface. New `collector_test.go` covers: VM-ID keying, unresolvable-name skip, CPU rate, counter-reset → nil, recovery after reset, network rate, balloon memory, stale prune at the 5-min boundary, partial-data (no prev) and CPU clamp. |

**Smaller items**

| # | Status | Notes |
|---|---|---|
| 5 — SSE auth via `?api_key=` | ✅ | `apiKeyAuth` now accepts the key via `?api_key=` query as a fallback to the `Authorization` header. New tests cover allow + reject. |
| 6 — Prometheus label escape | ✅ | Added `escapePromLabel` (`\\`, `\"`, `\n`) and dropped `%q`. Unit-tested. |
| 7 — unused `vmNames` map | ✅ | Now populated by the resolver path and cleaned up alongside the ring during prune. |
| 8 — `TestListVMs_PaginationSetsTotalCount` flake | ✅ | `ListVMs` sorts by ID before paginating; verified with `-race -count=10`. |

CI is green locally (`go test -race -count=1 ./...`). Waiting on GitHub Actions.

---

chinmay28 (commented):
Re-reviewing after `5e2fe37`. The four prior blockers and four smaller items are all addressed; verified locally against current `main`.

## Re-verification of prior blockers

| # | Item | Status | Evidence |
|---|---|---|---|
| 1 | Rings keyed by VM name vs ID | ✅ Fixed | `collector.go:32` introduces `NameToIDFunc`; `processSamples` (`collector.go:225-247`) uses `resolveVMID()` with cache → resolver → fallback; `vmNames` is now populated and used during prune. Daemon wires a store-backed resolver via `internal/daemon/metrics_resolver.go`. |
| 2 | Subscribe collision on duplicate name | ✅ Fixed | `bus.go:37` adds `subSeq atomic.Uint64`; `Subscribe` always appends a unique `#N` suffix (`bus.go:99-121`). Regression test `TestBusSubscribeDuplicateNamesDoNotCollide` is in place at `bus_test.go:173`. |
| 3 | `ListEvents` ignored `ListEventsFiltered` | ✅ Fixed | `handlers_events.go:70` now calls `s.store.ListEventsFiltered(filter)` with the full `EventFilter` (vm_id / type / source / severity / since / until / page). |
| 4 | No real-collector unit tests | ✅ Fixed | `collector_test.go` (281 lines) covers VM-ID keying, unresolvable-name skip, CPU rate, counter-reset → nil, recovery, network rate, balloon memory, stale prune at the 5-min boundary, partial data, and CPU clamp. |
| 5 | SSE auth via `?api_key=` | ✅ Fixed | `auth_middleware.go:45-51` falls back to `?api_key=` when no `Authorization: Bearer` header is present. Tests at `api_test.go:2969-2996`. |
| 6 | Prometheus label escape | ✅ Fixed | `escapePromLabel()` in `handlers_metrics.go:146` handles `\\`, `"`, `\n` per the text-format spec. |
| 7 | Unused `vmNames` map | ✅ Fixed | Now populated by `processSamples` and pruned alongside rings in `prune()`. |
| 8 | `TestListVMs_PaginationSetsTotalCount` flake | ✅ Fixed | `handlers_vm.go:234` adds an explicit `sort.Slice(vms, ...)` by ID before paginating. Verified non-flaky with `go test -count=10 -race -run TestListVMs_Pagination ./internal/api/...`. |

## Local verification
- Resolved a clean 3-way merge with current `main` (49e010f) — no manual conflict resolution needed despite the `client.ts`/`schema.d.ts` deletions visible in `git diff main..pr-145`; those are artefacts of the older base, not real conflicts.
- `go build ./...` clean.
- `go test -count=1 -race ./internal/metrics/... ./internal/events/... ./internal/api/...` — all green (`metrics: ok 1.031s`, `events: ok 1.070s`, `api: ok 2.516s`).
- `go test -count=1 -short ./...` — all packages green; the `TestUploadImage_MissingName` failure listed in the PR description as pre-existing is now also fixed by #148 in flight, and `TestCLI_TemplateCreate_RejectsInvalidInput/duplicate_name` is unrelated to this PR.

## Non-blocking observations for follow-up

- **`sseReplayLimit` is hardcoded** at `handlers_events.go:93` (`const sseReplayLimit = 1000`). The roadmap (4.2.10) specifies `daemon.events.sse_replay_limit` as a config knob with default 1000. Easy to add later, but worth a TODO or a follow-up issue so it doesn't get lost.
- **Event retention is implemented but never invoked.** `store.PruneEvents(maxRecords)` exists at `bolt.go:424`, but no caller wires it up — events will accumulate in bbolt indefinitely. The roadmap (4.2.7) covers this; if it's planned for a follow-up PR, a comment or `TODO(4.2.7)` next to `PruneEvents` would help.
- **`/metrics` is intentionally exposed without auth** when `metrics.enabled` and no `scrape_listen` is configured. Standard Prometheus convention, but on a multi-tenant host it leaks per-VM names. Acceptable default; just confirm this is the intended trust boundary in the README/config docs (`vmsmith.yaml.example` already mentions `scrape_listen` for separating it).
- **Daemon shutdown ordering looks correct** — `closeResources()` stops the bus first (drains pending events), then metrics manager, then VM manager / network / store. One thing worth a comment in the code: `eventBus.Stop()` blocks on the publish loop draining via `<-b.stopped`, but if a publisher's persist call hangs (libvirt/bbolt fsync), `Stop()` will block indefinitely — consider whether `Stop(ctx)` should accept a deadline, given the daemon's 10s shutdown timeout.

## Summary
This is a substantial, architecturally clean slice of phase 4. The interface boundaries (MetricsManager local interface to avoid CGO leak, EventBus.Store minimal persistence interface) are well-chosen. Prior review feedback is comprehensively addressed and the new tests are good. The three observations above are real but non-blocking; they're appropriate follow-ups rather than reasons to hold this back.

Treat as **approve** — happy to merge once GitHub Actions reports green on the merge result. Submitting as a comment review since GitHub blocks self-approve.

---

PR #146, opened 2026-04-29
feat(web): generate frontend API client from OpenAPI

march26bot:
## Summary
- generate frontend API path types from `docs/openapi.yaml` with `openapi-typescript`
- replace the handwritten REST wrapper with an `openapi-fetch`-backed `web/src/api/client.ts` while preserving auth handling, upload progress, and pagination metadata
- document the regeneration workflow and mark roadmap item 4.3.3 complete

## Testing
- `npm run build --prefix web`
- `make test-web`


chinmay28 (approved):
Roadmap 4.3.3 done. Verified locally:

- `cd web && npm install && npm run build` — clean build, output bundle sizes are unchanged from main (`index-*.js` 253.77 kB → 76 kB gzipped vs the old hand-written client; the openapi-fetch runtime is small enough that there's no bundle-size regression worth flagging).
- `npm run generate:api` regenerates `web/src/api/generated/schema.d.ts` with **zero diff** vs the checked-in copy — so the regeneration script is idempotent and the spec/source pair is consistent.
- API surface parity vs the old `client.js`: every method on `vms`, `snapshots`, `images`, `ports`, `templates`, `host`, `quotas`, and `logs` is preserved, including `withMeta` for paginated lists (`vms.list`, `images.list`, `templates.list`, `logs.list`) and the `XMLHttpRequest`-based `images.upload` with progress reporting (which intentionally bypasses `openapi-fetch` — fetch streaming uploads aren't widely supported, so this is the right call).
- 401 handling, `clearAuthToken` + `requireAuth` flow, and `Authorization: Bearer` injection via the `apiClient.use({ onRequest })` middleware are all preserved.
- Web E2E tests didn't run in this environment (Playwright wants chromium 1217 and only 1194 is available locally — env issue, not the PR's fault). I'd want green CI here before merge.

Approving — solid, focused implementation. A few notes inline below; treat them as follow-up suggestions, not blockers.

## Suggestions (non-blocking)

1. **`unwrap` returns `any` — half the openapi-fetch value goes unused.** The function signature is `unwrap(result: Promise<any>, ...)` returning `any`, which means callers get *typed request bodies* (good — `vms.create(spec: paths['/vms']['post']['requestBody']['content']['application/json'])`) but **untyped responses**. The big win of openapi-fetch is end-to-end inference; the response side currently throws it away. Consider:
   ```ts
   async function unwrap<T>(
     result: Promise<{ data?: T; error?: unknown; response: Response }>,
     options: { withMeta?: boolean } = {},
   ): Promise<T | { data: T; meta: { totalCount: number } } | null> { ... }
   ```
   …or split into `unwrap<T>` and `unwrapWithMeta<T>` so callers don't have to disambiguate the union. This unlocks IDE autocomplete on response shapes for every consumer (`vms.get(id).then(vm => vm.???)`).

2. **`client.js` re-export trick.** The new `client.js` is just `export * from './client.ts'; export { default } from './client.ts';`. That works, but Vite's default resolution will happily find `./client.ts` if the consumers' import is `'../api/client'` (no extension). Cleaner to delete `client.js` entirely — Vite resolves `.ts` ahead of `.js`/`.jsx` in its default extension order. The re-export shim leaves a confusing dead-end file in the tree. (Minor — feel free to skip if you want to merge clean.)

3. **PR includes #143's Swagger UI changes too.** `docs/apidocs.go`, `docs/swagger-ui/*`, the API-test additions in `internal/api/api_test.go`, and the docs/router changes are all from #143. That's fine if these are stacked, but please coordinate the merge order so we don't get a messy three-way merge — either land #143 first then rebase #146, or merge #146 with the understanding that #143 becomes a no-op.

4. **CI confirmation.** The PR description lists `npm run build --prefix web` and `make test-web` — please confirm both are green in CI before merge. `npm run build` passed locally; `make test-web` requires a Playwright runtime I couldn't get working in this env.

Approving — once #143's status is settled and CI is green, ship it.

---

PR #147, opened 2026-04-29
feat(events): Activity page, VMDetail Activity tab, vmsmith events list (4.2.2/12/13/14)

chinmay28:
## Summary

Vertical slice for the **Activity / events** surface, built on the `GET /api/v1/events` endpoint already shipped on main. Closes (or substantially closes) four roadmap items:

| Item | Status |
|---|---|
| **4.2.2** — Extend `pkg/types/event.go` with full record (Source, Severity, Attributes, Actor, OccurredAt, ResourceID) + `EventSchemaVersion` | ✅ |
| **4.2.12** — VMDetail "Activity" tab | ✅ (page-based pagination; `until=<id>` infinite scroll deferred until SSE / 4.2.10) |
| **4.2.13** — Top-level Activity page with filters and deep-link URL params | ✅ |
| **4.2.14** — `vmsmith events list` CLI | ✅ partial (`events follow` deferred until SSE / 4.2.10) |

### Frontend
- `web/src/pages/Activity.jsx` — new top-level page with VM / source / severity / type filters; filters are mirrored to URL search params so views can be deep-linked. Backend pagination via `X-Total-Count`. Polls every 5 s.
- `web/src/pages/VMDetail.jsx` — Overview / Activity tabs. The Activity tab embeds the same component pre-filtered with the current VM ID.
- `web/src/api/client.js` — adds `events.list({ vmId, type, source, severity, since, page, perPage })`.
- `web/src/components/Layout.jsx` + `App.jsx` — Activity nav link and `/activity` route.

### Backend
- `pkg/types/event.go` — extend the `Event` record with `Source`, `Severity`, `Attributes`, `Actor`, `ResourceID`, `OccurredAt`. All new fields use `omitempty`, so events written before this change round-trip unchanged. Adds `EventSchemaVersion = 1`.

### CLI
- `internal/cli/events.go` — `vmsmith events list` reads the bbolt store directly (no daemon required) and supports `--vm`, `--type`, `--source`, `--severity`, `--since` (Go duration like `5m` or RFC3339), and `--limit`. Output is a tab-aligned table sorted newest-first.

### Tests
- `internal/cli/events_test.go` — empty store, all-fields rendering, vm/source filters, newest-first ordering, `--since 5m`, `--limit`, invalid `--since` rejection (8 tests).
- `tests/web/mock-server.js` — fixtures for `/api/v1/events` (libvirt + app + system rows).
- `tests/web/gui.spec.js` — Playwright coverage: nav reaches Activity, rows render, source filter, VMDetail Activity tab.

## Test plan
- [x] `go build ./...` — clean
- [x] `go test -race -count=1 ./...` — only pre-existing failures remain (`TestUploadImage_MissingName`, `TestCLI_TemplateCreate_RejectsInvalidInput/duplicate_name`); both reproduce on bare `origin/main` and are unrelated to this change.
- [x] `npm run build --prefix web` — clean
- [x] `node --check tests/web/{gui.spec,mock-server}.js` — clean

## Notes
- The `events follow` SSE command is intentionally deferred until 4.2.10 (SSE endpoint) lands.
- `until=<id>` infinite scroll on the embedded VMDetail tab is also deferred until ID-cursor pagination ships in 4.2.9; for now the embedded tab uses page-based pagination.

---
_Generated by [Claude Code](https://claude.ai/code/session_01UGpJZK2aW44SqYCx9UcDhG)_


chinmay28 (commented):
Solid vertical slice and the test coverage is good — but there are two issues that need to be addressed before this can merge cleanly. CI is green, but only because the PR was opened against `aca9cdb`, before #146 (`feat(web): generate frontend API types from OpenAPI`) landed on `main`.

## Blocking

### 1. `web/src/api/client.js` will conflict with `main` and the resolution is non-trivial

After #146, `web/src/api/client.js` on `main` is just:
```js
export * from './client.ts';
export { default } from './client.ts';
```
The real client lives in `client.ts`, built on `openapi-fetch` against `web/src/api/generated/schema.d.ts`, which is generated from `docs/openapi.yaml` via `npm run generate:api`. I tried merging `main` into the PR head locally and got a hard conflict in `client.js`. Resolution paths, in order of preference:

- **(Preferred)** Add the `/events` endpoint to `docs/openapi.yaml` (it currently has no `/events` entry at all, despite the endpoint existing on `main`), regenerate the schema, and add `events.list` to `client.ts` using the typed client. This is consistent with how every other resource is wired now.
- Fall back to a raw `fetch`-based shim inside `client.ts` if the OpenAPI work is out of scope here, but at minimum it must live alongside `client.ts`, not as a parallel `client.js`.

### 2. The Activity page exposes filters the daemon doesn't honor

`internal/api/handlers_events.go` on `main` only filters on `vm_id` and `since`. This PR doesn't extend it — that work is roadmap 4.2.9, explicitly deferred. But the new `Activity.jsx` ships UI controls for **source / severity / type** and serializes them into query params (`?source=…&severity=…&type=…`). The mock server obeys those params, so `tests/web/gui.spec.js` passes, but against a real daemon the user will pick a filter and see no change in results. Two options:

- Land 4.2.9's filter additions to `handlers_events.go` as part of this PR (the diff is small — mirror the existing `vm_id` / `since` blocks, plus add `until=<id>` cursor support). This is the cleaner shipping story.
- Otherwise, hide / disable the source/severity/type chips behind a feature flag or comment them out until 4.2.9 lands, and call that out in the PR description.

## Non-blocking observations

- **CLI reads bbolt directly instead of calling the REST API.** The roadmap entry for 4.2.14 says "one-shot REST query"; this implementation opens the store via `openStoreForCLI` (`internal/cli/events.go:175`). The advantage is that `vmsmith events list` works without a daemon. The downside is that filters are applied client-side after `ListEvents()` returns the entire bucket, and in a few releases that becomes O(N) on every invocation. Probably fine for now given the bucket is bounded by retention, but worth either (a) noting in the command's `Short`/help that it reads the local DB or (b) flipping it to hit `/api/v1/events` with the same filters the API is about to grow.
- **`Activity.jsx:11-16`** has very long inline className strings for the warn/info/default badges. Not a correctness problem, but it'd be cleaner pulled into the existing `Shared.jsx` badge utilities (where `badge-error` already lives).

## Verification (local)
- `go build ./...` clean.
- `go test -count=1 -short ./...` — only the two pre-existing failures remain (`TestUploadImage_MissingName`, fixed by #148, and `TestCLI_TemplateCreate_RejectsInvalidInput/duplicate_name`, also pre-existing on `main`).
- `go test -run TestCLI_Events ./internal/cli/...` — all 8 new tests pass.
- `npm run build --prefix web` — clean.

Treat this as **changes requested** — happy to re-review once `client.js` is reconciled with `client.ts` and the UI/API filter mismatch is resolved (either by extending the handler or trimming the filters). Everything else looks good — the schema-versioned event type, the additive `omitempty` JSON tags, the URL-mirrored filter state, and the CLI test coverage are all in good shape.

---

chinmay28:
Pushed `651a64b` — rebased onto current `main` (`4b6dcea`, includes #145 and #146) and addressed review.

## Blockers

| # | Status | Notes |
|---|---|---|
| 1 — `client.js` conflict / OpenAPI alignment | ✅ | Added `/events` endpoint + `Event` schema to `docs/openapi.yaml`; ran `npm run generate:api`; moved `events.list` into the typed `client.ts` (`apiClient.GET('/events', …)`). The `client.js` re-export shim is left untouched. |
| 2 — UI/API filter mismatch | ✅ Resolved by rebase | After #145 merged, `handlers_events.go` honors `type`, `source`, `severity`, `until`, `since`, `page`, `per_page`, exactly the filters the Activity page exposes. |

## Non-blocking observations

| # | Status | Notes |
|---|---|---|
| 3 — CLI reads bbolt directly | ✅ Documented | Added `Long` help text on `vmsmith events list` calling out that it reads the local store and applies filters client-side; pointed to `GET /api/v1/events` for daemon-side filtering. Switching the CLI to REST is left as a follow-up. |
| 4 — Inline badge classes in `Activity.jsx` | ✅ Refactored | Added `SeverityBadge` to `web/src/components/Shared.jsx` and replaced the local `severityBadge()` helper. |

## Local verification

- `CGO_ENABLED=1 go build ./...` — clean.
- `CGO_ENABLED=1 go test -short -race -count=1 ./...` — all green (api / cli / events / metrics / store / vm / network / etc.).
- `npm run build --prefix web` — clean.
- `npx openapi-typescript ../docs/openapi.yaml -o src/api/generated/schema.d.ts` — clean.

Waiting on GitHub Actions.

---

PR #149, opened 2026-04-29
feat(events): emit app+system events from handlers; add retention loop (4.2.6/7/8)

chinmay28:
## Summary

Wires the `EventBus` shipped in #145 into the rest of the daemon so events are actually populated and bounded. Closes (or substantially closes) three roadmap items:

| Item | Status |
|---|---|
| **4.2.6** — Emit `app`-source events from API handlers | ✅ |
| **4.2.7** — Emit `system`-source events for daemon start / shutdown | ✅ partial (quota / DHCP / port-forward-restore failure events deferred) |
| **4.2.8** — Retention loop with `events.max_records` | ✅ (`max_age` knob deferred — current `Store.PruneEvents` is record-count-only) |

### Backend — app events (`4.2.6`)
Every mutating endpoint now calls `Server.publishAppEvent` (the no-op-when-bus-nil helper added by #145):

- `handlers_vm.go` — `vm.created`, `vm.updated`, `vm.cloned`, `vm.deleted`, `vm.start_requested`, `vm.stop_requested`
- `handlers_snapshot.go` — `snapshot.created`, `snapshot.restored`, `snapshot.deleted`
- `handlers_image.go` — `image.created`, `image.uploaded`, `image.deleted`
- `handlers_network.go` — `port_forward.added`, `port_forward.removed`

Each event carries useful attributes (snapshot name, source VM ID for clones, host/guest port + protocol for forwards, etc.). Actor capture via auth-context is parked behind 3.1 — the helper signature is unchanged so it's a one-line drop-in later.

### Daemon system events (`4.2.7`)
- `daemon.started` (info) emitted right after the bus starts — with `listen` + `tls` attributes.
- `daemon.shutdown` (info) emitted just before `closeResources()` so it's persisted before the bus is stopped.
- `events.retention_pruned` (info, system) emitted on each retention sweep that deletes anything.

### Retention loop (`4.2.8`)
- New `internal/events/retention.go` runs `Store.PruneEvents(maxRecords)` once at startup and then on a ticker. Configurable via `events.max_records` (default 50000) and `events.retention_interval` seconds (default 60). Zero on either disables the loop entirely.
- Wired into `internal/daemon/daemon.go` after the bus starts and runs under the same `signal.NotifyContext` so it stops cleanly on SIGINT/SIGTERM.

### Config + example
- `internal/config/config.go` — new `EventsConfig` block, `Events` field on root config, defaults filled in `DefaultConfig()`.
- `vmsmith.yaml.example` — new `events:` section with both knobs documented.

### Tests
- `internal/events/retention_test.go` (new, 7 tests):
  - disabled when `maxRecords == 0`
  - disabled when `interval == 0`
  - runs once at startup
  - periodic sweeps fire on the ticker
  - publishes `events.retention_pruned` when something was deleted (asserting type / source / attributes)
  - emits no event when nothing was deleted
  - prune error doesn't panic
  - `itoa` helper coverage
- `internal/api/handlers_events_emission_test.go` (new, 3 tests): wires a real `EventBus` to the test API server, exercises the VM lifecycle (create / start / stop / delete) and snapshot CRUD, and asserts the right event types / sources / VM IDs / attributes arrive on a subscriber. Plus a no-bus regression test for `publishAppEvent`.

### Docs
- `docs/ROADMAP.md` — 4.2.6 / 4.2.7 / 4.2.8 marked ✅ with the deferred sub-items called out next to the checkmark.

## Test plan

- [x] `CGO_ENABLED=1 go build ./...` — clean
- [x] `CGO_ENABLED=1 go test -short -race -count=1 ./...` — all packages green (api / cli / config / daemon / events / metrics / store / vm / network / storage / logger)
- [x] `CGO_ENABLED=1 go test -run TestRetention ./internal/events/...` — 7/7 pass
- [x] `CGO_ENABLED=1 go test -run TestVMLifecycleEmits ./internal/api/...` — 3/3 pass

## Notes / follow-ups

- **Actor capture (4.2.6 sub-item).** `publishAppEvent` doesn't yet read an actor from the auth-middleware context. Defer until 3.1 ships (no auth context exists today). One-line drop-in once it lands.
- **`max_age` retention (4.2.8 sub-item).** `Store.PruneEvents(maxRecords int)` from #145 is record-count-only. Adding `max_age` is a small follow-up — extend the signature, walk forward keys with timestamp filter, then re-thread `cfg.Events.MaxAge` through `NewRetention`.
- **Quota / DHCP / port-forward-restore system events (4.2.7 sub-items).** One-line publishes alongside the existing `logger.Warn` calls in those code paths; not done in this PR to keep the diff focused on the wiring story.

---
_Generated by [Claude Code](https://claude.ai/code/session_01CpsMtV2zgVvLmW8SKr7gYR)_


chinmay28 (commented):
## Review summary

LGTM — focused, well-tested wiring of the existing `EventBus` from #145 into the rest of the daemon. Aligns with roadmap items 4.2.6/4.2.7/4.2.8 and the deferred sub-items are clearly called out both in the PR body and inline in the roadmap entries.

(GitHub doesn't let me formally approve since this is your own PR, leaving as a comment review instead.)

### What I verified locally

- `CGO_ENABLED=1 go test -short -race -count=1 ./...` — all packages green
- `CGO_ENABLED=1 go test -short -race -count=3 -run "TestVMLifecycleEmits|TestSnapshotCRUDEmits|TestPublishAppEvent_NoBus|TestRetention" ./internal/api/... ./internal/events/...` — clean across repeats; no flake from the 2s `drainEvents` deadline.
- Walked the shutdown path: `Daemon.Run` publishes `daemon.shutdown` *before* `closeResources()`, and `EventBus.run` drains `publishCh` on `Stop()` (`internal/events/bus.go:147-165`), so the shutdown event is persisted reliably.
- Confirmed `publishAppEvent` is no-op when `s.eventBus == nil` (`internal/api/handlers_events.go:179`) so existing `httptest`-based callers without a bus continue to work.

### Architectural fit

- Event-emission helper centralised on `*Server` keeps every mutating handler one-line and consistent. Good call leaving the actor parameter out for now and keying off ctx once 3.1 lands.
- Retention loop is properly gated by config (`MaxRecords > 0 && RetentionInterval > 0`), runs once at startup so a daemon restart catches up, and emits an observable `events.retention_pruned` event only when something was actually deleted. Solid.

### Nits (none blocking)

1. `internal/events/retention.go` re-implements `strconv.Itoa` as `itoa()`; `daemon.go` already imports `strconv` for `FormatBool` so the dependency rationale doesn't really hold. Replacing with `strconv.Itoa` would drop ~25 lines (and the matching test) without changing behaviour.
2. `image.created` uses the source VM's ID as the event's `vm_id`, which is fine, but `image.deleted` / `image.uploaded` pass `""` because the VM context is unknown by then. Worth keeping consistent in mind for the upcoming `?vm_id=` filter UX — those two events won't show up under any per-VM activity tab. Not a blocker for this PR.
3. The deferred items (actor capture, `max_age`, quota / DHCP / port-forward-restore system events) are flagged in the roadmap entries — please open follow-up issues so they don't get lost.

---

PR #151, opened 2026-04-30
feat(metrics): top-N VMs leaderboard API + CLI + dashboard widget (4.1.8)

chinmay28:
## Summary

Closes roadmap item **4.1.8** with a full vertical slice — API, CLI, and GUI — for ranking VMs by their latest in-memory metric sample. No new storage, no new daemon goroutines: the endpoint just walks the existing rings.

### Backend — `GET /api/v1/vms/stats/top`
- New handler in `internal/api/handlers_metrics.go` and route in `internal/api/router.go` (alongside `/vms/bulk` so chi resolves the static segment before `{vmID}`).
- Query params:
  - `metric` (default `cpu`): one of `cpu | mem | disk_read | disk_write | net_rx | net_tx`.
  - `limit` (default 5, range 1–100).
  - `state` (default `running`, also `all`).
- Returns `503 metrics_disabled` when the metrics manager is nil; `400 invalid_metric` / `invalid_limit` / `invalid_state` for malformed input.
- Sort: descending by metric value with a deterministic VM-ID tiebreak so the response is stable across polls.
- Items missing the requested metric (or whose VM has no current sample) are skipped — they don't get bumped to the bottom with zeros.

### CLI — `vmsmith vm top`
- `internal/cli/vm.go`: new `vmTopCmd` with `--metric --limit --state --api-url` flags.
- Renders a `tabwriter` leaderboard with units-aware formatting (`%`, `MB`, `B/s` / `KB/s` / `MB/s` / `GB/s`).
- Surfaces the daemon's `metrics_disabled` 503 with a friendly message.

### GUI — Dashboard "Top 5 Machines" card
- `web/src/pages/Dashboard.jsx`: new `TopVMsCard` with a metric selector (CPU / Memory / Disk Read / Disk Write / Net RX / Net TX), 10s polling, click-through to VM detail, and an empty-state message.
- `web/src/api/client.ts`: `vms.top({ metric, limit, state })` wrapper around the typed `openapi-fetch` client.

### OpenAPI + types
- `docs/openapi.yaml`: new `/vms/stats/top` path plus `MetricSample`, `TopVMItem`, `TopVMResponse` schemas.
- `web/src/api/generated/schema.d.ts` regenerated via `npm run generate:api`.

### Tests
- `internal/api/metrics_test.go` — 10 new cases: metrics-disabled 503, default metric/limit, descending order, limit truncation, running-vs-all state filter, skipping VMs without the metric, invalid metric/limit/state, empty list.
[...]

PR #152, opened 2026-04-30
feat(web): VMDetail Metrics tab with REST polling (4.1.7 partial)

chinmay28:
## Summary

Catches the GUI up to the existing per-VM `/api/v1/vms/{id}/stats` endpoint (and the matching `vmsmith vm stats` CLI) by adding a **Metrics** tab to VMDetail. Closes the GUI half of roadmap item **4.1.7** with the table-based polling slice; chart visualisation (uPlot) and the SSE subscription path are deferred until 4.1.5 ships, as called out next to the ✅ in the roadmap entry.

### OpenAPI schema (`docs/openapi.yaml`)
- New `/vms/{vmID}/stats` path with `?since=` and `?fields=` query params plus `200 / 400 / 404 / 503` responses.
- New `MetricSample` and `VMStatsSnapshot` component schemas mirroring `pkg/types/metrics.go` (all metric fields nullable so the daemon's "counter reset / no guest agent / first sample" semantics survive the wire).

### Frontend client (`web/src/api/client.ts`)
- Regenerated `web/src/api/generated/schema.d.ts` (`npm run generate:api`).
- Added `vms.stats(id, { since, fields })` wrapper backed by the typed `openapi-fetch` client.

### VMDetail Metrics tab (`web/src/pages/VMDetail.jsx`)
- New **Metrics** tab between Overview and Activity.
- New `VMMetrics` component polls `vms.stats(id)` every 10s, renders three info cards (state, last sampled, history meta) and a table with current vs 5-minute average for CPU, memory used, memory available, disk read/write and net rx/tx.
- Uses the existing `useFetch` polling hook and `Shared.EmptyState` so stopped or freshly-created VMs see a friendly message instead of a half-empty table. Format helpers handle nulls (`n/a`) and choose units (`B/s` → `KB/s` → `MB/s` → `GB/s`).

### Tests
- `tests/web/mock-server.js`: serves a deterministic 6-sample `/api/v1/vms/{id}/stats` payload (oldest-first, increasing CPU/mem/I/O) so the test asserts CPU 35% on the most recent sample regardless of wall clock. Stopped VMs return frozen history with `current = null` to mirror the real daemon's stale-sample behavior.
- `tests/web/gui.spec.js`: 4 new Playwright cases covering the running-VM table, history meta, stopped-VM empty state, and tab independence (metrics ⇄ overview).
- `tests/web/run-gui-tests.js`: 4 matching cases for the standalone runner used in CI.

### Roadmap
- `docs/ROADMAP.md` 4.1.7 marked **✅ Partial** with the deferred sub-items spelled out next to the checkmark.

## Test plan

- [x] `npm run generate:api` — types regenerate cleanly from the new OpenAPI paths
- [x] `npm run build` (frontend) — bundle builds with no warnings
- [ ] `make test-web` — Playwright runs in CI (chromium download blocked in this sandbox)
- [ ] `make test` — Go suite runs in CI (libvirt headers unavailable locally; no Go code changed)

## Notes

- No Go code changed. `client.ts`, `mock-server.js` and `VMDetail.jsx` are the only product-code edits; the OpenAPI YAML is documentation that the typed client compiles against.
- Independent of the other open chinmay28 PRs (#149 events emission, #151 top-N leaderboard) — touches different paths and uses a different client method (`vms.stats` vs `vms.top`).

---

---
_Generated by [Claude Code](https://claude.ai/code/session_01SwzsPtPa8dasUbE2GM3WRE)_


chinmay28 (commented):
LGTM — review notes (cannot self-approve as PR author):

**Verified locally:**

- `CGO_ENABLED=1 go build ./...` clean (no Go code changed, but the OpenAPI yaml still validates by virtue of building with the existing tooling).
- `CGO_ENABLED=1 go test -short -count=1 ./...` — every package green.
- `npm run build` in `web/` succeeds and produces the embedded SPA bundle.
- Branch merges cleanly with current `main` (one commit ahead with #149).

**Architecture / design fit:**

- `web/src/api/client.ts:77-78` adds the `vms.stats(id, { since, fields })` wrapper using the typed `apiClient.GET('/vms/{vmID}/stats', ...)` path — consistent with the rest of the file.
- `VMMetrics` reuses the existing `useFetch(fn, deps, pollInterval)` hook with a 10 s poll cadence, matching the daemon's `metrics.sample_interval` default. Good fit for "table-based polling slice" until 4.1.5 (SSE) lands.
- The 5-minute average is computed client-side by filtering `history` in the browser — no extra API round trips. Matches the CLI's behaviour in `printStatsSnapshot`.
- Format helpers (`fmtPercent`, `fmtMB`, `fmtBps`) handle the nullable-pointer fields correctly: a missing field becomes `n/a` rather than `0` or `NaN`. Avoids the same misleading-zero pitfall the API skip-without-metric design was avoiding.
- Empty-state branches: stopped VM ("not running — metrics resume after start") vs. running but no samples yet. Both come from the same `cur` null check in `VMMetrics`, so the table only appears when data exists.

**OpenAPI fidelity:**

- `MetricSample` and `VMStatsSnapshot` schema fields (`cpu_percent`, `mem_used_mb`, `mem_avail_mb`, `disk_read_bps`, etc.) match `pkg/types/metrics.go` exactly, with `nullable: true` reflecting the pointer-fields-mean-unavailable contract.
- `?since=` and `?fields=` query params and the 200/400/404/503 response set match the existing handler at `internal/api/handlers_metrics.go`.

**Roadmap alignment:**

- 4.1.7 marked as **partial** with chart/SSE deferred until 4.1.5 ships — the right call. The polling table is a cheap and useful intermediate; uPlot/SSE work doesn't bottleneck on this PR.

**Minor nits (non-blocking):**

- I couldn't run `make test-web` locally (Playwright isn't preinstalled in this sandbox). Trusting CI for the 4 new mock-GUI cases — they all use stable `data-testid` selectors so they should be deterministic.
- `MetricRow` is repeated inline 7 times. If a future change adds rows or thresholds, a `[label, field, formatter]` config array would shrink it — but at 7 rows it's fine as-is.

---

PR #154, opened 2026-04-30
feat(events): `events follow` CLI + `useEventStream` hook (4.2.11 / 4.2.14)

chinmay28:
## Summary

Closes roadmap items **4.2.11** (frontend SSE consumer + live indicator) and **4.2.14** (`vmsmith events follow` CLI) on top of the `/api/v1/events/stream` endpoint shipped with #145. No new daemon endpoints — both consumers ride the existing SSE bus.

### CLI — `vmsmith events follow`

- New `eventsFollowCmd` in `internal/cli/events.go` opens an SSE stream against the daemon's `/api/v1/events/stream`, parses frames with a hand-rolled SSE reader, and prints matching events as they arrive in the same tab-separated format as `events list`.
- Filters (`--vm` / `--type` / `--source` / `--severity`) are applied client-side after the wire delivers each event so the daemon stays free of new query params.
- Reconnects on transient errors with exponential backoff (1s → 30s cap), passing the most recent observed event id via both `Last-Event-ID` header and `?since=` query so the daemon's replay window kicks in. 401 and 410 are treated as fatal so the user sees a clear error instead of a reconnect loop.
- Ctrl-C exits cleanly via `signal.NotifyContext`.

### Frontend — `useEventStream` hook + `LiveIndicator`

- New `web/src/hooks/useEventStream.js` wraps `EventSource` with auto-reconnect, exponential backoff, and a short-poll fallback that kicks in after 3 consecutive failures or an HTTP 410 (`event_stream_replay_window_exceeded`). The fallback path uses `GET /api/v1/events?until=<lastID+1>` so listeners keep advancing in chronological order.
- Resumption uses `?since=<lastID>` because browsers can't set custom headers on `EventSource`. The bearer token is appended as `?api_key=` (same-origin only) — the design called out in roadmap §4.2.0.
- `Dashboard.jsx` and `VMList.jsx` drop their 5 s poll cadence to 30 s, register an `onEvent` handler that immediately refetches on `vm.*` / `image.*` events, and surface the connection state via a new `LiveIndicator` pill in `components/Shared.jsx`.
- The hook exposes `{ status, lastEventId, lastEvent }` so other pages can reuse it incrementally without touching this PR.

### Tests

- `internal/cli/events_test.go`: 9 new Go cases — filter matching, `Authorization: Bearer` forwarding, fatal 401/410 handling, reconnect with `Last-Event-ID`, heartbeat-comment skipping, and `lastIDToSeq` edge cases. Uses `httptest.NewServer` to drive a deterministic mock SSE stream.
- `tests/web/mock-server.js`: new `/api/v1/events/stream` endpoint that emits two canned events and a heartbeat so the frontend's `onopen` → `STATE_LIVE` transition fires deterministically.
- `tests/web/run-gui-tests.js`: 2 new cases asserting the dashboard / VM list `live-indicator` reaches `data-status="live"`.
- `tests/web/gui.spec.js`: 3 new Playwright cases covering both pages plus a fallback test that returns HTTP 410 and asserts the indicator transitions to reconnecting/fallback.

### Roadmap

- `docs/ROADMAP.md` 4.2.11 and 4.2.14 marked ✅ Done with implementation notes.

## Test plan

- [x] `CGO_ENABLED=1 go test -race -count=1 ./...` — all packages green
- [x] `make web` — frontend bundle builds clean
- [x] `node tests/web/run-gui-tests.js` — 19/19 pass (incl. 2 new live-indicator cases)
- [x] `node tests/web/run-built-web-regression.js` — 4/4 pass

## Independence from open chinmay28 PRs

- Different code paths from #151 (`vm.top` leaderboard) and #152 (VMDetail Metrics tab). No merge conflicts expected.

---

---
_Generated by [Claude Code](https://claude.ai/code/session_015WaLJ9PSQwwCCjMFEEzMp2)_


chinmay28 (commented):
LGTM — review notes (cannot self-approve as PR author):

**Verified locally:**

- `CGO_ENABLED=1 go build ./...` clean.
- `CGO_ENABLED=1 go test -short -race -count=1 ./...` — every package green.
- `CGO_ENABLED=1 go test -run TestFollow -v -count=1 ./internal/cli/...` — all 8 new SSE follow cases pass: prints/filter/auth-header/auth-failure/410/reconnect/heartbeat-skip.
- `npm run build` in `web/` succeeds cleanly with the new hook.
- Branch merges cleanly into current `main`.

**CLI (`vmsmith events follow`)**

- `streamEventsOnce` correctly maps 401/403 and 410 to `streamErrFatal` so the user sees a real error instead of hammering reconnects. Other non-200s fall through to transient retry — sensible default.
- Reconnect uses both `Last-Event-ID` header and `?since=<seq>` query so the daemon can pick up replay either way. `lastIDToSeq` correctly rejects legacy `evt-<unix-nano>` ID strings instead of forwarding garbage; that fits the EventBus seq contract.
- `Client.Timeout: 0` for streaming is correct — net/http would otherwise close long-lived idle SSE.
- SSE parser's `scanner.Buffer(..., 1024*1024)` lifts the default 64K line cap so a single large JSON event doesn't break the stream. Good defensive choice.
- Filters are applied after parse (deliberate per the PR description), so the daemon stays free of new query params. Reasonable for a CLI consumer.
- Ctrl-C path via `signal.NotifyContext` is right — context cancel propagates into the Scanner via the closing reader.

**Frontend (`useEventStream` + `LiveIndicator`)**

- Reconnect ladder (1s → 2s → … → 30s cap) and the 3-failure fallback threshold match the roadmap spec for 4.2.11.
- Fallback `?until=<lastSeq+1>` polling against `/api/v1/events` is the right shape — listeners keep advancing in chronological order even when SSE is down. The reverse-iteration over the descending list to dispatch oldest-first is subtle but correct, and the `seq > cursor` guard prevents replay duplicates.
- `closedRef.current` early-return in every async path (`connect`, `scheduleReconnect`, `startFallback`, fallback `poll`) keeps us safe across unmount races.
- `onEvent` is held in a ref so callers can change `handleEvent` without ripping the connection — important since `Dashboard.jsx` rebuilds `handleEvent` on every refresh callback identity change.
- Browser-only `?api_key=` for EventSource matches the daemon's documented same-origin auth fallback (4.2.10).
- The Dashboard / VMList drop from 5 s to 30 s polling, with the hook poking refresh on `vm.*` / `image.*` events — net reduction in busy-poll while staying responsive. Nice.

**Test coverage**

- `internal/cli/events_test.go` runs a real `httptest.NewServer` against the SSE parser, so the fatal-vs-transient classification, header forwarding, heartbeat skip, and reconnect-on-disconnect are all hit end-to-end. Solid.
- I couldn't run `make test-web` here (no Playwright in sandbox), but the 3 new mock-server cases use stable `data-testid` selectors and the 410 fallback test mocks the route directly with `page.route()`. CI should be deterministic.

**Minor nit (non-blocking)**

- `connect` and `scheduleReconnect` form a circular `useCallback` pair (each calls the other; only `scheduleReconnect` lists `connect` in deps). It works because `connect`'s deps `[dispatchEvent, startFallback]` are stable refs, so neither callback churns. If a future refactor adds a non-stable dep to `connect`, the pair would start re-creating each render and the existing `EventSource` would tear down/reopen on every render. A short comment near `scheduleReconnect`'s definition explaining the contract would prevent that footgun, but the current code is fine.

---

Issue #156, opened 2026-05-01
docs: sync ROADMAP metrics status with shipped implementation


march26bot:
$## Summary\nSeveral Phase 4.1 roadmap items are still listed as pending even though the implementation is already in tree. This makes it harder to spot the real remaining work.\n\n## Scope\n- audit the 4.1 metrics task list against the current code/docs\n- mark shipped items as done with short notes\n- tighten dependency/follow-up notes so the remaining gaps are still clear\n\n## Why\nThe repo already contains the metrics types/collector wiring, VM stats API, Prometheus scrape endpoint, CLI `vm stats`, and metrics docs. Keeping ROADMAP.md aligned with the current tree makes issue selection and planning more reliable.

PR #157, opened 2026-05-01
docs(roadmap): sync shipped metrics status

march26bot:
$## Summary\n- remove the stray merge-conflict markers left in the Phase 4.1 metrics table\n- mark shipped metrics items as done for Prometheus scraping, CLI `vm stats`, and metrics docs\n- keep the partial-status note for the VM detail metrics tab so the remaining SSE/chart work stays visible\n\n## Testing\n- python assertion that `docs/ROADMAP.md` contains no merge-conflict markers

PR #150, opened 2026-04-30
docs(roadmap): sync shipped metrics and events status

march26bot:
## Summary\n- refresh the roadmap status header to reflect shipped VM metrics, events, and OpenAPI work\n- mark completed 4.1 and 4.2 tasks that are already in \n- update the suggested next-priority section so it focuses on the remaining gaps instead of already-landed work\n\n## Testing\n- not run (docs-only change)

PR #158, opened 2026-05-01
docs: fix roadmap metrics status sync

march26bot:
$## Summary\n- remove the stray conflict markers accidentally left in `docs/ROADMAP.md`\n- mark shipped Phase 4.1 metrics docs/scrape items as done\n- keep 4.1.7 and 4.1.8 pending so the remaining frontend gaps stay visible\n\n## Testing\n- not run (docs-only change)\n\nCloses #156


chinmay28 (changes requested):
Thanks for cleaning up the conflict markers, but this version of the resolution regresses a couple of statuses that are demonstrably shipped on `main`. Recommend closing this PR in favor of #157, which targets the same conflict block but with the correct status for each row.

**Issues with the current resolution:**

1. **4.1.8 is marked as not-done (empty Notes column).** The dashboard "Top 5 VMs" widget shipped via [#151](https://github.com/chinmay28/VMSmith/pull/151) — see merge commit `772bcaf feat(metrics): top-N VMs leaderboard API + CLI + dashboard widget (4.1.8) (#151)`. The endpoint, the dashboard widget, and the `vmsmith vm top` CLI command are all in `main`. Leaving the Notes column blank is a regression from the HEAD side of the conflict.

2. **4.1.7 loses its "✅ Partial" status note.** The HEAD side of the conflict (#150's merge) already documented that the VMDetail Metrics tab landed via [#152](https://github.com/chinmay28/VMSmith/pull/152) (`feat(web): VMDetail Metrics tab with REST polling (4.1.7 partial)`) with charts/SSE deferred. This PR drops that note back to the original "Charts reuse the same hook; pass a `field` prop" planning text, which suggests no progress.

The PR description says *"keep 4.1.7 and 4.1.8 pending so the remaining frontend gaps stay visible"* — that's accurate for 4.1.7 (still partial) but not for 4.1.8 (fully shipped). #157 captures both correctly.

Suggest closing this PR; #157 resolves the same conflict markers and accurately reflects shipped state for all five affected rows.

---

PR #155, opened 2026-04-30
feat(events): emit system events for quota / DHCP / port-forward failures (4.2.7)

chinmay28:
## Summary

Closes the remaining gap on roadmap item **4.2.7** so operators can see quota breaches, DHCP range exhaustion, and port-forward restore failures land in the events stream — and therefore the existing Activity page, VMDetail Activity tab, and `vmsmith events list` / `events follow` CLI — without any UI changes.

Roadmap status before this PR:

> ✅ Done — `daemon.started` (with listen + tls attrs) and `daemon.shutdown` are now published from `internal/daemon/daemon.go`. Retention sweeps emit `events.retention_pruned`. **quota / DHCP / port-forward-restore failure publishes still pending**

After this PR all three pending emission points are wired in.

### Backend — `quota.exceeded`

- `internal/api/quotas.go`: a new `quotaExceeded` helper wraps the four limit checks in `validateQuotaUsage`. On every breach it (a) publishes a `quota.exceeded` system event via `Server.publishSystemEvent` and (b) returns the existing typed `quota_exceeded` API error so the HTTP 429 / message stay byte-identical.
- The event carries structured attributes `field` (`max_vms` / `max_total_cpus` / `max_total_ram_mb` / `max_total_disk_gb`), `attempted`, and `limit` so dashboards and webhook receivers don't have to scrape the human-readable message.

### Backend — `dhcp.exhausted`

- `internal/vm/lifecycle.go`: `LibvirtManager` gains an `eventBus *events.EventBus` field with `SetEventBus(bus)` and an `emitDHCPExhausted(vmName, reason)` helper. The two existing fall-back-to-dynamic warn paths in `Create()` and `Clone()` now publish a `dhcp.exhausted` system event with `vm_name` + `reason` attrs alongside the warn log.
- `LibvirtManager` is no longer hidden behind the `vm.Manager` interface during daemon startup so the daemon can call `SetEventBus` on it; the rest of the daemon still operates on `vm.Manager`.

### Backend — `port_forward.restore_failed`

- `internal/daemon/daemon.go`: the event bus is created right after the VM manager (rather than later in startup) so it is available before `PortForwarder.RestoreAll()` runs. When `RestoreAll()` returns an error on startup, the daemon publishes a `port_forward.restore_failed` system event (with the error string in attrs) in addition to the existing warn log.

### Backend — helpers

[...]

PR #159, opened 2026-05-01
feat(events): age-based retention pruning (4.2.8)

chinmay28:
## Summary

Closes the remaining gap on roadmap item **4.2.8** so the events retention loop honors `daemon.events.max_age` in addition to `daemon.events.max_records`. Operators get a 30-day default cap on event history without having to wait for the record-count limit to kick in.

> ✅ Done in part — max-record retention plus observable prune events shipped; **age-based pruning remains future work**

After this PR both phases run on every tick.

### Backend — `Store.PruneEventsByAge`

- `internal/store/bolt.go`: new `PruneEventsByAge(maxAge time.Duration) (int, error)`. Walks the events bucket in forward (chronological) order, deletes uint64-keyed entries whose `OccurredAt` is older than `now - maxAge`, and stops at the first non-stale entry. Capped at 5000 deletes per call so a backlog cannot stall the writer. `maxAge <= 0` is a no-op.

### Backend — `events.Retention`

- `internal/events/retention.go`: `PruneStore` now also requires `PruneEventsByAge`; `Retention` carries `maxAge` alongside `maxRecords`. Each tick runs both phases (whichever are enabled), aggregates the deleted counts, and emits a single `events.retention_pruned` system event whose attributes break down per-phase deletes and the configured limits (`deleted_max_records`, `deleted_max_age`, `max_records`, `max_age_seconds`).
- The loop is disabled only when both caps are zero or `interval <= 0`. Either cap alone now starts the loop.

### Backend — config

- `internal/config/config.go`: `EventsConfig.MaxAgeSeconds` (`yaml:"max_age_seconds"`, default `2592000` = 30 days). Zero disables age-based pruning.
- `internal/daemon/daemon.go`: passes `MaxAge` into `events.NewRetention` and starts the loop when either cap is set.
- `vmsmith.yaml.example`: documents the new field.

### Tests

- `internal/store/bolt_test.go`: `TestPruneEventsByAge` (stale/fresh mix, count + survivor check), `TestPruneEventsByAge_ZeroDisables` (zero and negative are no-ops), `TestPruneEventsByAge_KeepsAllWhenAllFresh`, `TestPruneEventsByAge_StopsAtFirstFresh` (verifies the chronological-walk stop semantics: a stale event written after a fresh one survives because the cursor stops at the first non-stale entry).
- `internal/events/retention_test.go`: rewrites the fake store to track both phases. New cases: `TestRetention_DisabledWhenAllLimitsZero`, `TestRetention_AgeOnlyEnabled`, updated `TestRetention_PublishesSystemEventOnDelete` to assert the new per-phase attributes, and the existing periodic-sweep / error-tolerance tests now exercise both prune calls.

### Roadmap

- `docs/ROADMAP.md` 4.2.8 row is now ✅ Done with a short note describing the per-phase event payload and the 30-day default.

## Test plan

## Notes

- No frontend or CLI changes were needed. The Activity page and `vmsmith events list` / `events follow` are event-type-agnostic; they will simply observe the existing `events.retention_pruned` event with the richer attribute set.
- This change is independent from #155 (system events for quota / DHCP / port-forward failures); both touch the events subsystem but in disjoint files.

---
_Generated by [Claude Code](https://claude.ai/code/session_01AEdTHqfjVJUp6sWUpx53XQ)_


chinmay28 (commented):
LGTM — clean closure of roadmap 4.2.8. (Submitting as a comment review since GitHub does not allow self-approval.)

**Verified locally:**
- `CGO_ENABLED=1 go vet ./...` clean
- `CGO_ENABLED=1 go test -short -race -count=1 ./...` all packages green (incl. `internal/store`, `internal/events`, `internal/config`, `internal/daemon`)
- Test merge against current `main` (9627eeb) is clean — `mergeable_state: clean` confirmed

**Architecture/design notes:**
- `PruneEventsByAge` chronological-walk with break-at-first-fresh is correct given `AppendEvent` writes monotonic uint64 sequence keys (matches the comment in `bolt.go`). The `TestPruneEventsByAge_StopsAtFirstFresh` test pins this invariant — including the deliberately-out-of-order timestamp case where a newer-by-clock event written after a fresh event survives because the cursor stops.
- 5000-per-sweep cap matches the existing `PruneEvents` write-budget pattern, keeping the writer responsive.
- The `len(k) != 8` filter correctly skips legacy string-keyed entries (the `BucketEvents` migration story called out in 4.2.5).
- `Retention.sweep()` correctly aggregates per-phase counts and emits a single `events.retention_pruned` event with the richer attribute set; phase-level errors are warn-logged but do not abort the other phase.
- Default 30 days (`MaxAgeSeconds: 2592000`) matches the roadmap target of 720h.
- The disabled-when-both-zero gating preserves the existing escape hatch (set both to 0 to disable retention entirely).

**Roadmap alignment:** matches 4.2.8 spec (every-60s sweep honoring max_records + max_age, 5000-cap, observable system event).

Approving conceptually — ready to merge once a maintainer signs off.

---

PR #162, opened 2026-05-01
feat(events): surface active SSE connection count in host_stats (4.2.10)

chinmay28:
## Summary

Closes the remaining "future polish" gap on roadmap item **4.2.10** — the SSE endpoint now reports its in-flight consumer count via `host_stats`, and the dashboard surfaces it next to the existing `LiveIndicator`.

> ✅ Done in part — SSE replay, heartbeat, overflow handling, and EventSource auth fallback are implemented; **connection-count host stats remain future polish**

After this PR all of 4.2.10 lands. While wiring the test harness, I also fixed a latent bug where the `requestLogger`'s `responseRecorder` was hiding the underlying `http.Flusher`, which silently broke SSE for any handler routed through the middleware.

### Backend — connection counter

- `internal/api/router.go`: `Server` gains an `atomic.Int64 eventStreamConns` field plus a public `EventStreamConnections()` accessor.
- `internal/api/handlers_events.go`: `StreamEvents` increments the counter on entry and `defer`s the decrement, so the value reflects the handler lifetime including the `event_stream_replay_window_exceeded` path.
- `internal/api/handlers_host_stats.go`: `GET /api/v1/host/stats` now sets `HostStats.EventStreamConnections` from the live counter.
- `pkg/types/quota.go`: `HostStats` grows `event_stream_connections` (`int64`, `json:"event_stream_connections"`).

### Backend — middleware fix (drive-by)

- `internal/api/middleware.go`: `responseRecorder` now forwards `http.Flusher` to the underlying `ResponseWriter`. Without this, the `requestLogger` middleware (applied to `/api/v1/...`) wrapped the writer and SSE failed with `streaming not supported` (HTTP 500). The new tests caught it because they hit the live `/api/v1/events/stream` endpoint through the full middleware chain.

### Frontend

- `docs/openapi.yaml` + `web/src/api/generated/schema.d.ts`: schema additions for the new field.
- `web/src/pages/Dashboard.jsx`: a small `"<n> sse"` badge appears next to the `LiveIndicator` when `host_stats` reports active consumers (hidden when zero, so the chrome stays quiet on a fresh dashboard).

### Tests

- `internal/api/sse_connections_test.go` (new):
  - `TestEventStreamConnectionsCounter` — opens one SSE client, asserts `host_stats.event_stream_connections` rises from `0 → 1` and drops back to `0` after disconnect.
  - `TestEventStreamConnectionsCounterMultipleClients` — three concurrent clients, each on a fresh transport so HTTP/1.1 keep-alive doesn't serialize them; closes one and asserts the counter drops to two.
- `tests/web/mock-server.js`: implements `GET /api/v1/host/stats` (it was missing entirely) with a non-zero `event_stream_connections` so the dashboard badge renders during mock-server runs.
- `tests/web/run-built-web-regression.js`: extends the routed mock with the new field for built-SPA regression coverage.
- `tests/web/gui.spec.js` + `run-gui-tests.js`: Playwright assertions for the `sse-connection-count` badge against the mock server.

### Roadmap

- `docs/ROADMAP.md` — 4.2.10 marked ✅ Done with the new wiring spelled out. While there I also marked **4.2.5 ✅ Done** because PR #159 already shipped `Store.PruneEventsByAge`, which closes that "remains future work" note.
- `README.md` — regenerated by `scripts/roadmap-progress.sh`.

## Test plan

- [x] `npm run build` (frontend) — clean

## Notes

- The increment intentionally happens **before** `newSSEWriter`, not after — that gives a happens-before relationship to the headers flush, so by the time the client sees `200 OK` headers the counter is already incremented (and the new test relies on this ordering).
- The `responseRecorder.Flush` fix is a fairly small behavior change but it makes any future SSE handler "just work" through the middleware. The existing `/api/v1/events/stream` worked in the wild because it was apparently never exercised end-to-end via the middleware in tests; the new tests close that gap.

---

---
_Generated by [Claude Code](https://claude.ai/code/session_01JybDWuUhGN693x5voKexeK)_


chinmay28 (commented):

**What I verified**

- Built and ran `CGO_ENABLED=1 go test -short -race -count=1 ./...` — all packages green, including the new `TestEventStreamConnectionsCounter` and `TestEventStreamConnectionsCounterMultipleClients` (each ~0.4s).
- Atomic counter is correct: `Add(1)` happens before `newSSEWriter` writes headers; `defer Add(-1)` covers the 410 / replay-overflow path and post-write disconnects, so the counter accurately reflects handler lifetime.
- The `responseRecorder.Flush()` forwarder is a real fix for a latent bug — without it, any handler routed through `requestLogger` that calls `http.Flusher` would 500. The new tests exercise it end-to-end through the full middleware chain.
- Inline `Content-Type` stripping middleware on `/events/stream` correctly lets the SSE handler set `text/event-stream`.
- OpenAPI / generated TS schema / mock-server / Playwright assertions are all updated consistently. Dashboard badge is gated on `> 0` so the chrome stays quiet on a fresh install.

**Note on overlap**

PR #164 adds a different fix to the same middleware (`Unwrap()` so `http.ResponseController` reaches the inner Flusher). Whichever lands first, the other will need trivial conflict resolution — either approach is correct, but #164's `Unwrap()` is more idiomatic for Go 1.20+ and chains cleanly through nested middleware. Worth keeping #164's version when reconciling.

Closes the "future polish" gap on roadmap 4.2.10 cleanly.

---

PR #163, opened 2026-05-01
refactor(vm): publish lifecycle events to bus, persist via consumer (4.2.4)

chinmay28:
## Summary

Closes the remaining gap on roadmap item **4.2.4**. The libvirt domain-lifecycle callback in `internal/vm/events.go` no longer mutates the bbolt store directly. Instead it translates each event into a typed `Event` (`Source=libvirt`, dotted type, severity) and publishes to the `EventBus`. State persistence now flows through a dedicated subscriber (`vm.VMStatePersister`) so the libvirt event-loop goroutine never holds a store transaction.

> Done in part — lifecycle handling and event-bus plumbing improved elsewhere, but `internal/vm/events.go` on `main` still mutates VM state directly via `store.PutVM` inside the libvirt callback

After this PR all of 4.2.4 lands.

## Why

- **Pre-refactor:** `handleLifecycleEvent` did `store.PutVM` and the legacy `store.PutEvent` inside the libvirt callback goroutine. A slow store contended with the event loop and the legacy `evt-<unix-nano>` path bypassed the bus's monotonic ID assignment, so SSE/CLI consumers never saw libvirt-driven state transitions live.
- **Post-refactor:** callbacks publish through `bus.Publish`. The bus assigns a monotonic ID, persists via `Store.AppendEvent`, and fans out to every subscriber — including the new persister, the SSE handler, the CLI `events follow` stream, and the Activity page.

### Backend

- **`internal/vm/events.go`** — `handleLifecycleEvent` resolves the VM by name, picks a stable dotted event type via `lifecycleEventToTypeAndSeverity` (`vm.started`, `vm.stopped`, `vm.shutdown`, `vm.suspended`, `vm.resumed`, `vm.crashed`, `vm.defined`, `vm.undefined`), records `state` + libvirt-event-name in attributes, and publishes. `vm.crashed` is published at `error` severity so dashboards/alerts can distinguish crashes from clean stops. A no-bus fallback path keeps the pre-refactor behavior for tests / early-startup harnesses that don't wire a bus.
- **`internal/vm/state_consumer.go`** (new) — `VMStatePersister` subscribes to the bus, ignores non-libvirt sources, libvirt events without `state` attrs, and events with empty `vm_id`, then writes the new state to bbolt via `store.PutVM`. Idempotent: skips writes when state is unchanged. Safe under nil bus or store.
- **`internal/daemon/daemon.go`** — starts `VMStatePersister` immediately after `bus.Start()`, before publishing `daemon.started`, so any libvirt event arriving after startup is consumed.

### Tests

- `internal/vm/state_consumer_test.go` (new):
  - applies libvirt events to the store and verifies state mutation
  - ignores non-libvirt sources, missing `state` attr, empty `vm_id`
  - is a no-op without bus or store
  - silently drops events for VMs that no longer exist
- `internal/vm/events_test.go` — extends coverage to `lifecycleEventToTypeAndSeverity` for every libvirt event code (started/resumed/suspended/shutdown/stopped/crashed/defined/undefined and the nil fallback).

### Roadmap

- `docs/ROADMAP.md` — 4.2.4 marked ✅ Done with the new wiring spelled out.
- `README.md` — progress badge regenerated by `scripts/roadmap-progress.sh`.

## Test plan

## Notes

- This refactor does not change any HTTP / SSE / CLI contracts. Existing consumers that read events from `Store.ListEventsFiltered` or the bus see the same fields, just with monotonic uint64 IDs (replacing the legacy `evt-<unix-nano>` strings the libvirt callback used to write).
- The legacy `Store.PutEvent` API is intentionally left in place — it's still used by tests and remains a stable seam for any future migration utility — but no production code path calls it anymore.
- This stacks cleanly on top of #162 (4.2.10) and is independent: the active SSE connection counter and this state-persister change touch different files.

---
_Generated by [Claude Code](https://claude.ai/code/session_01ThQiKpzd1aCrqFa94LQb4i)_


chinmay28 (commented):

**What I verified**

- Built and ran `CGO_ENABLED=1 go test -short -race -count=1 ./...` — all packages green, including the new `TestVMStatePersister_*` cases (`AppliesLibvirtEvent`, `IgnoresNonLibvirtEvents`, `NoStoreOrBus_NoOp`, `HandlesMissingVM`) and the expanded `TestLifecycleEventToTypeAndSeverity` table.
- The decoupling works as advertised: the libvirt callback no longer calls `store.PutVM` / `store.PutEvent`. Instead it builds a typed `Event` with `Source=libvirt`, dotted type, severity (notably `error` for crashed), and publishes through the bus. The `VMStatePersister` consumer is the single writer for libvirt-driven state changes, idempotent (skips writes when state is unchanged), and tolerant of deleted VMs.
- The no-bus fallback path preserves pre-refactor behavior for tests / early startup (writes state directly), while production code path always has a bus wired.
- Daemon startup order is correct: `bus.Start()` → `go persister.Run(ctx)` → publish `daemon.started`. Persister subscription happens before any libvirt monitor registration kicks off.

**Minor architectural nit (not blocking)**

There's a tiny window between `go persister.Run(ctx)` being scheduled and the inner `bus.Subscribe` call. In production this is fine because libvirt event registration happens later in startup, but a `Started()` synchronization signal would close the race entirely. The tests work around it with `time.Sleep(50ms)` after `go persister.Run(ctx)`, which is also a hint that something cleaner is possible later.

Closes the remaining gap on 4.2.4 — libvirt-driven state transitions now flow through the same SSE/CLI fan-out as `app` events, with monotonic uint64 IDs replacing the legacy `evt-` strings. Nice cleanup.

---

PR #164, opened 2026-05-01
feat(metrics): add VM stats SSE stream endpoint (4.1.5)

chinmay28:
## Summary

Closes roadmap item **4.1.5**. Adds `GET /api/v1/vms/{id}/stats/stream` — a Server-Sent Events endpoint that delivers a `vm.stats` frame for every new `MetricSample` produced by the in-memory metrics ring.

> 4.1.5 — `GET /api/v1/vms/{id}/stats/stream` SSE endpoint reusing 4.2.10's SSE helper. Frame per sample. Closes on VM delete or daemon shutdown

After this PR all of 4.1.5 lands.

### Backend — SSE handler

- `internal/api/handlers_metrics.go` — new `StreamVMStats` handler:
  - 503 `metrics_disabled` when the metrics manager isn't wired
  - 404 `resource_not_found` when the VM is unknown to the manager
  - On 200, sends an initial frame from the most recent sample (if any) so subscribers see something immediately, then polls the metrics manager every 1s and emits a frame whenever `Current.Timestamp` advances
  - Sample timestamps (unix-nanos) double as SSE event IDs so clients can resume after a reconnect using `Last-Event-ID`
  - 30s heartbeat comments via the existing `sseWriter`
  - Stream closes when the request context is cancelled or the VM is no longer present in the manager

- `internal/api/router.go` — registers `/vms/{vmID}/stats/stream` with the same Content-Type-stripping inline middleware used by `/events/stream`, so the JSON `Content-Type` middleware doesn't override `text/event-stream`.

### Plumbing fix — SSE through middleware

While wiring the new tests I caught a latent bug: the `requestLogger` middleware's `responseRecorder` does not expose flushing, so any SSE handler routed through it was returning HTTP 500 (`streaming not supported`).

- `internal/api/middleware.go` — `responseRecorder` now exposes `Unwrap() http.ResponseWriter`, which lets `http.ResponseController` reach the underlying `http.Flusher` (and any future `http.Hijacker`).
- `internal/api/sse.go` — `newSSEWriter` switches from a direct `Flusher` cast to `http.NewResponseController(w)`. The probe Flush is performed after `WriteHeader(200)` so the headers are still set correctly.

### Docs

- `docs/openapi.yaml` — adds the new SSE endpoint under `/vms/{vmID}/stats/stream`.
- `docs/ROADMAP.md` — marks 4.1.5 ✅ Done with the new wiring spelled out.
- `README.md` — progress badge regenerated by `scripts/roadmap-progress.sh`.

### Tests

- `internal/api/metrics_test.go` (new test functions):
  - `TestStreamVMStats_MetricsDisabled` — 503 + `metrics_disabled` code
  - `TestStreamVMStats_VMNotFound` — 404 + `resource_not_found` code
  - `TestStreamVMStats_DeliversInitialAndNewSamples` — opens the stream, asserts the seeded sample arrives as the first frame, seeds a newer sample, asserts it arrives next with a distinct event id
  - `TestStreamVMStats_ClosesOnVMDelete` — deleting the VM mid-stream causes the next frame read to error within 5 seconds
- `readSSEFrame` helper parses one full SSE frame (skipping `: keepalive` comment frames) so the assertions are robust against heartbeat timing.

## Test plan

- [x] `CGO_ENABLED=1 go test -short -race -count=1 ./...` — all packages green (incl. the new SSE tests)

## Notes

- This stacks cleanly on top of #162 (4.2.10 SSE connection counter) and #163 (4.2.4 lifecycle bus refactor). The `responseRecorder.Unwrap()` fix overlaps in spirit with #162's `responseRecorder.Flush` forwarder; once one of them merges the other is a trivial rebase. Either approach fixes the same latent bug.
- The 1s poll cadence is faster than the metrics sampler's default 10s tick so subscribers receive new samples within ~1s of being produced. The poll is a no-op when there's no new sample, so the cost is one Snapshot call per second per active subscriber — bounded by `Snapshot()` already being a single map read under RWMutex.
- Replay isn't supported by the in-memory metrics ring; the REST `/stats` endpoint provides initial backfill, and clients should fetch it on connect / reconnect. This matches the architectural plan in `docs/ROADMAP.md` §4.1.0.

---
_Generated by [Claude Code](https://claude.ai/code/session_018A9kA4nWXtznR4uB5G6jd5)_


chinmay28 (commented):

**What I verified**

- Built and ran `CGO_ENABLED=1 go test -short -race -count=1 ./...` — all packages green, including the new `TestStreamVMStats_*` cases (metrics-disabled 503, missing-VM 404, initial+new-sample delivery, close-on-VM-delete).
- The `http.ResponseController` + `responseRecorder.Unwrap()` pair is the right pattern for SSE through middleware. It chains cleanly through any future middleware that also implements `Unwrap()`, which is more robust than a per-middleware `Flush` forwarder.
- Polling cadence (1s) > sampler cadence (10s default) ensures new samples are delivered within ~1s of being produced; the per-tick Snapshot is a single map read under RWMutex so the cost is bounded.
- VM-deletion close path: the periodic `vmManager.Get` re-check inside the poll loop catches deletes within one tick — verified by `TestStreamVMStats_ClosesOnVMDelete`.
- Initial frame from `Snapshot.Current` ensures subscribers see something on connect rather than waiting up to a full poll interval.
- OpenAPI updated; `internal/api/router.go` registers the route under the same `Content-Type`-stripping inline middleware used by `/events/stream`.

**Notes on overlap with #162**

This and #162 both touch `internal/api/middleware.go`, fixing the same latent SSE-through-middleware bug two different ways. I'd keep this PR's approach (`Unwrap()` + `http.ResponseController`) when reconciling: it's the idiomatic Go 1.20+ pattern and the SSE rewrite in `internal/api/sse.go` removes the redundant Flusher cast in favor of the controller.

**Minor design consideration (not blocking)**

The poll-based delivery is fine for v1 but means every active SSE subscriber adds a 1s tick load. A future enhancement would have the metrics collector publish samples through the event bus (re-using 4.2.10's plumbing), turning this from poll-pull into push. Not in scope here, but worth a roadmap note.

Closes 4.1.5 cleanly.

---

PR #170, opened 2026-05-02
feat(metrics): live uPlot charts + SSE useVMStats hook (4.1.7)

chinmay28:
## Summary

Closes roadmap item **4.1.7**. The Metrics tab on VMDetail previously polled `/api/v1/vms/{id}/stats` every 10s and rendered a single text table. With **4.1.5** already shipping the `/stats/stream` SSE endpoint, this PR finishes the frontend half: live samples flow into four uPlot charts and a `LiveIndicator` surfaces the connection state (live / reconnecting / polling).

### Frontend hook (live samples + fallback)

- **`web/src/hooks/useVMStats.js`** — initial REST fetch against `/stats`, then opens an `EventSource` against `/stats/stream` with `?api_key=` for the browser auth fallback the daemon already accepts. Reconnects with exponential backoff (`1s/2s/4s/...` capped at 30s) and falls back to short-poll mode after 3 consecutive failures so the chart keeps advancing on locked-down proxies. Listens to both the `vm.stats` named event channel and the default `message` channel so it works whether the server names the event or not.
- **`web/src/hooks/vmStatsHelpers.js`** — pure helpers (`appendSample`, `sampleTime`, `buildChartData`) extracted into a no-React module so they are testable under `node --test` without a DOM shim.

### Charts

- **`web/src/components/MetricChart.jsx`** — React wrapper around uPlot that owns the canvas lifecycle through refs, supports multi-series charts, and resizes the canvas via a `ResizeObserver`. Series colors and units are passed in by the caller; the wrapper handles cleanup on unmount or when the series shape changes.
- **`web/src/pages/VMDetail.jsx`** — rewires the Metrics tab to use the live hook, adds four charts (CPU, Memory, Disk I/O `read`+`write`, Network `rx`+`tx`), and shows the `LiveIndicator` next to the Resource Utilization header.

### Compatibility shim

- **`web/index.html`** — tiny preflight script that sanitizes `navigator.language` when the embedded browser reports an `Intl`-invalid POSIX tag (e.g. `en-US@posix` from headless CI Chromium). Without this, uPlot's top-level `new Intl.NumberFormat(navigator.language)` throws at module-eval time and crashes the entire app — symptom in CI was every Playwright test waiting forever for the React tree to mount.

### Tests

- **`tests/web/mock-server.js`** — adds a `/stats/stream` SSE endpoint whose first frame mirrors the REST-seeded baseline (so existing baseline assertions still pass) and subsequent frames advance every 600 ms.
- **`tests/web/gui.spec.js`** — new Playwright assertions for the four chart containers, the Disk I/O multi-series legend, and the `live` status transition.
- **`web/tests/use-vm-stats.test.js`** — unit coverage for the helpers under `node --test`: ordering, dedup, cap, parseable-timestamp guard, and the uPlot columnar layout.

### Docs

- **`docs/ROADMAP.md`** — 4.1.7 marked ✅ Done with the implementation summary; priority table updated to point at the remaining 4.1.10 (deeper unit/integration/E2E coverage) as the only metrics gap.
- **`README.md`** — progress block regenerated by `scripts/roadmap-progress.sh`.

## Test plan

- [x] `make web` — builds clean (uPlot bundled, ~26 KB additional gzipped JS)
- [x] `make test-web` — Playwright mock GUI suite **28/28 pass** (including the four new chart/live-status assertions)
- [x] `node --test web/tests/use-vm-stats.test.js` — **7/7 pass** for the helper unit coverage
- [x] `node tests/web/run-built-web-regression.js` — built-bundle regression checks pass

## Notes

- The frontend bundle grows from ~274 KB → ~333 KB (≈108 KB gzipped). Most of that is uPlot itself; we vendor it via npm rather than a CDN so the embedded GUI keeps working without internet access.
- The hook reuses the existing `LiveIndicator` from 4.2.11 by mapping its own `STATS_STATE_*` strings onto the same `'live'/'reconnecting'/'fallback'` values, so the live pill on the Metrics tab is visually identical to the one on Dashboard / VM list.
- 4.1.10 (real-VM stress E2E) is intentionally **not** in this PR — it requires a running daemon + Rocky image, which the mock CI tier can't provide. Filing it as the remaining metrics gap.

---
_Generated by [Claude Code](https://claude.ai/code/session_019JrawN2BcZZzyxCCV2Fz3N)_


chinmay28 (commented):
Looks good — would approve if I weren't the PR author. Verified locally on `claude/eager-cerf-evRsH`:

- `npm run build` (in `web/`) — clean. Bundle reports `333.47 kB │ gzip: 107.96 kB` — matches the PR's "≈108 KB gzipped" claim. The +108 KB is mostly uPlot.
- `node --test web/tests/use-vm-stats.test.js` — **7/7 pass** (sampleTime / appendSample ordering / dedup / cap / parseable-timestamp guard / buildChartData columnar layout & gap handling / empty rails).

Architecture / design checks:

- `useVMStats` cleanly separates the initial REST fetch from the live SSE subscription; uses the same `?api_key=` browser-auth fallback pattern the daemon already accepts (`internal/api/sse.go`). Exponential backoff (1s/2s/4s capped at 30s) and 3-failure threshold to fallback mirror `useEventStream`.
- `MetricChart` correctly splits the create/destroy effect from the data-update effect, so live ingest doesn't recreate the uPlot instance per frame. ResizeObserver-driven width handling lifts the responsive concerns out of the chart parent.
- `vmStatsHelpers.js` is deliberately React-free so `node --test` runs without a DOM shim — same pattern future hooks should follow.
- The `web/index.html` `navigator.language` shim is a documented workaround for headless POSIX-locale Chromium crashing uPlot's `Intl.NumberFormat(navigator.language)` on module eval. Hacky but isolated and well-commented.
- Hook listens to both `event: vm.stats` (typed) and the default `message` channel — works against either server contract, consistent with the daemon's actual frame format documented in §5 of `docs/ARCHITECTURE.md`.

Non-blocking observations (carry over to a follow-up):

- **Same indefinite-fallback pattern as the original `useEventStream`** — once `useVMStats` enters fallback mode it stays in 10s polling indefinitely instead of retrying SSE after a bounded window. PR #177 is fixing this for `useEventStream`; once it lands, the same pattern can be applied here for parity. Not blocking — fallback mode keeps the chart updating, which is the user-visible contract.
- Roadmap update is accurate: 4.1.7 ✅ Done, 4.1.10 (real-VM stress E2E) correctly remains the only outstanding metrics gap.

LGTM.

---

PR #171, opened 2026-05-02
feat(vm): add VM restart action across manager, REST, CLI, and GUI

chinmay28:
## Summary

New, fully independent vertical slice. Adds a graceful **Restart** primitive at every layer of the lifecycle stack so operators can reboot a VM with one call instead of a stop+poll+start dance. Roadmap row **2.3.4** marked ✅ Done; the future scheduler `restart` action (5.2.5) now reuses the same manager method.

### Manager
- `vm.Manager.Restart(ctx, id)` added to the interface.
- `LibvirtManager.Restart`: graceful `Domain.Shutdown()`, polls for SHUTOFF up to a 30s `restartShutdownTimeout`, then `Domain.Destroy()` fallback, then `Domain.Create()`. Already-stopped VMs just get started.
- `MockManager.Restart`: stateful mock with a `RestartErr` injection hook.
- `quotaManager.Restart`: passthrough (no resource delta to validate).

### REST
- `POST /api/v1/vms/{vmID}/restart` returns `{"status":"restarted"}`.
- Publishes `vm.restart_requested` (`app` source) on success.
- Wired into `internal/api/router.go` next to `/start` and `/stop`.

### CLI
- `vmsmith vm restart <id>` (single-id; no `--all` in v1 — bulk restart can land later).

### GUI
- "Restart" button on VMDetail beside Stop, only visible while the VM is running.
- New `vms.restart()` method in the typed API client.

### Tests
- **Unit**: `TestMockManager_Restart`, `TestMockManager_Restart_NotFound`, `TestMockManager_Restart_ErrorInjection`.
- **API integration**: `TestRestartVM`, `TestRestartVM_Error` in `internal/api/api_test.go`.
- **CLI**: `TestCLI_VMRestart`, `TestCLI_VMRestart_NotFound` in `internal/cli/commands_test.go`.
- **Mock GUI (Playwright)**: `restart running VM` in `tests/web/gui.spec.js`. Mock server learns the new endpoint.

### Docs / spec
- `docs/openapi.yaml`: new `/vms/{vmID}/restart` path; `web/src/api/generated/schema.d.ts` regenerated.
- `docs/ROADMAP.md`: new §2.3.4 (✅) describing the vertical slice; §5.2.5 schedule restart action now points at `vm.Manager.Restart`.
- `CLAUDE.md` REST quick reference includes the new endpoint.

### Drive-by
- Removed a duplicate `MetricSample` definition from `docs/openapi.yaml` that was blocking `npm run generate:api`.

[...]

PR #174, opened 2026-05-02
feat(vm): auto-start VMs at daemon boot (2.5.1)

chinmay28:
## Summary

New, fully independent vertical slice. Adds an `auto_start` flag at every layer of the VM lifecycle so operators can mark VMs that should boot automatically when vmsmith starts up — eliminating the need for an external init script or systemd unit chain. Roadmap row **2.5.1** added and marked ✅ Done.

### Types
- `pkg/types/vm.go` — `AutoStart bool` on `VMSpec` (always serialised, no `omitempty` so clients can distinguish "off" from "field absent") and `*bool` on `VMUpdateSpec` (nil = no change, *false = explicit off).

### Manager
- `LibvirtManager.Update` honours `patch.AutoStart` and treats an AutoStart-only edit as a metadata-only change (no stop/restart, no domain XML regeneration).
- `MockManager.Update` — same semantics for tests.

### Daemon
- `internal/daemon/daemon.go` — `runAutoStartSweep` runs after metrics + scrape startup but before the HTTP server begins serving. Lists VMs, calls `Start` on any with AutoStart=true that are currently stopped, publishes `vm.auto_started` (app source) on success and `vm.auto_start_failed` (system / warn) on failure so the events stream reflects boot health. Sweep failures never block daemon startup.

### REST + OpenAPI
- `docs/openapi.yaml` — `auto_start` added to `VMSpec` and `VMUpdateSpec`.
- `web/src/api/generated/schema.d.ts` regenerated.

### CLI
- `vmsmith vm create --auto-start`
- `vmsmith vm edit --auto-start <true|false>` (uses `Changed("auto-start")` to distinguish "not provided" from explicit false)
- `vmsmith vm info` now prints `Auto-start: true|false`.

### GUI
- VMList Create modal — "Lifecycle" section in Advanced tab with the auto-start checkbox.
- VMDetail Edit modal — auto-start checkbox under the IP field; sends the new `auto_start` boolean only when toggled.
- VMDetail summary — new "Auto-start at boot" info card showing On/Off.

### Tests
- **Unit** (`internal/vm/mock_manager_test.go`) — `TestMockManager_Update_AutoStart` covers enable / disable / no-change flows.
- **Daemon** (`internal/daemon/daemon_test.go`) — `TestRunAutoStartSweepStartsOnlyMarkedStoppedVMs` + `TestRunAutoStartSweepReportsFailures`.
- **API integration** (`internal/api/api_test.go`) — `TestCreateVM_AutoStartFlag` + `TestUpdateVM_AutoStartToggle`.
- **Mock-server** (`tests/web/mock-server.js`) — round-trips `auto_start` on create + PATCH so the UI tests below cover the wire format.
- **Mock GUI runner** (`tests/web/run-gui-tests.js`) — two new scenarios: the create modal exposes the checkbox on the Advanced tab, and the detail page exposes both the summary card and the edit toggle.
- **Playwright spec** (`tests/web/gui.spec.js`) — equivalent assertions for the `npx playwright` runner.

### Docs
- `CLAUDE.md` REST quick reference — `auto_start` added to create + patch bodies.
- `docs/ROADMAP.md` — new **§2.5 Daemon-managed VM Lifecycle** with 2.5.1 marked ✅ Done.

## Test plan

- [x] `CGO_ENABLED=1 go test -tags libvirt_dlopen -race -count=1 -short ./...` — all packages green
- [x] `make web` — builds clean
- [x] `node tests/web/run-gui-tests.js` — **30/30** mock GUI tests pass
- [x] `node tests/web/run-built-web-regression.js` — clean

## Notes

- The auto-start sweep is launched as a goroutine so it doesn't delay HTTP server startup. In practice, on a fresh boot the sweep will overlap with the first second of API availability — that's intentional. Operators querying VM state during the sweep will see VMs transition from `stopped` → `running` as the sweep progresses, mirroring the natural state of the system.
- The `omitempty` removal on `VMSpec.AutoStart` is deliberate: a Go bool's zero value is `false`, and `omitempty` would have erased the field from JSON when it was set to `false`, making clients unable to distinguish "off" from "field absent". `VMUpdateSpec.AutoStart` keeps `omitempty` because it's a `*bool` and `nil` legitimately means "leave unchanged".
- This PR is independent of #168 (webhooks), #170 (live charts), and #171 (VM restart) — touches different code paths and types.

https://claude.ai/code/session_01TF5biD4nAVRQoErTYnvEP3

---
_Generated by [Claude Code](https://claude.ai/code/session_01TF5biD4nAVRQoErTYnvEP3)_

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -414,10 +414,10 @@ Additional docs routes:
 
 ```
 GET    /vms                            List all VMs (`?tag=<tag>` and `?status=<state>` filters supported); CLI also supports local `--limit` / `--offset` pagination on `vmsmith vm list`
-POST   /vms                            Create VM (VMSpec JSON body: name, image, cpus, ram_mb, disk_gb, ssh_pub_key, default_user, networks; VM names must be unique, 1-64 chars, alphanumeric/hyphen)
+POST   /vms                            Create VM (VMSpec JSON body: name, image, cpus, ram_mb, disk_gb, ssh_pub_key, default_user, networks, auto_start; VM names must be unique, 1-64 chars, alphanumeric/hyphen). When `auto_start=true`, the daemon will start this VM automatically at boot via the auto-start sweep.
 GET    /vms/{id}                       Get VM
 POST   /vms/{id}/clone                 Clone VM (body: `{ "name": "clone-name" }`; validates the new name and returns the cloned VM in stopped state)
-PATCH  /vms/{id}                       Update VM resources (VMUpdateSpec: cpus, ram_mb, disk_gb, nat_static_ip, nat_gateway — zero/empty ignored; disk grow-only; IP change updates DHCP reservation + regenerates cloud-init ISO with new instance-id)
+PATCH  /vms/{id}                       Update VM resources (VMUpdateSpec: cpus, ram_mb, disk_gb, nat_static_ip, nat_gateway, auto_start — zero/empty ignored; `auto_start` uses a `*bool` so omit it to keep the current value; disk grow-only; IP change updates DHCP reservation + regenerates cloud-init ISO with new instance-id)
 POST   /vms/{id}/start                 Start VM
 POST   /vms/{id}/stop                  Stop VM
 DELETE /vms/{id}                       Delete VM
--- a/README.md
+++ b/README.md
@@ -18,14 +18,14 @@ CLI tool, HTTP REST server, and embedded web GUI for provisioning and managing Q
 ## Roadmap progress
 
 <!-- progress:start -->
-**Overall:** 95 / 138 tasks complete (68.8%)
+**Overall:** 96 / 139 tasks complete (69.1%)
 
-`▓▓▓▓▓▓▓▓▓▓▓▓▓▓░░░░░░` 68.8%
+`▓▓▓▓▓▓▓▓▓▓▓▓▓▓░░░░░░` 69.1%
 
 | Phase | Title | Done | Total | % |
 | ----- | ----- | ---- | ----- | - |
 | 1 | Foundation & Quality (Week 1-2) | 17 | 18 | 94.4% |
-| 2 | Core Feature Additions (Week 3-5) | 19 | 21 | 90.5% |
+| 2 | Core Feature Additions (Week 3-5) | 20 | 22 | 90.9% |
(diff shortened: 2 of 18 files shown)


chinmay28 (commented):
Looks good — would approve if I weren't the PR author. Verified locally on `claude/eager-cerf-nww64`:

- `CGO_ENABLED=1 go test -tags libvirt_dlopen -short -race -count=1 ./...` — all packages green, including the new `TestRunAutoStartSweepStartsOnlyMarkedStoppedVMs`, `TestRunAutoStartSweepReportsFailures`, `TestMockManager_Update_AutoStart`, `TestCreateVM_AutoStartFlag`, `TestUpdateVM_AutoStartToggle`.

Architecture / design checks:

- The `bool` vs `*bool` split between `VMSpec.AutoStart` and `VMUpdateSpec.AutoStart` is the right call. The PR justifies it correctly: dropping `omitempty` on `VMSpec.AutoStart` means clients can distinguish "off" from "field absent"; keeping the pointer on the patch makes "no change" unambiguous. Existing bbolt-stored VMs unmarshal cleanly to `AutoStart=false`.
- `runAutoStartSweep` correctly skips already-running VMs, only operates when `Spec.AutoStart=true`, never blocks startup (goroutine + log-and-continue), and emits `vm.auto_started` / `vm.auto_start_failed` so the events stream / Activity tab reflect boot health. The sweep is deliberately not gated on the HTTP server starting — overlap with first-second API requests is acceptable.
- `LibvirtManager.Update` correctly treats an AutoStart-only edit as metadata-only (no domain XML regeneration, no stop/restart), and falls through to the normal Update path when AutoStart is bundled with other resource changes. `MockManager.Update` mirrors this.
- The roadmap addition (new §2.5 "Daemon-managed VM Lifecycle") is a sensible sibling to §2.4 Templates and aligns with the broader operator-quality-of-life theme. Independent of #168/#170/#171 as advertised — touches different files.
- Frontend wiring is consistent: `normalizeSpec` is a passthrough, `!!spec.auto_start` defaults legacy responses cleanly, the edit modal only sends `auto_start` when the toggle actually changed.

Minor non-blocking nit: `TestRunAutoStartSweepReportsFailures` only asserts "must not panic". Could be tightened by subscribing to the bus and verifying `vm.auto_start_failed` arrives — useful follow-up but fine for this PR. LGTM.

---

PR #175, opened 2026-05-02
docs(roadmap): mark clone coverage complete

march26bot:
## Summary
- mark roadmap item 2.1.7 done now that clone API integration and Playwright coverage are both on 
- update the roadmap status line to call out the remaining clone gap as the libvirt-backed implementation itself
- tighten the priority summary so cloning only points at the still-open libvirt task

## Testing
- git diff --check

PR #177, opened 2026-05-03
feat(events): bound live-update polling fallback window

march26bot:
## Summary
- retry the event SSE hook with a bounded 30s polling fallback window after repeated failures
- keep reconnect and fallback cleanup explicit so the hook re-enters SSE mode cleanly
- sync the roadmap to mark clone E2E coverage and the completed 4.2.11 live-update behavior

## Testing
- `node --check ./web/src/hooks/useEventStream.js`
- `git diff --check HEAD~1 HEAD`

## Notes
- The frontend build toolchain is not installed in this maintenance worktree, so `npm --prefix web run build` could not be run here (`vite: not found`).


chinmay28 (changes requested):
The bounded fallback window is the right behavior per roadmap 4.2.11, and the exit-timer/reset-counter loop is correctly structured. A few issues to address before approving:

**Regression in `EventSource` constructor catch (must fix).** Old code:

```js
try { es = new window.EventSource(url); } catch { scheduleReconnect(); return; }
```

New code:

```js
try { es = new window.EventSource(url); } catch { return; }
```

When `EventSource` construction throws (CSP rejection, invalid URL, etc.), the old hook would back off and retry; the new hook silently dies with `status` stuck at `STATE_CONNECTING` and no recovery path. Please retain a fallback or reconnect on this branch — e.g., bump `failureCountRef`, trigger the same fallback path you use in `onerror`, or just call the same exit-timer recovery.

**Code duplication of the polling block.** The same ~25 lines of `poll()` + interval set-up are now copy-pasted into `es.onerror` and the `useEffect` "no EventSource" branch. The original `startFallback` callback existed to avoid this. Please extract a single `startFallback({ withExitTimer })` (or similar) and call it from both sites — keeps the bounded-window logic in one place and prevents drift.

**No test for the bounded behavior.** The Test plan section lists only `node --check` (syntax). The existing Playwright "indicator falls back to polling when SSE returns 410" only asserts the initial fallback transition. Please add at least:

- a Playwright assertion that after 30s of fallback the status returns to `connecting`/`live` once SSE is healthy again (mock-server can flip the 410 off after a short delay); or
- a `node --test` unit test using a fake `EventSource` + fake timers that drives the failure path and asserts the exit-timer re-runs `connect`.

**Roadmap row 2.1.7 update.** Marking 2.1.7 (clone E2E) Done is duplicated by #175 (which is scoped to that change against current main). Consider dropping the 2.1.7 / cloning portions of the ROADMAP.md diff and let #175 own that update; this PR can stay focused on 4.2.11. Either order merges cleanly, but keeping each PR scoped makes the history easier to read.

Once the catch-handler regression and at least one functional test are in, happy to re-review.

---

chinmay28 (approved):
Verified the bounded-fallback behaviour against the roadmap spec for 4.2.11 and the existing `useEventStream` hook on main.

**Architecture fit**
- After 3 consecutive SSE failures, the hook flips to `STATE_FALLBACK` and polls `/api/v1/events?until=<lastSeen+1>&per_page=50` every 10s for 30s, then re-attempts SSE — exactly matching the roadmap's "30s degraded-mode window with bounded fallback" wording. This is the behaviour that was missing on main, where the fallback was unbounded.
- Extracting `pollRecentEvents` and `startEventFallback` into `eventStreamHelpers.js` (with injectable `fetchImpl` / `setIntervalFn` / `setTimeoutFn`) is exactly the right shape for testable timer-driven hooks. The unit test in `web/tests/use-event-stream.test.js` exercises both helpers without needing a DOM or jsdom.
- The "no `EventSource`" path (e.g. server-side render or unsupported browser) now falls into a permanent fallback (no exit timer via `withExitTimer: false`), where the previous version called the helper that wasn't defined yet at that point. Cleaner.
- Cleanup correctly tears down both timers on unmount and on disabled mode, so navigating away while in fallback can't leak intervals.

**Tests**
- Unit tests cover the cursor-replay-only-newer-events behaviour and the start → poll → reconnect cycle of the fallback window.
- The existing Playwright "indicator falls back to polling when SSE returns 410" spec in `tests/web/gui.spec.js` (line ~720) still asserts `data-status` matches `/reconnecting|fallback/`, so the GUI side remains green. CI is green.

**Roadmap doc**
Status banner update + 4.2.11 row update accurately reflect the new behaviour, and the priority table no longer carries 4.2.11 in the "remaining work" cell.

**One small observation** (non-blocking): the `startEventFallback({...})` call site is duplicated three times inside the hook with the same `poll` / `onExit` / refs. Could be folded into a closure inside `useEffect`, but the current shape is correct and readable. Happy to merge as-is.

---

PR #181, opened 2026-05-03
feat(images): description + tag metadata across API, CLI, and GUI (2.2.7)

chinmay28:
## Summary

Mirrors the existing VM tags/description story onto images so operators can annotate, organise, and filter their image library the same way they do VMs. New roadmap row **2.2.7** marked ✅ Done.

### Types
- `pkg/types/image.go` — `Image` gains `Description string`, `Tags []string`, and `UpdatedAt time.Time` (`omitempty` so existing payloads keep their shape). New `ImageUpdateSpec` with documented PATCH semantics: empty `description` = no change; `nil` `tags` = no change; `[]` explicitly clears the tag set.

### Storage
- `internal/storage/image.go` — `CreateImage` and `ImportImage` now take `CreateImageOptions{Description, Tags}` and persist them. New `UpdateImage(id, ImageUpdateSpec) (*Image, error)` applies the patch in place.

### REST + OpenAPI
- `internal/api/handlers_image.go` — `CreateImage` and `UploadImage` validate the description (1024-char cap) and run `tags` through the same `normalizeTags` helper as VM tags. New `UpdateImage` handler powers `PATCH /api/v1/images/{id}` and emits `image.updated`. `ListImages` accepts `?tag=` (case-insensitive). Multipart upload also accepts a `description` field plus repeated `tags` form fields **or** a single comma-separated value.
- `internal/api/router.go` — registers `PATCH /images/{id}`.
- `docs/openapi.yaml` — extended `Image`, new `UpdateImageRequest`, `?tag=` query param on the list, multipart `description`/`tags` fields, and the new PATCH path. `web/src/api/generated/schema.d.ts` regenerated via `npm run generate:api`.

### CLI
- `vmsmith image create --description --tag <t>` (repeatable) — persists metadata.
- `vmsmith image edit <id> --description --tag <t>` — single-image PATCH; refuses to run when neither flag is set.
- `vmsmith image list --tag <t>` — case-insensitive client-side filter; the table now has a `TAGS` column.

### GUI
- `web/src/pages/ImageList.jsx` — tag-filter chips, inline description preview, per-row tag badges, an Edit modal for description/tags, plus description + tag inputs on the existing Upload modal.
- `web/src/api/client.ts` — typed wrappers for the new fields and `?tag=` filter; `images.update(id, patch)` and `uploadImageWithProgress` carry the new metadata.

### Docs
- `CLAUDE.md` REST quick reference reflects the new endpoint + form fields.
[...]

PR #179, opened 2026-05-03
feat(webhooks): Settings GUI + send-test-event endpoint (4.2.16)

chinmay28:
## Summary

New vertical slice that lights up the webhook UX on top of the 4.2.15 backend.

- **Settings page** (`web/src/pages/Settings.jsx`) lists registered webhooks with their persisted `last_delivery_at` / `last_status` / `last_error`, lets operators add and delete hooks, and exposes a per-row "Test" button that surfaces an inline success/failure verdict.
- **`POST /api/v1/webhooks/{id}/test`** synthesises a `system.webhook_test` event, delivers it once (no retries — UI wants a quick answer), records the outcome on the webhook's `LastDelivery*` fields, and returns a `WebhookTestResult` with `success` / `status_code` / `duration_ms` / `error` / `event_id`. The synthetic event is **not** published on the bus, so it doesn't fan out to other subscribers.
- **`Manager.TestDeliver`** reuses the same SSRF deny-list and pinned-dialer path as production deliveries. `ErrWebhookNotFound` → 404, otherwise the result is always 200 (inspect the body) so the UI gets one consistent success/error state machine.
- Roadmap row **4.2.16** marked ✅ Done; README progress regenerated (99 / 139 tasks complete, 71.2%).

### Backend
- `pkg/types/webhook.go` — new `WebhookTestResult` type.
- `internal/webhooks/manager.go` — `TestDeliver(ctx, id)` does one signed POST through the SSRF-pinned client, persists `LastDelivery*` via `recordExternalDeliveryResult`, returns a `WebhookTestResult`. New `ErrWebhookNotFound` sentinel.
- `internal/api/handlers_webhook.go` — `WebhookTester` interface, picked up automatically by `SetWebhookSubsystem` when the runtime manager satisfies it. `TestWebhook` handler maps `ErrWebhookNotFound` → 404 and returns 503 `webhook_test_unavailable` when no tester is wired (e.g., test servers without the runtime manager).
- `internal/api/router.go` — registers `POST /api/v1/webhooks/{id}/test`.
- Daemon wiring is unchanged: the existing `*webhooks.Manager` automatically satisfies `WebhookTester`.

### Frontend
- `web/src/pages/Settings.jsx` — list / add / delete / test webhook UI, modal create form, inline status indicators (`CheckCircle2` / `AlertCircle`), shows `last_status` + `last_error` from the daemon when no local probe has run yet.
[...]

PR #178, opened 2026-05-03
feat(vm): delete-protection lock flag (2.5.2)

chinmay28:
## Summary

New, fully independent vertical slice. Adds a per-VM `locked` flag that guards against accidental deletion across the manager, REST, CLI, and GUI. Stop/start/restart still work — Lock is a delete guard, not a freeze. Roadmap row **2.5.2** added and marked ✅ Done.

### Types
- `pkg/types/vm.go` — `Locked bool` on `VMSpec` (always serialised, no `omitempty` so clients can distinguish "unlocked" from "field absent") and `*bool` on `VMUpdateSpec` (nil = no change, *false = explicit unlock).

### Manager
- `LibvirtManager.Update` honours `patch.Locked` and treats a Locked-only edit (or a Locked+AutoStart-only edit) as metadata-only — no stop/restart and no domain XML regeneration. The earlier AutoStart-only fast path was generalised to a single `metadataOnly` branch.
- `LibvirtManager.Delete` returns `types.NewAPIError("vm_locked", ...)` when `Spec.Locked=true`, before touching libvirt or the disk directory.
- `MockManager.Update` and `MockManager.Delete` mirror the same semantics.

### REST + OpenAPI
- `internal/api/validation.go` — `statusForAPIError` maps `vm_locked` → HTTP 409. The DeleteVM/UpdateVM/CreateVM handlers already route through this helper, so no handler-level changes were needed.
- `docs/openapi.yaml` — `locked` added to `VMSpec` and `VMUpdateSpec`.
- `web/src/api/generated/schema.d.ts` regenerated.

### CLI
- `vmsmith vm lock <id>` and `vmsmith vm unlock <id>` — single-id toggles via `Manager.Update` with `*bool`.
- `vmsmith vm create --locked` — born-locked VMs.
- `vmsmith vm edit --locked` — uses `Changed("locked")` to distinguish "not provided" from explicit false.
- `vmsmith vm info` and create/edit summary now print `Locked: true|false`.

### GUI
- `web/src/pages/VMList.jsx` — small "locked" badge with a `Lock` icon next to the status pill when `spec.locked=true`.
- `web/src/pages/VMDetail.jsx` — new "Delete protection" info card alongside "Auto-start at boot"; edit modal now includes a Lock checkbox under the auto-start checkbox; sends `locked` only when toggled.

### Tests
[...]

PR #184, opened 2026-05-04
test(web): cover clone validation failure path

march26bot:
## Summary
- add Playwright coverage for clone-name validation failures in the VM detail flow
- keep the user on the source VM when clone validation fails and assert the API error is shown
- sync the roadmap note for clone test coverage to mention the daemon/API and GUI failure-path coverage

## Testing
- npx playwright test tests/web/gui.spec.js --grep "clone VM"

Closes #182

PR #185, opened 2026-05-04
feat(snapshots): description metadata across API, CLI, and GUI (2.2.8)

chinmay28:
## Summary

Mirror the existing VM tags/description story onto VM snapshots so operators can annotate point-in-time captures the same way they do VMs and images. Persist the value through libvirt's `<description>` element so it round-trips through `dumpxml`. While in there, parse `creationTime` out of the snapshot XML so list responses now carry an accurate `created_at` timestamp instead of the zero value. New roadmap row **2.2.8** marked ✅ Done.

Also fixes the unresolved merge conflict markers (`<<<<<<< HEAD … >>>>>>>`) currently shipped in `docs/ROADMAP.md`'s "Suggested Next Priority Order" table so the rendered roadmap reads cleanly again.

### Types
- `pkg/types/snapshot.go` — `Snapshot` gains `Description string` (`omitempty`); new `SnapshotSpec` (`Name`, `Description`) carries create payload through the manager.

### Manager
- `vm.Manager.CreateSnapshot` now takes `types.SnapshotSpec` instead of a raw `name string`. `LibvirtManager.CreateSnapshot` encodes a typed `<domainsnapshot>` document via `encoding/xml` (so descriptions are XML-escaped automatically) and `ListSnapshots` parses both description and `creationTime` back out via the new `parseSnapshotXML` helper. `MockManager` and `quotaManager` mirror the new signature.

### REST + OpenAPI
- `POST /api/v1/vms/{id}/snapshots` accepts an optional `description` field, validated to ≤1024 chars (new `invalid_description` error code). The new `maxDescriptionLength = 1024` constant lives next to the other API-boundary validation rules.
- `Snapshot` / `CreateSnapshotRequest` OpenAPI schemas updated; `web/src/api/generated/schema.d.ts` regenerated via `npm run generate:api`.

### CLI
- `vmsmith snapshot create --description …` (max 1024 chars) and a new `DESCRIPTION` column on `vmsmith snapshot list`.

### GUI
- VMDetail snapshot list renders the description inline beneath each snapshot name; the create modal exposes a multi-line `<textarea>` (`input-snap-description`, `maxLength=1024`).
- `snapshots.create(vmId, name, description?)` typed wrapper.

### Tests
- **Unit** (`internal/vm/snapshot_xml_test.go`) — description + creationTime parsing, trimming, malformed-XML error path, non-numeric creationTime, and round-trip through `xml.Marshal`.
- **API integration** (`internal/api/api_test.go`) — `TestCreateSnapshot_WithDescription` (round-trips via list) and `TestCreateSnapshot_RejectsLongDescription` (`invalid_description` 400).
- **CLI** (`internal/cli/commands_test.go`) — `TestCLI_SnapshotCreate_DescriptionFlag`, `TestCLI_SnapshotCreate_DescriptionTooLong`, and the existing `TestCLI_SnapshotList_WithSnapshots` updated for the new column.
- **Mock server** (`tests/web/mock-server.js`) — round-trips `description`, validates name + ≤1024-char description, seeds the existing snapshot with a description so the GUI tests have something concrete to assert on.
- **Playwright + JSDOM** (`tests/web/gui.spec.js`, `tests/web/run-gui-tests.js`) — `create snapshot with description`, `renders existing snapshot description`, and an updated `shows existing snapshots` scenario asserting the seeded description.

### Docs
- `docs/ROADMAP.md` — new row **2.2.8** marked ✅ Done; fixed three unresolved merge conflict markers (lines 732/736/740 on `main`) in the priority-order section.
- `CLAUDE.md` REST quick reference reflects the new payload + parsed `created_at`.
- `README.md` progress block regenerated by `scripts/roadmap-progress.sh` (101 / 140 tasks complete, 72.1%).

## Test plan

- [x] `CGO_ENABLED=1 go test -tags libvirt_dlopen -race -count=1 -short ./...` — all packages green
- [x] `npm --prefix web run build` — frontend bundle builds
- [x] `npm --prefix web run generate:api` — schema regenerated; new `description` field flows through
- [x] `node -c tests/web/{mock-server,run-gui-tests,gui.spec}.js` — syntax check on the touched JS test files (Chromium download is blocked in this sandbox; CI runs the actual Playwright suite)

## Notes

- **No overlap with current open work.** This PR is independent of #178 (VM lock flag), #179 (webhook test endpoint), and #181 (image description/tags) — touches a brand-new field on a different resource (snapshots), a separate manager method, and a separate VMDetail subsection.
- **Breaking interface change is contained to `vm.Manager.CreateSnapshot`.** Every implementation (`LibvirtManager`, `MockManager`, `quotaManager`) is updated in this PR; no external consumers exist outside this repo. The wire format (REST request body) is *additive* — old `{ "name": "..." }` requests still work because `description` is optional.
- **Description PATCH semantics.** Snapshots are immutable in libvirt (you cannot edit a snapshot's metadata after creation; you'd take a new one), so this PR intentionally does **not** add a PATCH endpoint. Mirrors the libvirt model.
- **Why not switch to `*string`?** An empty `description` is treated as "absent", same as the rest of the codebase's description fields. Switching to `*string` later is a non-breaking change if it ever becomes a need.
- **`creationTime` parsing.** Today's `ListSnapshots` returns snapshots with a zero `CreatedAt` because the libvirt timestamp is never read. The new `parseSnapshotXML` helper fixes that as a side-effect, which is also why the `created_at` field is now reliable in the GUI table.

https://claude.ai/code/session_01G5LQe9EXTk9ZSfzbj6DABj

---
_Generated by [Claude Code](https://claude.ai/code/session_01G5LQe9EXTk9ZSfzbj6DABj)_


chinmay28 (commented):
**Verdict: changes recommended before merge — core snapshot work is solid, but the diff bundles unrelated changes and has one file naming bug. (Cannot self-REQUEST_CHANGES; flagging here as a comment.)**

Reviewed end-to-end and ran the suite locally on this branch.

**Verified**
- `CGO_ENABLED=1 go test -tags libvirt_dlopen -count=1 -short ./...` — all packages pass
- `npm --prefix web run build` — frontend builds
- Spot-checked `parseSnapshotXML` against the 5 unit cases (description present, no description, whitespace-trimmed description, malformed XML, non-numeric `creationTime`) and the round-trip via `xml.Marshal` — coverage looks correct.

**Architecture alignment (the core change)**
- `vm.Manager.CreateSnapshot` interface taking `types.SnapshotSpec` instead of a raw name is the right signature change — matches the existing `VMSpec` / `VMUpdateSpec` precedent and gives room to add future fields (e.g. memory inclusion, child snapshot policy) without another interface break. All implementations (`LibvirtManager`, `MockManager`, `quotaManager`) updated in this PR.
- Using `encoding/xml` to marshal the `<domainsnapshot>` document instead of `fmt.Sprintf` is a real correctness/security fix — descriptions are now XML-escaped automatically. Worth calling out explicitly: this matters once a description contains `<`, `&`, or quotes.
- `parseSnapshotXML` filling in `entry.CreatedAt` from libvirt's `<creationTime>` is a genuinely useful side-effect — `ListSnapshots` previously returned a zero-value `CreatedAt`, so the GUI snapshot table will now actually show real timestamps.
- API description cap (`maxDescriptionLength = 1024`) lives next to the other API-boundary validators with a typed `invalid_description` error code — consistent with the surrounding pattern.
- 4-layer test coverage (unit XML parsing, API integration, CLI, mock-server + Playwright/JSDOM) is on par with #178/#181.

## Issues to fix before merge

### 1. `internal/vm/snapshot_xml_test_helper.go` is misnamed and unnecessary
The file ends in `.go`, not `_test.go`, so it ships in the regular package compile (the Go linker happens to dead-code-eliminate the symbol because nothing in the production code path calls it, but that's coincidence — the file is treated as production code by `go build`, `go vet`, golangci-lint, coverage tooling, etc.).

Worse, the function is a one-line wrapper around `xml.Marshal(doc)`:

```go
func marshalSnapshotXMLForTest(doc snapshotXMLDoc) (string, error) {
    out, err := xml.Marshal(doc)
    if err != nil {
        return "", err
    }
    return string(out), nil
}
```

…and it's used in exactly one place in `snapshot_xml_test.go:106`. Since the test file is in the same package, it can call `xml.Marshal(doc)` directly. Please either:
- Delete `snapshot_xml_test_helper.go` and inline `xml.Marshal(doc)` in the test, or
- Rename to `snapshot_xml_export_test.go` (any `*_test.go` filename works) so the build system treats it as test-only.

### 2. Unrelated `web/src/hooks/useEventStream.js` rewrite (190+/76-)
The PR title is `feat(snapshots): description metadata across API, CLI, and GUI (2.2.8)`, but the diff also rewrites the SSE hook end-to-end (adds `FALLBACK_WINDOW_MS = 30_000` + `fallbackExitTimerRef`, inlines the previously shared `startFallback` callback into `onerror` and the SSE-unsupported branch). That's roadmap row 4.2.11, not 2.2.8.

PRs #183 and #184 carry the **identical** rewrite of `useEventStream.js` (verified `diff -q` against this branch returns no differences). All three must have inherited it from a common pre-#177 base. The change itself may be desirable (it does close the bounded-window gap that #177 left as "Partial"), but it does not belong in any of these three PRs and has zero JS-side test coverage for the new exit behaviour.

Please drop `web/src/hooks/useEventStream.js` from this PR — `git checkout main -- web/src/hooks/useEventStream.js` after rebase will surgically remove it.

### 3. Roadmap row 4.2.11 status edit is unrelated to 2.2.8
Same root cause: the diff flips 4.2.11 from "Partial" to "Done" and updates the priority table. That belongs in whichever PR finishes the bounded-window work, not here.

## Nits (non-blocking)

- `parseUnixSeconds` reimplements `strconv.ParseInt(strings.TrimSpace(s), 10, 64)`. The hand-rolled loop is fine but the stdlib call is shorter and gives identical "non-numeric → error" behaviour for free.
- `attrs["description"] = req.Description` in `CreateSnapshot` — the `description` attribute could push toward the event-attribute size budget if combined with other future attrs. Probably fine at the 1024 cap, but worth keeping in mind.
- The `Last updated:` date in `docs/ROADMAP.md` regresses 2026-05-02 → 2026-05-03 in the diff (the PR description says 2026-05-04). Cosmetic.

## What's right
- The 2.2.8 row addition itself, the `<description>` round-trip via libvirt XML, the `creationTime` parsing fix, the validation/CLI/GUI/test coverage, the OpenAPI + generated TS schema regeneration, and the conflict-marker cleanup in `docs/ROADMAP.md` (which currently exists in main from a botched merge of #177) are all real wins.

---

chinmay28 (commented):
LGTM (cannot self-approve from this account; leaving as a comment review).

Reviewed locally: `go vet -tags libvirt_dlopen ./...` clean, `go test -race -count=1 -short ./...` green across every package. CI is also green.

Architecture/roadmap alignment:
- `vm.Manager.CreateSnapshot` taking `types.SnapshotSpec` rather than a raw `name string` is the right call — it gives us a place to add future fields (TTL, parent ID, kind) without another breaking interface change. Every implementation (`LibvirtManager`, `MockManager`, `quotaManager`) is updated in this PR, and the wire format stays additive (`description` is optional).
- `snapshotXMLDoc` uses `encoding/xml` for both encode and decode, which means descriptions with `&`, `<`, `>` round-trip safely without manual escaping. The unit tests in `internal/vm/snapshot_xml_test.go` cover the round-trip plus the malformed-XML, no-description, whitespace-trim, and non-numeric-creationTime edge cases.
- Side benefit: `parseSnapshotXML` now also reads `creationTime`, so `ListSnapshots` returns a real `CreatedAt` instead of a zero value — that's a correctness fix the PR description called out and the GUI will now render properly.
- `maxDescriptionLength = 1024` constant is the right shape — the snapshot path uses it via `validateCreateSnapshotRequest`. (Note: the image path still has its own hard-coded 1024 in `validateImageDescription`. Out of scope for this PR but a small follow-up could DRY the two together.)
- Description PATCH not added is consistent with libvirt's snapshot model (immutable). Documenting that decision in the PR body is helpful.
- Roadmap row 2.2.8 added and marked Done; the merge-conflict markers in the priority-order table are also fixed (worth the diff noise).

Tests cover the right surfaces: unit XML parsing, API integration (round-trip + 1025-char rejection -> `invalid_description`), CLI flag + table column, mock-server round-trip, Playwright + JSDOM scenarios.

One small note (non-blocking): `invalid_description` isn't listed in `statusForAPIError`'s switch in `internal/api/validation.go`. It works because `CreateSnapshot` passes `http.StatusBadRequest` explicitly to `writeAPIError`, but if a future handler relies on `statusForAPIError(err, fallback)` for an `invalid_description` value, it would get the fallback rather than 400. Worth adding to the switch alongside `invalid_name` / `invalid_image` / `invalid_spec` for consistency.

---

chinmay28:
Addressed prior review feedback in `76e352e`:

- Rebased onto current `main`. The earlier diff against `useEventStream.js` and the unrelated 4.2.11 status edit are gone — those changes are now part of `main` via PR #177, so a clean rebase resolves them.
- The merge-conflict markers previously visible in `docs/ROADMAP.md`'s "Suggested Next Priority Order" table are now removed (they remained on main from a botched merge of #177).
- Restored main's longer 2.1.7 row description that referenced `tests/e2e/test_api_vm_lifecycle.py` (regression from a stale rebase).
- Added `invalid_description` to `statusForAPIError`'s switch alongside `invalid_name` / `invalid_image` / `invalid_spec` so any future handler that uses the helper maps it to 400 consistently.
- Regenerated README progress block.
- The original review's note about `internal/vm/snapshot_xml_test_helper.go` being misnamed: that file does not exist on this branch — the helper was already inlined as `xml.Marshal(doc)` in `snapshot_xml_test.go:106`.

`CGO_ENABLED=1 go test -tags libvirt_dlopen -count=1 -short ./...` is green locally.

---

chinmay28 (commented):
**Review verdict: ✅ Looks good — would approve if I weren't the author.**

**Architecture alignment:** Roadmap row 2.2.8 mirrors the VM/image description story onto snapshots, with the right libvirt-native primitive (`<description>` in the domainsnapshot XML) so descriptions round-trip through `dumpxml` and survive daemon restarts. Encoding via `encoding/xml`'s `Marshal` (not raw `fmt.Sprintf`) means descriptions are properly XML-escaped — important since this is operator-supplied free text.

**Bonus side-effect:** `parseSnapshotXML` also extracts `creationTime`, fixing the long-standing zero `CreatedAt` in `ListSnapshots`. `creationTime` is libvirt-canonical (Unix seconds) so the parse + `time.Unix` is correct.

**Cross-PR consistency:** This PR introduces `maxDescriptionLength = 1024` and the `invalid_description` API error code. PR #181 (image description) currently uses `invalid_spec` — once one of these lands, the other should rebase to use the new shared constant + `invalid_description` code so the error vocabulary is uniform.

**Roadmap markers cleanup:** This PR also resolves the unfinished merge-conflict markers from #177 in the priority-order table. PR #192 does the same — whichever lands first wins; the other will rebase clean.

**Interface change is contained:** `vm.Manager.CreateSnapshot(ctx, vmID, name)` → `(ctx, vmID, types.SnapshotSpec)`. All three implementations (`LibvirtManager`, `MockManager`, `quotaManager`) updated, plus all callers (CLI, API). Wire format remains additive (`description` is optional) so existing API consumers aren't broken.

**Test coverage:** Snapshot XML round-trip, no-description, trim, malformed XML, non-numeric `creationTime`, plus `parseUnixSeconds` unit table. API integration covers `WithDescription` round-trip + `RejectsLongDescription`. CLI covers create/edit/list. Playwright/JSDOM cover create-with-description and rendering. CI green.

**Minor nits (not blocking):**

1. `parseUnixSeconds` reimplements what `strconv.ParseInt(strings.TrimSpace(s), 10, 64)` already does. The hand-rolled version is correct (rejects non-digits, rejects `12a3`) but is needless surface area. `strconv.ParseInt` would also handle leading `+`/`-` which you may want to preserve as rejection — easy enough to add.

2. The CLI's new `DESCRIPTION` column is appended after `CREATED`, but the existing list table is fine; just be aware long descriptions in tabwriter will look ragged. Acceptable for a v1.

3. `validation.go`'s `statusForAPIError` adds `invalid_description` to the 400 group — good. Worth a follow-up to also add `invalid_port_forward` (introduced in #192) to the same line so all "invalid_*" codes route through 400 by default.

---

chinmay28 (commented):
**Verdict: LGTM with one small nit.**

Solid mirror of the description story onto snapshots, with a nice drive-by fix:

- `vm.Manager.CreateSnapshot` signature change (`name string` → `types.SnapshotSpec`) is contained — `LibvirtManager`, `MockManager`, `quotaManager` are all updated in this PR; no external consumers exist outside the repo. The wire format on `POST /api/v1/vms/{id}/snapshots` is *additive* (just an optional `description` field), so old `{"name": "..."}` clients keep working.
- Using `encoding/xml` for the `<domainsnapshot>` document gets correct XML escaping for free — operator descriptions can contain `<`, `&`, etc. without breaking libvirt parsing. Much better than `fmt.Sprintf` with raw interpolation.
- The drive-by `parseSnapshotXML` helper that extracts `creationTime` is a real bug fix: today's `ListSnapshots` returns a zero `CreatedAt` because the libvirt timestamp is never read. The unit tests in `internal/vm/snapshot_xml_test.go` cover happy path, no description, malformed XML, non-numeric `creationTime`, trim-on-parse, and the round trip — thorough.
- Validation: `invalid_description` typed error code (1024-char cap), correctly mapped to HTTP 400 in `statusForAPIError`. The new `maxDescriptionLength = 1024` constant lives next to the other API-boundary rules and is nice and discoverable.
- Snapshot description PATCH intentionally not added (libvirt snapshots are immutable). Right call — mirrors libvirt's own model.
- Drive-by cleanup of the unresolved `<<<<<<< HEAD … >>>>>>> 6652e5d` markers in `docs/ROADMAP.md` is welcome; overlaps with PR #192 and PR #195, whichever lands first wins.

**One nit (non-blocking):** in `internal/api/handlers_snapshot.go`, the description is passed to `validateCreateSnapshotRequest` and into `SnapshotSpec` raw — no `strings.TrimSpace` at the API boundary. PR #192 (port forwards) trims at the boundary. Not a bug because `parseSnapshotXML` trims on read, but the round trip would persist `"  hello  "` into libvirt XML and return `"hello"` on list — a minor inconsistency worth tightening. A `req.Description = strings.TrimSpace(req.Description)` before validation would do it.

**Conflict alert:** PR #198 (snapshots bulk delete) calls `mockMgr.CreateSnapshot(nil, "vm-bd", "snap-name")` (the old signature). When this PR lands first, #198 will need to update its test calls to the new `SnapshotSpec` signature; if #198 lands first, its tests will fail to compile against this PR. Trivial to fix on whichever rebases.

The README progress block also overlaps with PRs #178, #181, #192, #194, #198 — the regen will need a rebase pass on each.

---

chinmay28 (commented):

Reviewed against the architecture and 2.2.8. CI green; locally `TestParseSnapshotXML_*`, `TestSnapshotXMLDoc_RoundTrip`, `TestCreateSnapshot_WithDescription`, `TestCreateSnapshot_RejectsLongDescription`, `TestCLI_SnapshotCreate_DescriptionFlag`, `TestCLI_SnapshotCreate_DescriptionTooLong` all pass.

Strengths:

- **Switched the snapshot XML write from `fmt.Sprintf` to `encoding/xml`** (`internal/vm/lifecycle.go` `snapshotXMLDoc`). This is more than a cosmetic refactor — operator-supplied descriptions could otherwise inject XML (`</description><disk file='...'>`-style payloads) into `dom.CreateSnapshotXML`. Marshalling through a typed struct escapes everything correctly. Worth calling out as a small security-hygiene win on top of the feature.
- **`parseSnapshotXML` side-effect fix.** Today's `ListSnapshots` returns zero `CreatedAt` because libvirt's `creationTime` is never read; this PR fixes that. Good catch.
- **Validation as a shared constant.** `maxDescriptionLength = 1024` lives next to the other API-boundary rules, with a new `invalid_description` API error code mapped to 400. Consistent with how image descriptions get capped in #181.
- **Right call to *not* add a PATCH endpoint.** Libvirt snapshots are immutable post-create — re-snapshotting is the supported path. The PR description correctly explains this trade-off.
- **Manager interface change is contained.** `LibvirtManager`, `MockManager`, and `quotaManager` all updated; the wire change is additive (`description` is optional, old `{ "name": "..." }` requests still work).
- **Coverage is broad** — XML parser unit tests cover the happy path, missing description, whitespace trim, malformed XML, non-numeric `creationTime`, and round-trip. Nice to see the negative cases.

Non-blocking observations:

1. **Overlap with PR #195.** Both PRs delete the merge-conflict markers in `docs/ROADMAP.md`'s "Suggested Next Priority Order" table. The wordings of the resolved Events / VM Cloning rows are nearly identical between the two but not byte-identical (e.g., this PR's "Main clone flows **and their** integration/E2E coverage ship today" vs #195's "Main clone flows **plus** integration/E2E coverage ship today"). Whichever PR lands second will need a small rebase. Not a blocker — just a heads-up so the second author knows it's a textual conflict, not a logical one.
2. `parseSnapshotXML` returns `(string, time.Time, error)` and silently swallows malformed XML in the call site (`if rawXML, xmlErr := s.GetXMLDesc(0); xmlErr == nil { … parseErr == nil … }`). That's a reasonable best-effort posture for `ListSnapshots`, but consider a single `logger.Debug("vm", "snapshot xml parse failed", ...)` so operators can diagnose if descriptions stop appearing for a corrupted snapshot.
3. The new `description` attribute is emitted alongside `snapshot.created` events (`s.publishAppEvent("snapshot.created", vmID, …, attrs)`). Future webhook consumers will see this — fine, just confirming the wire shape is intentional.

Roadmap row 2.2.8 correctly marked Done. README progress reconciles.

---

PR #192, opened 2026-05-05
feat(network): description metadata for port forwards (2.2.9)

chinmay28:
## Summary

Mirror the existing VM/image/snapshot description story onto port forwards so operators can label rules ("ssh-jumpbox", "metrics scrape", "web") without having to remember which port maps to what. New roadmap row **2.2.9** marked ✅ Done.

Drive-by cleanup: removes the unresolved merge-conflict markers (`<<<<<<< HEAD … >>>>>>>`) currently sitting in `docs/ROADMAP.md`'s "Suggested Next Priority Order" table on `main` (left over from the bounded-fallback merge in #177).

### Types
- `pkg/types/network.go` — `PortForward` gains `Description string` with `omitempty`, so existing API consumers see no payload-shape change.

### Manager
- `internal/network/portforward.go` — `PortForwarder.Add` now takes `AddOptions{Description}` instead of a bare arg list, so future fields (e.g. `tags`) can land without another signature break. The description is persisted alongside the rule.

### REST + OpenAPI
- `internal/api/handlers_network.go` — `addPortRequest` accepts `description`. Trimmed and validated to ≤256 chars (a tighter cap than the 1024-char VM/image/snapshot limit because port-forward descriptions are intended to be one-line labels). Over-long descriptions return HTTP 400 + typed `invalid_port_forward` error code.
- `internal/api/validation.go` — new `maxPortForwardDescriptionLength = 256` constant + `validatePortForwardDescription` helper next to the existing API-boundary validation rules.
- `docs/openapi.yaml` — `description` added to `AddPortRequest` (with `maxLength: 256`) and `PortForward`. `web/src/api/generated/schema.d.ts` regenerated via `npm run generate:api`.

### CLI
- `vmsmith port add --description …` (max 256 chars, fails fast before any VM-manager init).
- `vmsmith port list` adds a new `DESCRIPTION` column.

### GUI
- `web/src/pages/VMDetail.jsx` — the port-forward card now renders the description inline beneath each rule when present (`port-description-<id>` testid), and the Add Port modal exposes a labeled text input (`input-port-description`, `maxLength=256`).
- `web/src/api/client.ts` — `ports.add(...)` typed wrapper now accepts an optional description.

### Tests
- **Unit** (`internal/network/portforward_test.go`) — `TestPortForwarder_Add_PersistsDescription` and `TestPortForwarder_Add_DescriptionOmittedWhenEmpty`. Existing tests updated for the new `AddOptions` arg.
[...]

PR #191, opened 2026-05-05
feat(store): add schedule persistence primitives

march26bot:
$## Summary\n- add schedule, schedule_run, and schedule_meta bolt buckets plus schedule CRUD helpers\n- persist per-schedule run history with bounded retention and newest-first listing\n- add store tests for schedule CRUD, run retention, and last_tick round-trips\n\n## Testing\n- not run (Go toolchain is unavailable in this maintenance workspace)

PR #190, opened 2026-05-05
security: redact sensitive query params in request logs

march26bot:
## Summary
- redact `ticket` and `api_key` values before request logging stores the raw query string
- add middleware coverage proving secrets are replaced with `REDACTED` while non-sensitive params remain visible
- track the roadmap follow-up in #189

## Testing
- Not run in this environment (`go` and `gofmt` are unavailable in the sandbox)

PR #194, opened 2026-05-06
feat(version): build identification across API, CLI, and GUI (6.2.5)

chinmay28:
## Summary

New, fully independent vertical slice. Adds a `pkg/version` package with linker-injected `Version`, `Commit`, and `BuildDate` vars and surfaces the build identification through three coordinated entry points (REST, CLI, GUI). Roadmap row **6.2.5** added and marked ✅ Done.

The previous `Makefile` already passed `-X main.version=$(VERSION)` to the linker, but `main.go` never declared a `version` variable — so the flag was a silent no-op. This PR fixes the wiring and makes the build identification visible everywhere it's useful.

### Types
- `pkg/types/version.go` — new `BuildInfo` wire type (`version`, `commit`, `build_date`, `go_version`, `os`, `arch`).
- `pkg/version/version.go` — new package with linker-set `Version`/`Commit`/`BuildDate` vars and an `Info()` accessor that fills in `runtime.Version`/`GOOS`/`GOARCH`. Defaults are deliberately human-readable (`"dev"` / `"unknown"` / `"unknown"`) so an unconfigured build still returns sensible JSON.

### REST + Build wiring
- `internal/api/handlers_version.go` — `GetVersion` returns `BuildInfo` as JSON with `Cache-Control: no-store`.
- `internal/api/router.go` — registers `GET /api/version` **outside** the authenticated `/api/v1` tree, mirroring `/api/docs` and `/api/openapi.yaml`. Health checks, monitoring agents, and the GUI footer can read it before any API key is entered.
- `Makefile` — `LDFLAGS` now passes `-X github.com/vmsmith/vmsmith/pkg/version.Version=$(VERSION) -X $(VERSION_PKG).Commit=$(COMMIT) -X $(VERSION_PKG).BuildDate=$(BUILD_DATE)`. `COMMIT` and `BUILD_DATE` are computed alongside `VERSION` so release builds carry the full identification.

### CLI
- `vmsmith version` — human-readable summary (`vmsmith <version>` + commit / built / go / platform lines).
- `vmsmith version --json` — same payload as `GET /api/version`, suitable for scripts.
- `vmsmith --version` — wired via cobra's built-in `Version` field so users get a one-shot version flag too.

### GUI
- `web/src/components/Layout.jsx` — fetches `/api/version` on mount and renders the live build label in the existing footer (`data-testid="layout-version"`) with commit/date in the title tooltip. Falls back to a static `VM Smith` label when the request fails (e.g., during transient daemon outages).
- `web/src/api/client.ts` — `system.version()` wrapper using a plain `fetch` (the typed OpenAPI client only covers `/api/v1/*`).

### Tests
- **Unit (`pkg/version`)** — `TestInfo_DefaultsAreSane` (defaults + runtime fields), `TestInfo_ReflectsLDFlagOverrides` (linker-injected values flow into `Info()`).
- **API integration (`internal/api`)** — `TestGetVersion_ReturnsBuildInfo` (response schema + `Cache-Control: no-store`), `TestGetVersion_DoesNotRequireAuth` (sanity-checks `/api/v1/vms` returns 401 without a key, then asserts `/api/version` returns 200 with no key).
- **CLI (`internal/cli`)** — `TestCLI_Version_HumanReadable` and `TestCLI_Version_JSON` cover both output paths with linker-injected fields.
- **Mock-server (`tests/web/mock-server.js`)** — serves `/api/version` with deterministic mock build info so the GUI tests can assert exact strings.
- **Playwright (`tests/web/gui.spec.js`)** — `renders the build version returned by /api/version` (label + tooltip) and `falls back to a static label when /api/version fails` (intercept + 500).
- **JSDOM runner (`tests/web/run-gui-tests.js`)** — `layout footer shows mock build version`.

### Docs
- `CLAUDE.md` REST quick reference — new `/api/version` row with full payload description and CLI mention.
- `docs/ROADMAP.md` — new row **6.2.5** marked ✅ Done.

## Test plan

- [x] `CGO_ENABLED=1 go test -race -count=1 -short ./...` — all packages green
- [x] `npm --prefix web run build` — frontend bundle builds (108.82 KB gzip)
- [x] `node -c tests/web/{mock-server,run-gui-tests,gui.spec}.js` — JS test files parse cleanly
- [x] `vmsmith version` and `vmsmith version --json` smoke-tested with `-ldflags -X ...=v9.9.9` overrides — fields populate correctly

## Notes

- **No overlap with current open work.** Touches a brand-new package, a new HTTP path, a new CLI subcommand, and the existing footer slot in `Layout.jsx`. No conflict with #178 (VM lock), #179 (webhook UI), #181 (image desc/tags), #185 (snapshot description), #190 (request log redaction), #191 (schedule store), or #192 (port forward description).
- **Why `/api/version` and not `/api/v1/version`?** Build info is benign and routinely consumed by load-balancer health checks, monitoring agents, and the embedded GUI footer before any API key has been entered. Putting it under `/api/v1` would force every operator to either disable auth for monitoring or burn an API key on a public-by-design endpoint. `/api/docs` and `/api/openapi.yaml` already follow this pattern.
- **Why not added to `docs/openapi.yaml`?** The OpenAPI spec is scoped to `/api/v1/*` (its `servers` block is `http://localhost:8080/api/v1`), and `/api/docs` / `/api/openapi.yaml` are also intentionally outside the spec for the same reason. CLAUDE.md picks up the documentation duty.
- **Drive-by `README.md` regen.** Running `scripts/roadmap-progress.sh` swept up four ✅ Done rows that exist in `docs/ROADMAP.md` on `main` but weren't reflected in `main`'s progress block — those previous PRs forgot to re-run the regen. Pure bookkeeping; no roadmap rows were toggled by this PR beyond the new 6.2.5 entry.
- **Default values are intentionally human-readable.** A binary built without `-ldflags` returns `version: "dev"`, `commit: "unknown"`, `build_date: "unknown"` instead of empty strings. Operators running `make build-go` locally still get a sensible response.

https://claude.ai/code/session_016k2JcnTo7p4CWYbW8wEG2g

---
_Generated by [Claude Code](https://claude.ai/code/session_016k2JcnTo7p4CWYbW8wEG2g)_


chinmay28 (commented):
**Verdict: LGTM.**

Solid build-identification slice. Notable design choices that all check out:

- `pkg/version` decoupled from `pkg/types/version.go` — linker overrides land on a tiny package with no transitive imports, so future linkage hygiene stays simple.
- `Makefile` correctly targets `pkg/version.{Version,Commit,BuildDate}` with full path; that fixes the silent no-op where the previous `-X main.version=$(VERSION)` flag never had a `version` symbol to land on.
- `/api/version` lives **outside** `/api/v1` so health checks and the GUI footer can read it without an API key — verified by `TestGetVersion_DoesNotRequireAuth` which sanity-checks `/api/v1/vms` returns 401 then asserts `/api/version` returns 200. This is the right call: forcing operators to either disable auth for monitoring or burn an API key on a public-by-design endpoint would be a UX regression.
- `Cache-Control: no-store` is set so old build labels don't get cached after a deploy.
- Defaults are intentionally human-readable (`"dev"` / `"unknown"` / `"unknown"`) — unconfigured local builds still produce a sensible JSON response.
- GUI `Layout.jsx` uses a plain `fetch` for `/api/version` (the typed OpenAPI client is `/api/v1/*`-scoped) and falls back to a static label when the request fails — covered by Playwright with a 500 intercept.
- `/api/version` deliberately omitted from `docs/openapi.yaml`, matching the existing `/api/docs` and `/api/openapi.yaml` precedent (the spec's `servers` block is `/api/v1`).

Tests cover unit defaults + ldflag overrides, API integration (response schema + cache header + 200-without-auth), CLI human-readable + JSON output, mock-server stub, and two Playwright scenarios (renders / fallback).

Note: this PR shares the regenerated README progress block with several other open PRs (#178, #181, #185, #192, #198); first one through forces a trivial rebase on the rest.

---

chinmay28 (commented):
LGTM with one nit and a heads-up on a merge conflict. Commenting because GitHub blocks self-approval.

Reviewed against the architecture and 6.2.5. CI green; locally `pkg/version`, `TestGetVersion_ReturnsBuildInfo`, `TestGetVersion_DoesNotRequireAuth`, `TestCLI_Version_HumanReadable`, and `TestCLI_Version_JSON` all pass.

Strengths:

- **Fixes a silent linker no-op.** The old `Makefile`'s `-X main.version=$(VERSION)` was being ignored because `main.go` never declared a `version` var. Re-pointing at `pkg/version.Version` (and adding `Commit`/`BuildDate`) actually makes the value visible at runtime.
- **`/api/version` outside `/api/v1` is the right call.** Build identification is benign and routinely consumed by load-balancer health checks and the GUI footer before any API key is in localStorage. `TestGetVersion_DoesNotRequireAuth` pins the contract — `/api/v1/vms` returns 401, `/api/version` returns 200, both without an API key. Mirrors the existing `/api/docs` and `/api/openapi.yaml` pattern.
- **Defaults are human-readable** (`"dev"` / `"unknown"` / `"unknown"`) rather than empty strings, so `make build-go` without ldflags still returns sensible JSON. Captured by `TestInfo_DefaultsAreSane`.
- **`Cache-Control: no-store` on the response.** Right call for a build-info endpoint — the value should always reflect the running binary, never a stale CDN/proxy cache.
- **Cobra wiring** picks up both `vmsmith --version` (via `rootCmd.Version`) and `vmsmith version` / `vmsmith version --json`. Two valid muscle-memory paths covered.
- **GUI footer fetches /api/version on mount with a graceful fallback** to a static `VM Smith` label when the request fails. Playwright covers the failure path via `page.route` interception — nice.

Heads-up:

- **Merge conflict on `README.md`.** Currently `mergeable_state: dirty` because main has advanced (PRs landed that regenerated the progress block). Pure mechanical conflict — running `scripts/roadmap-progress.sh` after rebase will reproduce the right numbers.

Non-blocking nit:

- **`BUILD_DATE := $(shell date -u +%Y-%m-%dT%H:%M:%SZ)`** changes on every `make build` invocation, so two `make build` runs of the same SHA produce different binaries. That breaks any reproducible-build downstream (and Go's content-based caching for the linked binary). Consider sourcing from `git log -1 --format=%cI HEAD` instead — same field, deterministic per commit. Won't matter for the GUI footer use case, but it's a low-cost upgrade if anyone ever wants to verify "is this the same binary I built last week?"

Roadmap row 6.2.5 correctly marked Done.

---

PR #198, opened 2026-05-07
feat(snapshots): bulk delete by names or prefix (2.3.5)

chinmay28:
## Summary

New, fully independent vertical slice. Adds **bulk snapshot deletion** so operators can clean up automated snapshot series in a single round-trip — addresses the long tail of "50 `auto-nightly-*` snapshots, delete all but the last 7" maintenance pain. Roadmap row **2.3.5** added and marked ✅ Done.

The endpoint is built on top of existing manager primitives (`ListSnapshots` + `DeleteSnapshot`) so neither `LibvirtManager` nor `MockManager` needed an interface change. The bulk-delete logic is testable end-to-end through the mock without any libvirt-side support.

### Backend

- `POST /api/v1/vms/{vmID}/snapshots/bulk_delete` accepts either `{names: [...]}` or `{prefix: "..."}` (exactly one). Mutual-exclusion + empty-payload rejection returns HTTP 400 `invalid_bulk_request`.
- Per-target results surface partial failures alongside successes (404 / network errors on individual snapshots don't abort the rest).
- Each successful target still emits `snapshot.deleted`, with `bulk="true"` in attributes so receivers can collapse fan-out into a single Activity-timeline entry if desired.

### CLI

- `vmsmith snapshot delete <vm-id> --prefix <s>` enumerates matching snapshots through the existing manager API and prints `OK`/`FAIL` per name with a summary log.
- The existing `--name` path is preserved; `--name` and `--prefix` are mutually exclusive (rejected at flag parse time).

### GUI

- `web/src/pages/VMDetail.jsx` snapshot card now exposes per-row checkboxes, a select-all/indeterminate toggle, and a "Delete selected" action driving the typed `snapshots.bulkDelete` client.
- After completion a `snap-bulk-result` line surfaces `X of Y succeeded` so partial failure is visible without trawling network tabs.

### Tests

- **API integration** (`internal/api/api_test.go`): `TestBulkDeleteSnapshots_ByNames`, `_ByPrefix`, `_PartialFailure`, `_EmptyRequestRejected`, `_BothNamesAndPrefixRejected`, `_PrefixNoMatchesEmptyResponse`.
- **CLI** (`internal/cli/commands_test.go`): `TestCLI_SnapshotDelete_NoFlags`, `_BothFlags`, `_PrefixDeletesAllMatching`, `_PrefixNoMatchPrintsMessage`.
- **Mock-server** (`tests/web/mock-server.js`): replays the bulk endpoint with proper 400 / partial-failure / prefix paths and seeds two `auto-2026-05-*` snapshots so the GUI tests have something concrete to act on.
- **Playwright** (`tests/web/gui.spec.js`): `bulk delete selected snapshots`, `bulk delete via select-all snapshots`.
- **JSDOM mock-GUI** (`tests/web/run-gui-tests.js`): `bulk-delete selected snapshots`.

### Spec / docs

- `docs/openapi.yaml` — new `/vms/{vmID}/snapshots/bulk_delete` path and `BulkDeleteSnapshotsRequest`/`Response`/`Result` schemas.
- `web/src/api/generated/schema.d.ts` — regenerated via `npm run generate:api`.
- `CLAUDE.md` REST quick reference — new row documents the endpoint plus CLI mention.
- `docs/ROADMAP.md` — new row **2.3.5** marked ✅ Done.

## Test plan

- [x] `CGO_ENABLED=1 go test -tags libvirt_dlopen -count=1 -short ./internal/api/... ./internal/cli/... ./internal/vm/... ./internal/storage/... ./internal/network/... ./internal/webhooks/...` — all green
- [x] Targeted `go test -run TestBulkDeleteSnapshots ./internal/api/...` — 6/6 pass
- [x] Targeted `go test -run TestCLI_SnapshotDelete ./internal/cli/...` — 5/5 pass
- [x] `npm --prefix web run build` — frontend bundle builds (109.08 KB gzip)
- [x] `npm --prefix web run generate:api` — OpenAPI schema picks up new path + types
- [x] `node -c tests/web/{mock-server,run-gui-tests,gui.spec}.js` — JS test files parse cleanly (Chromium download is blocked in this sandbox; CI runs the actual Playwright suite)

## Notes

- **No overlap with currently-open chinmay28 PRs.** Verified against #178 (vm lock — different field on VMSpec), #179 (webhook Settings GUI — separate page+endpoint), #181 (image desc/tags — different resource), #185 (snapshot description — only adds the description field on create, doesn't touch deletion), #192 (port forward description — different resource), #194 (build version info — orthogonal). The snapshot multi-select UI lives next to the existing per-row Restore/Delete buttons; no shared state with #185's create-modal description input.

- **Why prefix matching is server-side, not client-side.** If a future authorisation layer (3.1.5 RBAC) wants to enforce "operators can only delete snapshots they own", the server is the right place to compute the match set. Doing it client-side would force every consumer to re-implement the check. Keeping it on the server also avoids round-tripping a giant ID list when a VM has hundreds of automated snapshots.

- **Why the API returns 200 on partial failure, not 207.** Multi-status (HTTP 207) is technically more correct but adds friction to every consumer. The per-target `success` flag is precise enough; the existing bulk VM action endpoint (`POST /api/v1/vms/bulk`) already follows this convention and operators don't have to learn two different status semantics.

- **Why no `force=true` flag.** libvirt rejects deletes when a snapshot has children that haven't been deleted yet; that error rightly surfaces as a per-target failure rather than a silent data loss. Adding `force=true` would invite "deleted ten snapshots and got eleven I didn't expect" surprises.

- **Bulk events carry `bulk="true"`.** Receivers (Activity timeline, webhooks, future schedulers) can either render each `snapshot.deleted` as its own entry or collapse them. The decision stays at the consumer side; the server just records what happened.

https://claude.ai/code/session_01BvjVnva44U5smtEaeyuFRM

---
_Generated by [Claude Code](https://claude.ai/code/session_01BvjVnva44U5smtEaeyuFRM)_


chinmay28 (commented):
**Verdict: LGTM — pending one rebase against PR #185.**

Well-designed bulk endpoint:

- Built on existing manager primitives (`ListSnapshots` + `DeleteSnapshot`) — no `vm.Manager` interface change needed, so neither `LibvirtManager` nor `MockManager` had to grow new methods. That keeps the change tiny and testable end-to-end through the mock.
- Mutual exclusion of `names` / `prefix` returns HTTP 400 `invalid_bulk_request` with both halves of the contract covered (`TestBulkDeleteSnapshots_EmptyRequestRejected` + `TestBulkDeleteSnapshots_BothNamesAndPrefixRejected`).
- 200 + per-target results on partial failure (matching the existing `POST /api/v1/vms/bulk` convention), partial-failure case explicitly tested with `_PartialFailure`. Returning 207 would have been more semantically correct but uglier to consume — the per-target `success` flag is precise enough.
- Prefix matching is server-side, which is the right call: future RBAC (3.1.5) can enforce ownership server-side, and operators with hundreds of snapshots don't have to round-trip giant ID lists.
- `bulk=true` attribute on each emitted `snapshot.deleted` event lets receivers (Activity timeline, webhooks, future schedulers) collapse fan-outs at consumer side without losing per-snapshot fidelity.
- Empty-result happy path (`_PrefixNoMatchesEmptyResponse`) returns 200 with `results: []` — better than 404 because "the prefix matched nothing" is semantically valid input.
- CLI mirrors the same shape: `--name` / `--prefix` mutually exclusive, prints `OK` / `FAIL` per snapshot with a summary log line; non-zero exit when any deletion fails.

Test coverage: API (6 cases — names / prefix / partial failure / empty / both / no-match), CLI (5 cases), mock-server replays the bulk endpoint with all four error branches, two Playwright + one JSDOM scenario.

**Conflict alert:** PR #185 changes `vm.Manager.CreateSnapshot(ctx, vmID, name string)` → `(ctx, vmID, types.SnapshotSpec)`. Several test cases here call `mockMgr.CreateSnapshot(nil, "vm-bd", "auto-1")` (the old signature), so they will fail to compile if #185 lands first. Trivial rebase — wrap each call in `types.SnapshotSpec{Name: ...}`. Whichever lands first forces this update on the other.

Notes (non-blocking):
- Drive-by would have been to update `MockManager.CreateSnapshot` to honor `bulk` semantics for fan-out, but that's a different concern and unnecessary here.
- README progress block + new ROADMAP row will conflict with PRs #178, #181, #185, #192, #194 — first PR through forces a trivial rebase on the rest.

---

chinmay28 (commented):

Reviewed against the architecture and 2.3.5. CI green; locally `TestBulkDeleteSnapshots_*` (6/6) and `TestCLI_SnapshotDelete_*` (5/5) all pass.

Strengths:

- **Built on existing manager primitives.** No `vm.Manager` interface change — `BulkDeleteSnapshots` calls `ListSnapshots` (for prefix mode) plus per-target `DeleteSnapshot`. Both `LibvirtManager` and `MockManager` already implement these, so the bulk endpoint is fully testable through the mock without any libvirt side support. Good architectural restraint — exactly the kind of "build on what's there" that CLAUDE.md prefers.
- **Mutual-exclusion validation lives at the boundary**, returns typed `invalid_bulk_request` with explicit `http.StatusBadRequest` rather than relying on `statusForAPIError` to map it. Passes the empty-payload + both-fields cases via `TestBulkDeleteSnapshots_EmptyRequestRejected` and `_BothNamesAndPrefixRejected`.
- **Whitespace cleanup before validation.** `cleanedNames` trims and drops empty entries, so a payload like `{"names": ["", "  ", "auto-1"]}` is treated as one-snapshot, not three. Defensive but proportionate.
- **Per-target result list** rather than aborting the whole request on the first failure — the right call for cleanup workflows where one missing snapshot shouldn't undo the rest. `TestBulkDeleteSnapshots_PartialFailure` pins this behavior.
- **`bulk: "true"` event attribute** lets the Activity timeline / future webhook consumers collapse the fan-out. Future-friendly without forcing the decision now.
- **Server-side prefix matching** is the right call. Doing it client-side would force every consumer (CLI, GUI, future API clients) to re-implement the loop — and the rationale you gave about future RBAC/3.1.5 is correct: the server is the place to enforce ownership filters.
- **No `force=true` flag.** Letting libvirt's "snapshot has children" rejection bubble up as a per-target failure is safer than silent force-deletes. Right call.
- **CLI mutual-exclusion** is enforced at flag-parse time, so the CLI never even initialises a VM manager for bad input. `TestCLI_SnapshotDelete_BothFlags` and `_NoFlags` cover this.

Non-blocking observations:

1. **Prefix matches are case-sensitive.** That's consistent with libvirt snapshot naming (which is also case-sensitive), but worth a one-liner in `docs/openapi.yaml`'s `BulkDeleteSnapshotsRequest` description so operators don't expect case-insensitive matching like the `?tag=` filters elsewhere in the API.
2. The `Results` slice preserves order: explicit `names` follow input order; `prefix` mode follows `ListSnapshots` order. Not documented in the schema — operators might assume sorted output. Consider either documenting "results are returned in the order targets were processed" in the schema, or sorting by name before returning.
3. `prefix == "/"` (or any single non-empty char) would still match every snapshot. That's the intended behavior, but pairing it with confirmation in the GUI ("Delete 47 snapshots matching prefix `''`?") would be a nice future safety belt — outside the scope of this PR.

Roadmap row 2.3.5 correctly marked Done. README progress reconciles.

---

PR #195, opened 2026-05-06
docs(roadmap): resolve summary merge conflict

march26bot:
## Summary
- remove the leftover merge-conflict markers from the roadmap summary table
- reconcile the metrics/events priority blurbs so they reflect the current shipped state
- keep the remaining priority ordering unchanged

## Testing
- not run (docs-only change)

PR #200, opened 2026-05-07
test(events): cover bus slow-drop, persistence-error, and overflow paths

chinmay28:
## Summary

Fills three unit-test gaps in `internal/events/bus.go` called out by the **4.2.17** task list (`EventBus` fan-out / slow-drop / ID monotonicity, etc.). ID monotonicity, fan-out, cancel, drain, and duplicate-name collision are already covered. The remaining slow-drop / persistence-error / publish-overflow paths are now covered too.

These three branches in `bus.go` are operationally important — the bus is invoked from the libvirt event loop, where blocking would freeze every domain lifecycle callback in the daemon. A regression here is silent and hard to debug.

## What changed

**`internal/events/bus_test.go`** — three new tests, no production code touched:

- **`TestBusSlowSubscriberIsDropped`** — a subscriber whose channel is full must be dropped, not block the bus. Asserts a fast subscriber still receives every event and the store persists the full set even while a slow subscriber is being dropped. Triggers `subscriberBufSize + 32` publishes and confirms the bus continued processing.
- **`TestBusPersistenceErrorStillFansOut`** — when `store.AppendEvent` returns an error, the bus must still deliver the event to live subscribers (with a `transient-*` ID prefix) instead of silently dropping it. Pins the contract documented in `process()`.
- **`TestBusPublishOverflowDropsWithoutBlocking`** — `Publish` must never block, regardless of pressure. Uses a `blockingStore` whose `AppendEvent` waits on a gate so the bus's run goroutine stalls; that fills `publishCh`; subsequent `Publish` calls must return within 200 ms (drop, not deadlock). Catches a regression where a future change accidentally makes the publish channel a blocking write.

## Test plan

- [x] `CGO_ENABLED=1 go test -tags libvirt_dlopen -count=1 -short -run "TestBusSlowSubscriberIsDropped|TestBusPersistenceErrorStillFansOut|TestBusPublishOverflowDropsWithoutBlocking" -v ./internal/events/...` — 3/3 pass
- [x] `CGO_ENABLED=1 go test -tags libvirt_dlopen -race -count=1 -short ./internal/events/...` — clean
- [x] `CGO_ENABLED=1 go vet -tags libvirt_dlopen ./internal/events/...` — clean
- [x] `gofmt -l internal/events/bus_test.go` — clean

## Notes

- **No conflicts with any open chinmay28 PR.** Touches a single test file (`internal/events/bus_test.go`) that none of #178 / #179 / #181 / #185 / #192 / #194 / #198 modify.
- **4.2.17 is not fully ✅** — webhook integration retry/final-failure tests already exist in `internal/webhooks/manager_test.go` and SSE replay/heartbeat/410 already exist in `internal/api/handlers_events_stream_test.go`. The remaining gap on 4.2.17 is the **E2E** scenario ("start a real VM and assert `vm.started` arrives on the live SSE stream"), which depends on real libvirt + QEMU and belongs in `tests/e2e/`. This PR is a focused incremental piece, not the closing PR for the row.
- **Why not also touch `docs/ROADMAP.md`?** Three of the open PRs (#185, #192, #198) plus #195 already toggle roadmap rows / regen the README progress block; adding a fourth would force every other rebase to re-resolve the same conflict. Better to keep this test-only PR isolated and let whichever roadmap PR lands first claim the row text.

https://claude.ai/code/session_01XR77LPYprfmo9m6bZhC7sR

---
_Generated by [Claude Code](https://claude.ai/code/session_01XR77LPYprfmo9m6bZhC7sR)_

### DIFF
--- a/internal/events/bus_test.go
+++ b/internal/events/bus_test.go
@@ -229,6 +229,131 @@ func TestNewSystemEvent_NoAttrs(t *testing.T) {
 	}
 }
 
+// TestBusSlowSubscriberIsDropped guards the invariant that a subscriber whose
+// channel is full does not block the bus.  We fill the buffered subscriber
+// channel without ever reading from it, then publish more events than the
+// subscriber buffer can hold and confirm the bus continued processing — the
+// store reflects every persisted event even though the subscriber received
+// at most subscriberBufSize.
+func TestBusSlowSubscriberIsDropped(t *testing.T) {
+	store := &mockStore{}
+	bus := New(store)
+	bus.Start()
+	defer bus.Stop()
+
+	// Slow subscriber — never reads from ch.
+	_, cancel := bus.Subscribe("slow")
+	defer cancel()
+
+	// Fast subscriber — reads everything to confirm the bus still fans out
+	// to healthy subscribers while a slow one is being dropped.
+	fastCh, cancelFast := bus.Subscribe("fast")
+	defer cancelFast()
+
+	const total = subscriberBufSize + 32
+	for i := range total {
+		bus.Publish(&types.Event{Type: "tick", Message: fmt.Sprintf("%d", i)})
+	}
+
+	deadline := time.After(2 * time.Second)
+	for received := 0; received < total; {
+		select {
+		case <-fastCh:
+			received++
+		case <-deadline:
+			t.Fatalf("fast subscriber received %d/%d before timeout", received, total)
+		}
+	}
+
+	stored := store.all()
+	if len(stored) != total {
+		t.Fatalf("expected %d persisted events, got %d", total, len(stored))
+	}
+}
+
+// TestBusPersistenceErrorStillFansOut covers the branch in process() where
+// store.AppendEvent fails: the bus must still deliver to subscribers (with a
+// transient ID prefix) instead of silently dropping the event.
+func TestBusPersistenceErrorStillFansOut(t *testing.T) {
+	store := &mockStore{Err: fmt.Errorf("disk full")}
+	bus := New(store)
+	bus.Start()
+	defer bus.Stop()
+
+	ch, cancel := bus.Subscribe("test")
+	defer cancel()
+
+	bus.Publish(&types.Event{Type: "vm.started"})
+
+	select {
+	case evt := <-ch:
+		if evt.Type != "vm.started" {
+			t.Fatalf("type = %q, want vm.started", evt.Type)
+		}
+		if evt.ID == "" {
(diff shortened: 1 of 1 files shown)


chinmay28 (commented):

Reviewed against 4.2.17 unit-test list. CI green; locally `go test -race` of the three new tests passes cleanly, and the captured `WARN` log lines confirm each test exercises the exact branch it targets:

- `slow subscriber dropping event subscriber=slow#1` (slow-drop path)
- `failed to persist event type=vm.started error=disk full` (persistence-error path; subscriber still received with `transient-*` ID)
- 15× `publish channel full, dropping event type=flood` (publish-overflow non-blocking path)

Strengths:

- **Targets the right invariants.** The `Publish` non-blocking property is the single most important correctness guarantee of the bus — it's invoked from the libvirt event loop, where blocking would freeze every domain lifecycle callback in the daemon. The PR's framing ("a regression here is silent and hard to debug") is accurate; this is exactly the kind of test that earns its keep over time.
- **`TestBusPublishOverflowDropsWithoutBlocking`** uses a `blockingStore` whose `AppendEvent` waits on a gate, then asserts each subsequent `Publish` returns within 200 ms. Catches the regression where `publishCh <- evt` is accidentally changed to a blocking write.
- **`TestBusPersistenceErrorStillFansOut`** verifies the `transient-` ID prefix contract, so any future change that loses the prefix breaks loudly. That's the right behavior to pin — consumers may depend on it.
- **`TestBusSlowSubscriberIsDropped`** uses `total = subscriberBufSize + 32` (ties the test to the production constant), and asserts the *fast* subscriber still gets every event while the slow one is being dropped. Confirms the bus didn't just drop everyone.
- **Test-only PR with no roadmap edit.** Right call — keeps it conflict-free with the other roadmap-touching PRs (#185, #192, #194, #195, #198) so it can be merged independently.

Non-blocking nit:

- `blockingStore` declares `released chan struct{}`, `seq uint64`, and `events []*types.Event`, none of which are read by `TestBusPublishOverflowDropsWithoutBlocking` (only `gate` is used for the wait). Dead test scaffolding — could be trimmed to keep future readers from wondering what the fields do. The `mu` lock is also unused since the bus's run goroutine can't pile up calls (only one at a time waits on `gate`). Pure cleanup, no behavior change.

Approving in spirit. PR description's positioning of "this is *part of* 4.2.17, not the closing PR" is accurate — E2E (`vm.started` over real SSE with a real VM) is still TBD per `tests/e2e/`.

---

PR #193, opened 2026-05-06
feat(console): add in-memory ticket store

march26bot:
## Summary
- add an internal console ticket store with one-time VM-scoped tickets and a background janitor
- expose IssueTicket and ConsumeTicket APIs plus typed errors for missing, expired, and VM-mismatch cases
- add focused unit tests for single-use behavior, expiry, janitor cleanup, uniqueness, and concurrent consumption

## Testing
- not run locally in this workspace because the Go toolchain is not installed ()

PR #203, opened 2026-05-07
feat(console): add ticket-issuance endpoint (5.1.2)

chinmay28:
## Summary

Adds `POST /api/v1/vms/{id}/console/ticket` on top of the in-memory ticket store landed in #193. Returns a single-use opaque token plus the expiry timestamp and the websocket URL the client should dial once the console websocket itself lands (5.1.4). Roadmap row **5.1.2** marked ✅ Done.

## Behaviour

| Outcome | Status | Code |
|---|---|---|
| Ticket issued | 200 | `{ticket, expires_at, websocket_url}` |
| Unknown VM ID | 404 | `resource_not_found` |
| VM not running | 409 | `vm_not_running` |
| Console subsystem disabled | 503 | `service_unavailable` |

The handler reads the caller's bearer token via the existing `extractAPIKey()` helper and stores it on the ticket so the websocket handler in 5.1.4 can forward it. The daemon wires `console.NewStore()` once at startup and registers `Close()` in the shutdown path so the janitor goroutine drains cleanly.

### Backend
- `pkg/types/console.go` — new `ConsoleTicket` wire type.
- `internal/api/handlers_console.go` — new handler + `Server.SetConsoleStore` setter.
- `internal/api/router.go` — adds `consoleStore *console.Store` field; registers `POST /vms/{id}/console/ticket`.
- `internal/daemon/daemon.go` — instantiates `console.NewStore()`, wires it into the API server, and stops it during shutdown alongside the webhook manager.

### Spec / docs
- `docs/openapi.yaml` — new path + `ConsoleTicket` schema.
- `web/src/api/generated/schema.d.ts` — regenerated via `npm run generate:api`.
- `CLAUDE.md` REST quick reference — new row.
- `docs/ROADMAP.md` — 5.1.2 marked ✅ Done.

### Tests
- `TestIssueConsoleTicket_RunningVM_ReturnsTicket` — happy path; round-trips through the store and asserts the ticket is single-use.
- `TestIssueConsoleTicket_StoppedVM_Returns409` — confirms `vm_not_running` typed code.
- `TestIssueConsoleTicket_UnknownVM_Returns404` — `resource_not_found`.
- `TestIssueConsoleTicket_NoStoreReturns503` — uncovered fixtures get a clean `service_unavailable` instead of a panic.
- `TestIssueConsoleTicket_TicketCarriesCallerAPIKey` — pins the contract that the bearer token is captured on the ticket so 5.1.4 can forward it.

## Test plan

- [x] `CGO_ENABLED=1 go test -tags libvirt_dlopen -count=1 -short ./internal/api/... ./internal/console/...` — all green; the 5 new tests are 5/5 pass
- [x] `npm --prefix web run build` — frontend bundle builds (354 KB / 113 KB gzip)
- [x] `npm --prefix web run generate:api` — schema regenerates without diff drift

## Notes

- **No CLI/GUI in this PR.** The console-ticket endpoint is consumed by the (forthcoming) console page (5.1.7) — there is no separate operator workflow that issues a ticket from the CLI. Following the same pattern as the webhook *signing* path: it's invisible plumbing until the consumer ships.
- **Why 503 when the store is nil instead of registering nothing?** Keeping the route registered means GUI / probe traffic gets a typed error code (`service_unavailable`) rather than a 404 that's indistinguishable from a missing VM. The same pattern is used by the webhook test endpoint when no `WebhookTester` is wired (#179).
- **Ticket TTL** is the store's default (60s) and `JanitorInterval` is 30s. Both are tuneable via `console.NewStoreWithOptions(...)` if a follow-up wants to expose them as config knobs (5.1.6).
- **No event emission.** Issuing a ticket is a routine operator action — flooding the activity timeline with `console.ticket_issued` events would be noise. The `console.session_terminated` event from 5.1.5 is the right place to surface session lifecycle.

https://claude.ai/code/session_01Bx7j2BTHr4oPjE5XV7Z4HN

---
_Generated by [Claude Code](https://claude.ai/code/session_01Bx7j2BTHr4oPjE5XV7Z4HN)_


chinmay28 (commented):
## LGTM — implementation matches roadmap 5.1.2 and is well-tested

(Cannot formally approve as the same account that authored the PR — leaving as a review comment.)

### What I checked

- **Handler logic** (`internal/api/handlers_console.go`): walks the VM via `vmManager.Get`, returns 404 when missing (via `sanitizeManagerError`, which maps "not found" to `resource_not_found`), 409 `vm_not_running` when state isn't `Running`, 503 `service_unavailable` when no console store is wired. All four typed error codes match the OpenAPI doc and the PR body. ✅
- **Bearer-token capture**: `extractAPIKey(r)` is the same helper the auth middleware uses, so the ticket carries whatever credential the caller authenticated with. The `_TicketCarriesCallerAPIKey` test pins this contract, which is exactly the seam 5.1.4 will need. ✅
- **Daemon wiring** (`internal/daemon/daemon.go`): `console.NewStore()` is created at startup and `consoleStore.Close()` is called from `closeResources()` ahead of the event-bus shutdown — clean drain order. ✅
- **Route registration** (`internal/api/router.go`): the new endpoint sits inside `/api/v1/vms/{vmID}` and inherits the existing chain (`rateLimitMiddleware` → `apiKeyAuth`), so authn + rate-limit are both applied. The roadmap note phrased the order as "auth → rate-limit → handler"; the actual chain is `rate-limit → auth → handler`, which matches every other route in the API and is the safer ordering (unauthenticated callers can still be rate-limited). Worth noting but not a blocker.
- **Tests**: 5 fresh cases covering happy path + single-use + 409 + 404 + 503 + bearer-token round-trip. The `consoleTestServer` helper is a clean copy of the existing test fixture pattern.
- **Docs**: OpenAPI `paths` and `ConsoleTicket` schema added, generated TS schema regenerated, `CLAUDE.md` REST quick reference and `docs/ROADMAP.md` row 5.1.2 both updated to ✅ Done.
- **CI**: Backend + frontend checks both green.

### Minor observations (non-blocking)

- The websocket URL is built as `"/api/v1/vms/" + v.ID + "/console?ticket=" + token`. Since `v.ID` is the daemon-generated `vm-<unix-nano>` and `token` is base64-RawURL, neither needs URL escaping today — but if either ever becomes user-controlled this should switch to `url.PathEscape` / `url.QueryEscape`. Fine to leave as-is for 5.1.2.
- The 503 path when `consoleStore == nil` is good defensive plumbing, but the daemon today always wires the store, so this branch is only exercised in tests. The PR body acknowledges this consciously and it matches the `WebhookTester` pattern from #179, so 👍.

LGTM.

---

chinmay28 (commented):
Clean implementation of 5.1.2; matches the roadmap's threat model and integrates cleanly. (Cannot self-approve as repo owner — leaving as Comment with verified-LGTM.)

**Verified locally** (PR branch applied on top of main, CGO build with `-tags libvirt_dlopen`):
- `go vet` and `go build` clean.
- `go test ./internal/api/... -run TestIssueConsoleTicket` — all 5 cases pass: running VM happy path (round-trip + single-use), stopped VM → 409 `vm_not_running`, unknown VM → 404 `resource_not_found`, no-store → 503 `service_unavailable`, and the bearer-token-carries-through pin.
- `go test ./internal/console/...` clean.
- Daemon wires `console.NewStore()` and `Close()` is invoked from `closeResources()` in the same block as `webhookMgr.Stop()` — janitor goroutine drains correctly.
- OpenAPI + generated TS schema regenerated consistently.

**Likely-conflict heads-up — coordinate with #205**

Both this PR and #205 add a new `pkg/types/console.go`. The contents are complementary (this PR defines `ConsoleTicket`; #205 defines `ConsoleIntent` + `ConsoleEndpoint`), but git will mark the file as a conflicting "added by both" on whichever PR merges second. Whichever lands first, the other will need a small rebase to merge the two type sets into one file.

**Minor observations (non-blocking):**
- Handler maps every `vmManager.Get` failure to 404 via `sanitizeManagerError` — same pattern other handlers use, so a transient libvirt error here would mis-report as 404. Consistent with the codebase; mention in case 5.1.4 wants to refine.
- The `WebsocketURL` is intentionally a relative path (matches the test pin and the OpenAPI schema description). This differs from the roadmap example's `wss://...` form but is more flexible — the client resolves against its current scheme. Worth a one-line note in `docs/ARCHITECTURE.md` when 5.1.12 lands.
- `apiErrorStatusCode` doesn't have an explicit `vm_not_running` mapping; this PR's handler doesn't rely on it (it sets 409 directly), but a future caller of `s.consoleStore` that uses `writeAPIError(..., apiErrorStatusCode(...))` would fall through to the default. Worth filing for 5.1.4.

Pre-existing test `TestStreamEvents_ReplayOverflowReturnsGone` hangs in the broader `internal/api` suite both with and without this PR — not introduced here, but worth flagging separately as a flaky/blocked test on `main`.

---

PR #207, opened 2026-05-08
feat(vm): suspend/resume across API, CLI, and GUI (2.5.3)

chinmay28:
## Summary

Adds libvirt pause/unpause as a first-class VM lifecycle operation. Suspend freezes a running VM's CPU + memory state without releasing host resources; resume picks up exactly where suspend left off, so operators can pause workloads instantly without the cost of a stop/start cycle. Roadmap row **2.5.3** marked ✅ Done.

## Behaviour

| Outcome | Status | Code |
|---|---|---|
| VM running → suspended | 200 with `{"status":"suspended"}` | (state transitions to `paused`) |
| VM paused → resumed | 200 with `{"status":"resumed"}` | (state transitions to `running`) |
| Suspend a stopped VM | 409 | `vm_not_running` |
| Suspend an already-paused VM | 409 | `vm_already_paused` |
| Resume a non-paused VM | 409 | `vm_not_paused` |
| Unknown VM ID | 404 | `resource_not_found` |

### Backend
- **`pkg/types/vm.go`** — new `VMStatePaused = "paused"` state constant.
- **`internal/vm/manager.go`** — `Suspend(ctx, id)` / `Resume(ctx, id)` on the `Manager` interface.
- **`internal/vm/lifecycle.go`** — `LibvirtManager.Suspend` / `.Resume` call `dom.Suspend()` / `dom.Resume()` after pre-flight `GetState` checks; `domainStateToVMState` now maps `DOMAIN_PAUSED` to `VMStatePaused` instead of conflating it with `VMStateStopped`.
- **`internal/vm/events.go`** — libvirt lifecycle handler now maps `DOMAIN_EVENT_SUSPENDED` / `DOMAIN_EVENT_PMSUSPENDED` to `paused` (was `stopped`). Pure relabel; the existing `vm.suspended`/`vm.resumed` event types are untouched.
- **`internal/vm/mock_manager.go`** — `Suspend` / `Resume` with the same state guards plus `SuspendErr` / `ResumeErr` injection hooks.
- **`internal/vm/quota_manager.go`** — wrapper forwards both methods.
- **`internal/api/handlers_vm.go`** — `SuspendVM` / `ResumeVM` handlers; both publish `vm.suspend_requested` / `vm.resume_requested` app events.
- **`internal/api/router.go`** — `POST /vms/{id}/suspend` and `/resume`.
[...]

PR #205, opened 2026-05-08
feat(console): add GetConsoleEndpoint manager method (5.1.3)

chinmay28:
## Summary

Adds the foundation for the console websocket proxy in 5.1.4: a `vm.Manager.GetConsoleEndpoint(ctx, id, intent)` method the proxy will call to discover where to dial for a given VM/intent pair. Roadmap row **5.1.3** marked ✅ Done.

This is independent of #203 (5.1.2 ticket-issuance) — different files, different layer of the stack.

## Behaviour

| Outcome | Code | Where |
|---|---|---|
| VNC running | 200 with `{intent:"vnc", host, port}` | parsed from `<graphics type='vnc' port='...' listen='...'/>` |
| Serial running | 200 with `{intent:"serial", path}` | parsed from `<console type='pty' tty='/dev/pts/N'>` (falls back to `<serial><source path>`) |
| Stopped VM | typed `vm_not_running` (→ HTTP 409) | runs only against live domains |
| Port not yet assigned | typed `console_unavailable` | libvirt records `port='-1'` until autoport resolves |
| No graphics device | typed `console_unavailable` | |
| pty not yet allocated | typed `console_unavailable` | stopped or pre-start XML carries no `tty=` / `<source path>` |
| Unknown intent | typed `invalid_console_intent` | `vnc` / `serial` are the only valid values |

### Backend
- `pkg/types/console.go` — `ConsoleIntent` enum (`vnc`/`serial`) and structured `ConsoleEndpoint` (Host/Port for VNC, Path for serial). Struct return is more future-proof than the `(host, port, err)` signature in the roadmap because it cleanly carries the serial pty path.
- `internal/vm/manager.go` — `GetConsoleEndpoint(ctx, id, intent)` on the interface.
- `internal/vm/lifecycle.go` — `LibvirtManager.GetConsoleEndpoint` looks up the live domain, refuses with `vm_not_running` when stopped, and parses XML via a pure-Go `parseConsoleEndpointFromXML` helper that's unit-testable without libvirt.
- `internal/vm/mock_manager.go` — synthetic VNC/serial defaults for running VMs, plus `SeedConsoleEndpoint` (pin an endpoint) and `SeedConsoleListener` (bind a real loopback listener so 5.1.4's websocket-proxy tests can dial a known target — the "synthetic listener" the roadmap calls for).
- `internal/vm/quota_manager.go` — quota-wrapping manager forwards the new method.

### Spec / docs
[...]

PR #209, opened 2026-05-08
feat(vm): force-stop across API, CLI, and GUI (2.5.4)

chinmay28:
## Summary

Adds an explicit immediate-destroy lifecycle action for the case where the guest OS is wedged or the operator deliberately wants to skip ACPI shutdown. Today the only path that calls `dom.Destroy()` is the fallback inside `Stop` when `dom.Shutdown()` itself errors — there's no way to skip ACPI from the start. Roadmap row **2.5.4** marked ✅ Done.

## Behaviour

| Outcome | Status | Code |
|---|---|---|
| VM running → destroyed | 200 with `{"status":"force_stopped"}` | (state transitions to `stopped`) |
| Force-stop a stopped VM | 409 | `vm_already_stopped` |
| Unknown VM ID | 404 | `resource_not_found` |

### Backend
- **`internal/vm/manager.go`** — new `ForceStop(ctx, id)` on the `Manager` interface.
- **`internal/vm/lifecycle.go`** — `LibvirtManager.ForceStop` looks up the live domain, refuses with typed `vm_already_stopped` when state is `DOMAIN_SHUTOFF`, and calls `dom.Destroy()` directly. Persists `VMStateStopped` on success.
- **`internal/vm/mock_manager.go`** — `ForceStop` mirrors the state guard with a `ForceStopErr` injection hook for tests.
- **`internal/vm/quota_manager.go`** — wrapper forwards the new method.
- **`internal/api/handlers_vm.go`** — `ForceStopVM` handler; publishes `vm.force_stop_requested` for audit.
- **`internal/api/router.go`** — `POST /vms/{id}/force-stop`.
- **`internal/api/validation.go`** — `statusForAPIError` returns 409 for `vm_already_stopped`.

### CLI
- **`internal/cli/vm.go`** — `vmsmith vm force-stop <id>`.

### Frontend / GUI
- **`web/src/api/client.ts`** — `vms.forceStop(id)`.
- **`web/src/pages/VMDetail.jsx`** — confirmation-gated "Force Stop" button (running-only) using a Lightning icon to differentiate from the regular Stop button.

### Spec / docs
- **`docs/openapi.yaml`** — new path documented.
- **`CLAUDE.md`** — REST quick reference and CLI command summary updated.
- **`docs/ROADMAP.md`** — row 2.5.4 marked ✅ Done.

### Tests
- **MockManager (4 cases):** `TestMockManager_ForceStop`, `_AlreadyStopped` (409 `vm_already_stopped`), `_NotFound`, `_ErrorInjection`.
- **API integration (4 cases):** `TestForceStopVM`, `TestForceStopVM_AlreadyStopped_Returns409`, `TestForceStopVM_NotFound_Returns404`, `TestForceStopVM_Error`.
- **CLI (3 cases):** `TestCLI_VMForceStop`, `_AlreadyStopped`, `_NotFound`.
- **Mock server + Playwright:** `tests/web/mock-server.js` honours the 409 `vm_already_stopped` contract; new `tests/web/gui.spec.js` spec `force-stop running VM` round-trips through the confirm dialog and asserts the button disappears once the VM transitions to stopped.

## Test plan

- [x] `go build -tags libvirt_dlopen ./...` — clean
- [x] `go test -count=1 -short -tags libvirt_dlopen -skip TestStreamEvents_ReplayOverflowReturnsGone ./...` — green (the skipped test is a pre-existing flake on `main`, called out in #203's review and again in #207)

## Notes

- **Why a separate endpoint instead of `?force=true` on `/stop`?** A boolean modifier hides a destructive action behind a query parameter that's easy to miss in CLI history and proxy logs. A dedicated path makes it explicit in the audit trail (`POST /vms/{id}/force-stop` vs `POST /vms/{id}/stop`) and lets the GUI confirm-dialog and CLI command name carry the warning text. Same reasoning the existing `Restart` follows.
- **Why a `vm.force_stop_requested` event rather than `vm.stop_requested`?** Same precedent as `vm.suspend_requested` vs `vm.start_requested` in #207 — keeps the audit stream distinguishable from the graceful path. The libvirt-source `vm.stopped` will still fire from the lifecycle callback once the domain transitions, regardless of which path triggered the destroy.
- **Independent of #205, #207.** Touches `internal/vm/manager.go` (new method), `internal/vm/lifecycle.go`, `internal/vm/mock_manager.go`, `internal/vm/quota_manager.go`, `internal/api/handlers_vm.go` + `router.go` + `validation.go`, plus client/GUI surfaces. No overlap with the console PRs (`pkg/types/console.go`) or the suspend/resume PR (`internal/vm/events.go`, `pkg/types/vm.go`).
- **No DHCP / cloud-init impact.** ForceStop keeps the VM's existing libvirt domain definition, MAC, and IP reservation — the cloud-init ISO and instance-id are untouched.

---
_Generated by [Claude Code](https://claude.ai/code/session_0123UTNMMk5DUf3QTHbQS2Fg)_


chinmay28 (commented):
LGTM — clean implementation of roadmap row 2.5.4. Approved-equivalent (author can't self-approve).

**Verified locally**
- `CGO_ENABLED=1 go test -tags libvirt_dlopen -short ./...` green across all packages.
- 4 mock + 4 API + 3 CLI test cases all pass; Playwright `force-stop running VM` spec exercises the GUI confirm-dialog path.

**What I liked**
- A dedicated `POST /vms/{id}/force-stop` endpoint rather than `?force=true` on `/stop` is the right call. Audit trails and proxy logs make the destructive action explicit, and the `vm.force_stop_requested` app event is distinguishable from the graceful path's `vm.stop_requested` for downstream webhooks.
- `LibvirtManager.ForceStop` correctly pre-flights the domain state and returns typed `vm_already_stopped` instead of letting `dom.Destroy()` fail with a low-level libvirt error. The `statusForAPIError` extension to map that code to 409 lines up with the `vm_locked` / `vm_not_running` precedent.
- The GUI `window.confirm` gate before firing `forceStopMut.execute` is the right UX guard for an operation that bypasses ACPI shutdown — and the dialog text spells out the data-loss risk. Good choice of the `Zap` icon to differentiate visually from the regular `Square` Stop button.
- Mock manager includes `ForceStopErr` injection hook matching the existing pattern for other lifecycle methods.
- The PR notes call out that ForceStop preserves the libvirt domain definition, MAC, and DHCP reservation — that's the right behavior; we don't want a force-stop to surprise-rotate the IP on next start.

**Things to be aware of (non-blocking)**
- This PR adds row 2.5.4 to the same `Phase 2 / 2.5` roadmap table as #207's row 2.5.3. Whichever PR merges second needs a one-line conflict resolution in `docs/ROADMAP.md`.
- `internal/api/validation.go`'s `statusForAPIError` is also touched by #207 — different codes, same stanza. Mechanical conflict only.
- Consider whether ForceStop should also be gated by the new `Locked` flag from 2.5.2. Right now `Locked=true` blocks `Delete` but not lifecycle actions, and the PR description correctly says "Lock is a delete guard, not a freeze." That's consistent — just calling out the design choice for the record.

---

chinmay28 (commented):
## Review — LGTM (cannot self-approve)

Implements roadmap row **2.5.4** (force-stop / immediate destroy). Verified locally on `origin/claude/eager-cerf-force-stop`:

- `CGO_ENABLED=1 go test -tags libvirt_dlopen -count=1 -run "TestForceStop|TestMockManager_ForceStop|TestCLI_VMForceStop" ./...` — green (4 mock + 4 API + 3 CLI cases pass)
- `CGO_ENABLED=1 go test -tags libvirt_dlopen -count=1 -short -skip TestStreamEvents_ReplayOverflowReturnsGone ./...` — all green except for the `TestBusSlowSubscriberIsDropped` flake that exists on `main` independent of this PR

### Architecture alignment

- A dedicated endpoint (`POST /vms/{id}/force-stop`) instead of a `?force=true` modifier on `/stop` is the right call — the audit trail clearly distinguishes graceful vs forced, and the GUI confirm dialog has somewhere to live. Same precedent as `/restart` vs `?restart=true`.
- The libvirt path checks `DOMAIN_SHUTOFF` only before destroying — paused VMs (#207's new state) flow through to `dom.Destroy()`, which is correct: an operator force-stopping a paused VM almost certainly wants the QEMU process killed, not "no-op because it's not running."
- `vm.force_stop_requested` event mirrors the `vm.suspend_requested` / `vm.start_requested` precedent (request vs libvirt-confirmed truth).

### Merge-order interaction with #207

Both PRs touch `internal/api/validation.go::statusForAPIError`:

```go
// This PR (209)
case "vm_locked", "vm_already_stopped":
    return 409
```

```go
// #207
case "vm_locked", "vm_not_running", "vm_not_paused", "vm_already_paused":
    return 409
```

Trivial textual conflict only — whoever merges second should fold the other's codes into the same `case`. Both sets of codes need to map to 409.

### Minor nits (non-blocking)

1. The `quotaManager.ForceStop` declaration is split across multiple lines, breaking the alignment of the `Start`/`Stop`/`Restart` one-liners just above it. Cosmetic, but `gofmt` will likely realign once the file is touched again.
2. CLI prints `force-stopped` (with hyphen) but the API status field is `force_stopped` (underscore). Not wrong, but it's a tiny stylistic split — most other CLI surfaces print snake_case. Trivial.

LGTM, ready to merge after #207 lands (or first, depending on which conflicts the other prefers to resolve).

---

PR #208, opened 2026-05-08
api: reject unsupported metrics stream replay params

march26bot:
$## Summary\n- reject `Last-Event-ID` and `?since=` on `/api/v1/vms/{id}/stats/stream` with a clear 400 error\n- add integration coverage for both unsupported replay inputs\n- document the no-replay contract in the OpenAPI description\n\n## Testing\n- not run locally (`go` toolchain is unavailable in this environment)

PR #211, opened 2026-05-09
fix(events): SSE replay-overflow returns 410 before SSE response is committed (4.2.17)

chinmay28:
## Summary

Fixes a real bug in `StreamEvents` and the partial flake the rest of the codebase has been working around (`-skip TestStreamEvents_ReplayOverflowReturnsGone` shows up in #205 and #207's test plans). Roadmap row **4.2.17** moves forward — the SSE replay/live-delivery slice of the integration test matrix is now shipped.

## The bug

`internal/api/handlers_events.go::StreamEvents` called `newSSEWriter` (which writes `200 OK` + `Content-Type: text/event-stream` and flushes the response status) **before** the replay-overflow check ran. The subsequent `w.WriteHeader(http.StatusGone)` was a silent no-op — clients saw 200 OK with the JSON error body buffered behind an SSE content-type and hung reading an SSE body that never terminated.

The companion test was also miswired: it appended 1001 events with `?since=1`, leaving only 1000 events to replay — exactly equal to `sseReplayLimit`, so the `len > sseReplayLimit` branch never triggered. The test consistently failed against the real handler (600s timeout), which is why every recent PR ran with `-skip TestStreamEvents_ReplayOverflowReturnsGone`.

### Handler (`internal/api/handlers_events.go`)
- `afterSeq` parsing and `Store.ListEventsAfterSeq` now run **before** `newSSEWriter`, so the 410 short-circuit can still commit `application/json` status.
- Replay events are deferred to a `replayed []*types.Event` and forwarded after `newSSEWriter` succeeds.

### Test fix (`handlers_events_stream_test.go`)
- Bump `TestStreamEvents_ReplayOverflowReturnsGone` to 1002 events so `?since=1` leaves 1001 events — strictly greater than `sseReplayLimit` (1000) — and the overflow path actually triggers.

### New 4.2.17 coverage
- **`TestStreamEvents_ReplayOverflow_…ContentTypeIsJSON`** (extends the existing test): asserts the 410 response carries `application/json` (not `text/event-stream`) so the silent-no-op regression can't reappear.
- **`TestStreamEvents_LiveDeliveryAfterReplay`**: exercises the previously uncovered SSE-handler → `bus.Subscribe` path. Opens an SSE stream, publishes via the bus on a 50ms tick, asserts the client receives the `vm.deleted` frame. Race-robust by design — there is no synchronous "Subscribe returned" signal, so the test publishes on a tick rather than once.
- **`wireEventBusWithStore` helper** (`handlers_events_emission_test.go`): routes the bus through the real bbolt store fixture so `AppendEvent` assigns monotonic IDs the SSE handler can replay. The existing `wireEventBus` uses a one-event `memStore` that cannot exercise replay.

### Roadmap (`docs/ROADMAP.md`)
- Row **4.2.17** annotated to record the SSE replay/live-delivery slice as shipped; webhook + retention unit and real-VM E2E coverage still pending in this row.

## Test plan

- [x] `CGO_ENABLED=1 go test -count=1 -tags libvirt_dlopen ./internal/api/...` — green (`6.4s`)
- [x] `CGO_ENABLED=1 go test -count=1 -tags libvirt_dlopen ./internal/events/... ./internal/store/...` — green
- [x] `CGO_ENABLED=1 go test -race -count=5 -tags libvirt_dlopen -run TestStreamEvents ./internal/api/...` — green (`10.99s`), no race-detector hits
- [x] **No `-skip TestStreamEvents_ReplayOverflowReturnsGone` needed** — the test passes consistently

## Notes

- **Why a separate "ContentType is JSON" assertion?** The original test only checked `resp.StatusCode == 410` and parsed the body, but Go's `http` will happily decode a JSON body off a 200 OK response too. Asserting `application/json` content-type pins down the actual fix surface — without that, an alternative regression that returned 410 with SSE content-type would still pass.
- **Why publish on a tick in the live-delivery test?** The bus's `Subscribe` happens after the SSE handler's replay loop. There is no synchronous signal that subscribe completed. Publishing once before subscribe leaves the event missed (bus fan-out targets only currently-registered subscribers). Publishing on a 50ms tick until the reader picks up a frame eliminates the race without sleeps or arbitrary delays. The deadline cap is 2s.
- **Independent of #205 / #207 / #209.** Touches `internal/api/handlers_events.go`, `handlers_events_stream_test.go`, `handlers_events_emission_test.go`, `docs/ROADMAP.md`. No overlap with the VM-manager work in the other open PRs.

https://claude.ai/code/session_01HDUnVJG4tosXDRmfpEmeHH

---
_Generated by [Claude Code](https://claude.ai/code/session_01HDUnVJG4tosXDRmfpEmeHH)_


chinmay28 (commented):
## Review — LGTM, this is a real bug fix (cannot self-approve)

Fixes a genuine bug — `newSSEWriter` was committing the 200 + `text/event-stream` headers before the replay-overflow check ran, so the subsequent `w.WriteHeader(http.StatusGone)` was a silent no-op. Anyone hitting that path got 200 OK with the JSON error body buffered behind an SSE content-type and an unterminated stream.

The reorder (parse `afterSeq` → query store → overflow check → newSSEWriter → write replay frames) is the right fix and exactly mirrors the order in #208's metrics-stream replay-rejection.

### Verified locally on `origin/claude/eager-cerf-WUA1b`

- `CGO_ENABLED=1 go test -tags libvirt_dlopen -count=1 -run TestStreamEvents ./internal/api/...` — green
- `CGO_ENABLED=1 go test -tags libvirt_dlopen -race -count=5 -run TestStreamEvents ./internal/api/...` — green (12s, no race-detector hits)
- `CGO_ENABLED=1 go test -tags libvirt_dlopen -count=1 -short ./internal/api/... ./internal/store/...` — green
  - `TestBusSlowSubscriberIsDropped` in `internal/events` is a pre-existing flake on `main` (verified independently), unrelated to this PR.

### What I particularly like

1. **The companion test fix.** The old test appended 1001 events with `?since=1`, leaving exactly 1000 to replay = `sseReplayLimit`, so `len > sseReplayLimit` was never true. Bumping to 1002 events fixes the off-by-one and the test now actually exercises the overflow branch. Explains every recent PR's `-skip TestStreamEvents_ReplayOverflowReturnsGone`.
2. **The content-type assertion** (`!strings.HasPrefix(ct, "application/json")`) on the 410 path. Without that, a regression that returned 410 with `text/event-stream` content-type would still pass the original `StatusCode == 410` check while still hanging clients. Tight regression test.
3. **`TestStreamEvents_LiveDeliveryAfterReplay`** with the tick-publish loop is exactly the right pattern for a race-sensitive subscribe-then-publish flow. No `time.Sleep`, no synchronization gymnastics, just "publish until the reader gets a frame or the deadline elapses." 2s deadline is generous.
4. **`wireEventBusWithStore` helper** correctly recognizes that the existing `wireEventBus` uses a one-event `memStore` that can't replay, so it adds a real-store variant. Future SSE replay/persist tests can build on this.

### Architecture alignment

- The 410 "client too far behind" semantics align with the roadmap's 4.2.17 task ("SSE replay + heartbeat + 410 + shutdown frame").
- The roadmap update annotates 4.2.17 as partially shipped (SSE replay/live-delivery slice done; webhook + retention units + real-VM E2E still pending) — that's accurate.

### Tiny nit (non-blocking)

`io.Copy(io.Discard, resp.Body)` after the test's `cancel()` is good hygiene to avoid a leak warning, but happens after the success-return path is essentially decided. If the inner reader goroutine still has unread bytes, this Copy may block briefly behind the ctx-cancelled http2 frames flushing. Not seen in the 5-iteration race run, so leaving it.

Ready to merge once approvals land. Strong recommend prioritizing this — every other open PR has been carrying `-skip TestStreamEvents_ReplayOverflowReturnsGone` because of this bug.

---

PR #212, opened 2026-05-09
feat(snapshots): editable description across API, CLI, and GUI (2.2.11)

chinmay28:
## Summary

Adds in-place editing of an existing snapshot's free-text description. Today snapshot descriptions can only be set at create time (2.2.8); a typo or after-the-fact "what was this for again?" required deleting and re-taking the snapshot. Roadmap row **2.2.11** marked ✅ Done.

Backed by libvirt's `SnapshotCreateXML(VIR_DOMAIN_SNAPSHOT_CREATE_REDEFINE)` primitive — the snapshot's disk/memory state, parent pointer, creation timestamp, and root-element attributes are all preserved verbatim. Only the `<description>` element is rewritten.

## Behaviour

| Outcome | Status | Code |
|---|---|---|
| Description updated | 200 with the snapshot JSON | (description re-emitted) |
| Description cleared (`""`) | 200 | (description omitted in response) |
| Description omitted (no-op) | 200 | (snapshot returned unchanged) |
| Description over 1024 chars | 400 | `invalid_description` |
| Snapshot not found | 404 | `resource_not_found` |
| Malformed JSON | 400 | `invalid_request_body` |

### Backend
- **`pkg/types/snapshot.go`** — new `SnapshotUpdateSpec{Description *string}`. Pointer semantics: `nil` = leave untouched; `&""` = clear.
- **`internal/vm/manager.go`** — `UpdateSnapshot(ctx, vmID, name, SnapshotUpdateSpec) (*Snapshot, error)` on the `Manager` interface.
- **`internal/vm/lifecycle.go`** — `LibvirtManager.UpdateSnapshot` looks up the existing snapshot, dumps its XML, threads a pure-Go `rewriteSnapshotDescription` helper, then calls `dom.CreateSnapshotXML(newXML, libvirt.DOMAIN_SNAPSHOT_CREATE_REDEFINE)`. Skips the libvirt round-trip entirely if the new value equals the existing description.
- **`rewriteSnapshotDescription`** — pure-Go helper unmarshals the snapshot's outer envelope (root-element attributes preserved), strips any existing `<description>` element(s), re-injects a properly-escaped one after `</name>`, and reassembles the doc. Unit-testable without libvirt.
- **`internal/vm/mock_manager.go`** — `UpdateSnapshot` mirrors the contract with a `UpdateSnapshotErr` injection hook and trims whitespace, matching the libvirt path.
- **`internal/vm/quota_manager.go`** — wrapper forwards the new method.
- **`internal/api/handlers_snapshot.go`** — `UpdateSnapshot` handler. Publishes `snapshot.updated` for audit.
- **`internal/api/router.go`** — `PATCH /vms/{id}/snapshots/{snapName}`.
- **`internal/api/validation.go`** — `validateUpdateSnapshotRequest` enforces the 1024-char cap.

### CLI
- **`internal/cli/snapshot.go`** — `vmsmith snapshot edit <vm-id> <snap-name> --description "..."`. Omitting `--description` is a no-op (useful as a connectivity probe). Passing `--description ""` clears.

### Frontend / GUI
- **`web/src/api/client.ts`** — `snapshots.update(vmId, snapName, { description })`.
- **`web/src/pages/VMDetail.jsx`** — Pencil-icon Edit button next to Restore on each snapshot row. Opens an `EditSnapshotModal` pre-populated with the current description; live character counter; round-trips through the new client.

### Spec / docs
- **`docs/openapi.yaml`** — new path documented; `UpdateSnapshotRequest` schema.
- **`CLAUDE.md`** — REST quick reference + CLI command summary updated.
- **`docs/ROADMAP.md`** — new row 2.2.11 marked ✅ Done.

### Tests
- **API integration (7 cases):** `TestUpdateSnapshot_SetsDescription`, `_ClearsDescription`, `_OmittedDescriptionIsNoOp`, `_RejectsLongDescription` (400 `invalid_description`), `_NotFound` (404 `resource_not_found`), `_BadJSON` (400 `invalid_request_body`), `_ManagerError`.
- **MockManager (6 cases):** `TestMockManager_UpdateSnapshot_SetsDescription`, `_TrimsWhitespace`, `_NilDescriptionIsNoOp`, `_VMNotFound`, `_SnapshotNotFound`, `_ErrorInjection`.
- **CLI (5 cases):** `TestCLI_SnapshotEdit_SetsDescription`, `_ClearsDescription`, `_NoFlagIsNoOp`, `_DescriptionTooLong`, `_NotFound`.
- **Pure-Go XML rewrite (6 cases):** `TestRewriteSnapshotDescription_AddsDescription`, `_ReplacesExistingDescription`, `_ClearsDescription`, `_PreservesAttributes` (xmlns + custom attrs survive), `_EscapesUnsafeText` (re-parses cleanly through `parseSnapshotXML`), `_BadXML`.
- **Mock server + Playwright:** `tests/web/mock-server.js` honours the same PATCH contract; new `tests/web/gui.spec.js` specs `edit snapshot description` and `clear snapshot description via edit` round-trip through the GUI.

## Test plan

- [x] `CGO_ENABLED=1 go test -tags libvirt_dlopen -count=1 -short -skip TestStreamEvents_ReplayOverflowReturnsGone ./...` — all green (the skipped test is a pre-existing flake on `main`, called out in #203/#207/#209/#211 — the fix lives in #211)
- [x] `CGO_ENABLED=1 go test -tags libvirt_dlopen -race -count=1 -run "TestUpdateSnapshot|TestRewriteSnapshotDescription|TestMockManager_UpdateSnapshot|TestCLI_SnapshotEdit" ./...` — green, no race-detector hits
- [ ] GitHub Actions to confirm (Playwright Chromium download is firewalled in this dev sandbox; CI runs the GUI specs)

## Notes

- **Why a struct return rather than `error`-only?** The handler needs to echo the canonical post-edit description back to clients (which may have trimmed whitespace, etc.) without the GUI making a second round-trip to refresh. Mirrors the existing `Update(VM)` shape.
- **Why pointer-typed `Description`?** Snapshot description is editable in two distinct ways — set or clear. A non-pointer string can't distinguish "leave untouched" from "set to empty string". Same convention as `VMUpdateSpec.AutoStart` / `Locked` (`*bool`) introduced in 2.5.1 / 2.5.2.
- **Why `SnapshotCreateXML(REDEFINE)` instead of an explicit edit primitive?** libvirt has no in-place metadata edit for snapshots — the official idiom is to redefine the snapshot from a modified XML doc. The REDEFINE flag is exactly designed for this case: it preserves disk/memory state, parent pointer, and creation timestamp, replacing only the metadata fields the operator changed.
- **Why pure-Go `rewriteSnapshotDescription`?** A naive `Marshal(snapshotXMLDoc{...})` would lose every field libvirt emits that we don't model (state, parent, memory, disks). The helper is byte-level, preserves the inner XML verbatim, and is unit-testable without a libvirt connection.
- **Why no rename support?** Renaming a snapshot would require copy + delete of the underlying disk overlay (the snapshot name is part of the disk path). That's a meaningful chunk of work and intentionally out of scope for 2.2.11 — operators who need a different name today can take a fresh snapshot and delete the old one.
- **Independent of #205 / #207 / #209 / #211.** Touches `pkg/types/snapshot.go`, `internal/vm/{manager,lifecycle,mock_manager,quota_manager}.go`, `internal/api/{handlers_snapshot,router,validation,api_test}.go`, `internal/cli/{snapshot,commands_test}.go`, plus the GUI/mock-server surface. No overlap with the console (`pkg/types/console.go`), VM-state lifecycle (`pkg/types/vm.go`, `internal/vm/events.go`), force-stop (`internal/api/handlers_vm.go::ForceStopVM`), or SSE replay (`internal/api/handlers_events.go`) PRs.

https://claude.ai/code/session_01SwX3L3yBcRd5Pwg6GxtRAL

---
_Generated by [Claude Code](https://claude.ai/code/session_01SwX3L3yBcRd5Pwg6GxtRAL)_


chinmay28 (commented):
## Review — LGTM (cannot self-approve)

Implements roadmap row **2.2.11** (snapshot description editing) cleanly across the stack. Verified locally on `origin/claude/eager-cerf-Em4Wh`:

- `CGO_ENABLED=1 go test -tags libvirt_dlopen -count=1 -run "TestUpdateSnapshot|TestRewriteSnapshotDescription|TestMockManager_UpdateSnapshot|TestCLI_SnapshotEdit" ./...` — green (24 cases pass)
- `CGO_ENABLED=1 go test -tags libvirt_dlopen -count=1 -short -skip TestStreamEvents_ReplayOverflowReturnsGone ./...` — green except for the pre-existing `TestBusSlowSubscriberIsDropped` flake on `main`

### Architecture alignment

1. **`SnapshotCreateXML(REDEFINE)` is the right primitive.** libvirt has no in-place edit for snapshot metadata — REDEFINE is the documented idiom and preserves disk/memory state, parent pointer, and creation timestamp. Skipping the libvirt round-trip when the description is unchanged is a nice optimization that avoids needless XML writes.
2. **Pointer-typed `Description *string`** is consistent with the existing `*bool` convention used by `VMUpdateSpec.AutoStart` / `Locked` (2.5.1 / 2.5.2). Distinguishing "leave alone" from "clear" is exactly the right reason to use a pointer.
3. **Pure-Go `rewriteSnapshotDescription`** sidesteps the trap of a naive `xml.Marshal(snapshotXMLDoc{...})` round-trip, which would lose every libvirt field the codebase doesn't model (state, parent, memory, disks). The byte-level approach preserves the inner XML verbatim and is unit-testable without a libvirt connection — exactly the pattern other recent PRs (#205's `parseConsoleEndpointFromXML`) follow.
4. **The 6 XML-rewrite cases** — including `_PreservesAttributes` (xmlns + custom attrs survive) and `_EscapesUnsafeText` (re-parses cleanly) — give me confidence the helper is robust against the real domain XML libvirt emits.
5. **`snapshot.updated` app event** with the `description` attribute included is the right audit-trail level. Operators can grep the event stream for "what was renamed when."

### Subtle nits (non-blocking)

1. **The publish-event `description` attribute is the new value, not a diff.** That's reasonable for audit, but worth noting that consumers can't reconstruct "what was the old description" from the event alone — they'd need to consult prior snapshot state. Probably fine; if a diff later becomes useful, the extension is `old_description` / `new_description`.
2. **Empty-string vs nil semantics in handler**: when `req.Description == nil`, the handler doesn't add a `description` attribute to the event (good — distinguishes no-op from clear). When `*req.Description == ""` it sets `description=""` in the event payload. That round-trip is consistent.
3. **Out-of-scope rename.** PR description correctly notes that rename is intentionally not supported (it would require copy + delete of the underlying disk overlay). That's the right call for this scope.

### Test coverage

24 new cases is thorough — API integration (7), MockManager (6), CLI (5), XML rewrite (6), plus mock-server + Playwright. The `_EscapesUnsafeText` re-parse-through-`parseSnapshotXML` assertion is a particularly nice symmetry check.

LGTM. No conflicts with the other open PRs (touches `internal/vm/{lifecycle,manager,mock_manager,quota_manager}.go`, `internal/api/{handlers_snapshot,router,validation}.go`, plus client/GUI surfaces — no overlap with the console PRs, the suspend/resume/force-stop trio, or the SSE replay fix).

---

PR #213, opened 2026-05-09
feat(vms): sortable list across API, CLI, and GUI (5.4.2)

chinmay28:
## Summary

Roadmap row **5.4.2** originally called for `?status=` plus `sort=created_at&order=desc` filtering on `GET /vms`. The status filter shipped earlier; this PR rounds out the row by adding the `sort` / `order` half across the whole stack. Roadmap row **5.4.2** marked ✅ Done.

## Behaviour

| Outcome | Status | Code |
|---|---|---|
| Sort + order applied | 200 (sorted page) | — |
| Unknown sort field | 400 | `invalid_sort` |
| Unknown order value | 400 | `invalid_order` |

Sort whitelist: `id` (default) | `name` | `created_at` | `state`.
Order whitelist: `asc` (default) | `desc`. All comparators tiebreak on `id` so pagination is deterministic across `LibvirtManager` (bbolt-key order) and `MockManager` (Go-map order).

### Types (`pkg/types/vm_sort.go`)
- Public `VMSort*` / `SortOrder*` constants — part of the API contract so OpenAPI consumers and the CLI/GUI can share one source of truth.
- `SortVMs([]*VM, sortField, order)` is a pure-Go in-place sort with a case-insensitive name comparator and an `id` tiebreak on every code path. Unknown fields silently fall back to id-asc; surface validation errors at the parsing layer.

### API (`internal/api/`)
- `vm_sort.go` — `parseVMSort` validates the request and returns a typed `*APIError` (400 `invalid_sort` / `invalid_order`) so unknown values cannot silently drop into a default ordering.
- `handlers_vm.go::ListVMs` parses sort first (so a 400 short-circuits before the manager call), then applies the existing tag/status filter, then `types.SortVMs`. The hand-rolled id sort is gone.
- `validation.go::statusForAPIError` adds `invalid_sort` / `invalid_order` to the 400 group.

### CLI (`internal/cli/vm.go`)
- `vmsmith vm list --sort <id|name|created_at|state> --order <asc|desc>`. Validation is symmetric with the API; bad values fail before the manager call with a clear error pointing at the whitelist.

### OpenAPI / TS schema
- `docs/openapi.yaml` — new `VMSort` and `SortOrder` reusable parameters on `/vms`, plus an explicit 400 response. `web/src/api/generated/schema.d.ts` regenerated via `npm run generate:api`.
- `web/src/api/client.ts` — `vms.list({ sort, order, ... })` adds the optional sort surface, threading the new params into the typed client.

### GUI (`web/src/pages/VMList.jsx`)
- Sort field + order dropdowns above the bulk action bar. Selection is reflected in URL search params (`?sort=`, `?order=`) so a deep-linked view keeps its ordering across reload, and changing either resets to page 1.

### Tests
- **API integration (7 cases):** sort by name asc/desc, sort by `created_at` desc, sort by `state` (with the running-pair tiebreak on id), page-spanning determinism (page 1 + page 2 read continuously), `invalid_sort`, `invalid_order`.
- **`pkg/types.SortVMs` unit (5 cases):** id-asc, name case-insensitive, created_at desc with id tiebreak, state ordering, unknown-field fallback, repeat-sort stability for paginated equal-key data.
- **CLI (4 cases):** sort by name, created_at desc, `--sort` whitelist rejection, `--order` whitelist rejection.
- **Mock-server:** the `/vms` GET stub now honours `sort` / `order` with the same 400 contract.
- **Playwright:** sort dropdowns reorder the live list and the URL captures the change so deep links survive reload.

### Docs
- `CLAUDE.md` — `GET /vms` quick-reference now lists sort/order params and the CLI flag set.
- `docs/ROADMAP.md` — row 5.4.2 marked ✅ Done with the sort/order slice called out, including the deterministic-pagination rationale.

## Test plan

- [x] `CGO_ENABLED=1 go test -tags libvirt_dlopen -count=1 -short -skip 'TestStreamEvents_ReplayOverflowReturnsGone|TestBusSlowSubscriberIsDropped' ./...` — all green (the two skipped tests are pre-existing flakes on `main`, called out in #205/#207/#209/#211 — `TestStreamEvents_ReplayOverflow…` is fixed in #211)

## Notes

- **Why a tuple return for the sort helper rather than two `error`s?** Returning `(field, order, err)` keeps the call site one line and prevents the handler from forgetting to validate either half. The 400 response carries the typed code so clients can branch on `invalid_sort` vs `invalid_order` without parsing the human message.
- **Why is the comparator in `pkg/types` rather than `internal/api`?** CLI `vm list` runs against the VM manager directly (no HTTP), so it needs the same sort logic. Putting it in `pkg/types` (alongside the `VM` type itself) lets both call sites share one implementation without creating an `internal/api` ← `internal/cli` import edge.
- **Tiebreak on `id` everywhere.** Without it, equal-key elements shuffle between requests in the `MockManager` Go-map iteration — a known flake source in the existing pagination tests.
- **No conflict with the open suspend/resume / force-stop / snapshot-edit PRs.** Those touch different `handlers_vm.go` functions (Suspend, Resume, ForceStop) plus snapshot/event surfaces. `ListVMs` is untouched by any of them.

https://claude.ai/code/session_01D3irdfcLE9MxA8HSXCBoua

---
_Generated by [Claude Code](https://claude.ai/code/session_01D3irdfcLE9MxA8HSXCBoua)_


chinmay28 (commented):
Reviewed end-to-end — LGTM (cannot self-approve; leaving as a comment).

**What I checked**
- `pkg/types/vm_sort.go::SortVMs` — case-insensitive name compare, `id` tiebreak on every branch, unknown field falls back to id-asc. Putting it in `pkg/types` (not `internal/api`) is the right call because both the API and the CLI surface call it directly. ✅
- `internal/api/vm_sort.go::parseVMSort` — whitelist enforced, returns typed `*APIError` (`invalid_sort` / `invalid_order`) before the manager call so a bad query never hits the data layer. `validation.go::statusForAPIError` updated to map both new codes to 400. ✅
- `handlers_vm.go::ListVMs` now parses sort first (short-circuits 400), then filters, then sorts. The hand-rolled `sort.Slice(... vms[i].ID < vms[j].ID)` is gone. ✅
- CLI symmetry: `vmsmith vm list --sort --order`, with the same whitelist enforced in CLI before hitting the manager. Error message points at the allowed values. ✅
- TS client + `VMList.jsx` — sort/order dropdowns, URL search params keep the deep-link state, page resets to 1 when sort/order changes. ✅
- OpenAPI: `VMSort` and `SortOrder` reusable parameters, explicit 400 response on `/vms`. Schema regenerates clean. ✅
- Mock server honours the same 400 contract.
- Test coverage: 5 sort-helper unit cases (id/name/created_at/state, unknown fallback, stable pagination), 7 API integration cases (sort by name asc/desc, created_at desc, state with tiebreak, page-spanning determinism, invalid_sort, invalid_order), 4 CLI cases, 1 Playwright spec. ✅

**Tests run locally**
- `CGO_ENABLED=1 go test -tags libvirt_dlopen -count=1 -short ./...` — all green (excluding the pre-existing flakes)
- `gofmt -l` over touched files — clean

**Architectural alignment**
Closes the remaining half of roadmap row 5.4.2. The status filter shipped earlier; this rounds it out with sort/order. The deterministic-pagination property (every comparator tiebreaks on `id`) is the right correctness invariant — without it, the `MockManager` Go-map iteration would shuffle equal-key elements between requests.

**Minor note (not blocking)**
`SortVMs` inverts the entire compare result in desc mode, which means an equal-key tiebreak also inverts (id-desc instead of id-asc within a desc group). `TestSortVMs_ByCreatedAt_DescTiebreaksOnID` explicitly asserts this. Pagination determinism within a single call is preserved either way; if you ever want "equal-key elements always appear in id-asc order regardless of primary direction" — e.g. for stable cursor pagination — that would need a non-inverting tiebreak. Worth flagging in a future cursor-pagination PR; no change needed here.

---

chinmay28 (commented):
## Review: LGTM (would approve, but cannot self-approve)

Verified locally on the PR branch:
- `CGO_ENABLED=1 go test -tags libvirt_dlopen -count=1 -short ./...` — all green (excluding the pre-existing flakes called out in the description). `pkg/types` now reports `ok` because of the new `vm_sort_test.go`.

### Architecture / roadmap alignment

- Roadmap row **5.4.2** matches the original task wording (`?status=running&sort=created_at&order=desc`); the status-filter half shipped earlier and this PR rounds out the row exactly as the row description promised.
- Putting `SortVMs` in `pkg/types` rather than `internal/api` is the right call — the CLI's `vm list` runs against the manager directly (no HTTP), so the API and CLI now share a single comparator. Avoids an `internal/api` ← `internal/cli` import edge.
- Validation is centralised in `internal/api/vm_sort.go::parseVMSort` (returns typed `*APIError`) and mirrored in `internal/cli/vm.go` — both surfaces use the same whitelist (`id`, `name`, `created_at`, `state` × `asc`, `desc`) and emit the same `invalid_sort` / `invalid_order` codes.
- `statusForAPIError` extension to map both new codes to 400 is the minimal correct change; no new error path needed in handlers.

### Implementation notes

- All comparators tiebreak on `id` — important for `MockManager` (Go-map iteration order) and necessary for the page-spanning determinism test. The `pkg/types` unit test specifically exercises this with equal-name and equal-timestamp inputs.
- Descending-order tiebreak: because `desc` simply inverts the entire compare result, equal-key pairs reverse on tiebreak too. The unit test (`TestSortVMs_ByCreatedAt_DescTiebreaksOnID`) explicitly asserts that behaviour, which is good — it documents the contract rather than leaving it implicit.
- **Minor behavior change for CLI users:** `vmsmith vm list` defaults change from `sort.SliceStable(... by CreatedAt asc)` to `sort=id asc`. Because VM IDs are `vm-<unix-nano>`, this is functionally equivalent for the common case (recent VMs sort the same way), so the change is safe in practice.
- Mock-server validation mirrors the API contract (`localeCompare` with id tiebreak, 400 on unknown values), so the Playwright spec exercises the same surface as live API tests.
- VMList URL round-trip: `sort=id` and `order=asc` are deliberately stripped from the URL when they equal the defaults, keeping shareable links clean. Reading from `searchParams` on first render means deep links (`?sort=name&order=desc`) keep their ordering across reload, matching the test assertion.

### Test coverage

- API integration: 7 cases (sort by name / desc / created_at desc / state / page-spanning / invalid_sort / invalid_order)
- `pkg/types.SortVMs` unit: 5 cases (id asc / case-insensitive name / created_at desc with tiebreak / state / unknown-field fallback / stable-pagination)
- CLI: 4 cases (sort by name / created_at desc / invalid sort / invalid order)
- Playwright: sort dropdowns reorder + URL captures change

No conflict with #215 (`RebootVM` is a separate handler).

---

chinmay28 (commented):
LGTM (can't formally approve as the author). Solid execution of the missing half of roadmap row 5.4.2.

Verified locally (with a stub `internal/web/dist/`):
- `CGO_ENABLED=1 go test -tags libvirt_dlopen -short -run "TestSortVMs|TestListVMs_Sort|TestListVMs_Reject|TestCLI_VMList_Sort|TestCLI_VMList_Reject" ./...` — all green
- `pkg/types`, `internal/api`, `internal/cli` build cleanly

Architecture notes:
- Putting `SortVMs` + the constants in `pkg/types/` is the right call — the CLI runs against the manager directly, not through HTTP, so they need to share the comparator without creating an `internal/cli` ↔ `internal/api` import edge.
- `parseVMSort` returning a typed `*APIError` keeps the validation symmetric with how the rest of the API surfaces 400s, and `statusForAPIError` was correctly extended to map `invalid_sort` / `invalid_order`.
- The `if desc { return !less }` form means the `id` tiebreak also flips with the primary direction. The test `TestSortVMs_ByCreatedAt_DescTiebreaksOnID` documents this; pagination remains deterministic across invocations, which is what 5.4.2's "deterministic pagination across backends" rationale requires.
- `sort` import in `internal/cli/vm.go` is correctly retained — there's still a `sort.Slice` at line 1248.

Minor (non-blocking) thoughts:
- The whitelist-validation logic is duplicated across `internal/api/vm_sort.go` and `internal/cli/vm.go`. Pulling a single `types.NormalizeVMSort(field, order) (field, order, error)` into `pkg/types` would let both call sites share the validation. Worth a small follow-up if you touch this again, but not blocking.
- Some users might expect a tiebreaker to always sort ascending regardless of primary direction (so deep-linked rows stay in the same secondary order when toggling `asc`/`desc`). The current "flip everything" is defensible and well-tested — just calling it out as a UX choice.

---

PR #215, opened 2026-05-09
feat(vm): in-guest reboot via libvirt dom.Reboot() (2.5.5)

chinmay28:
## Summary

Adds graceful in-guest reboot as a first-class VM lifecycle action. The existing `Restart` flow is a stop+start cycle that power-cycles the QEMU process — slower, more disruptive, and renegotiates the libvirt domain. `Reboot` keeps the domain alive and asks the guest OS to reboot itself via libvirt's `dom.Reboot()` (ACPI signal), preserving the IP / MAC / DHCP reservation and avoiding a fresh cloud-init pass. Roadmap row **2.5.5** marked ✅ Done.

This is a natural extension of the 2.5.X "Daemon-managed VM Lifecycle" section that recently shipped suspend/resume (2.5.3) and force-stop (2.5.4).

## Behaviour

| Outcome | Status | Code |
|---|---|---|
| VM running → reboot signal sent | 200 with `{"status":"rebooted"}` | (state stays `running`) |
| VM stopped/paused | 409 | `vm_not_running` |
| Unknown VM ID | 404 | `resource_not_found` |

### Backend
- **`internal/vm/manager.go`** — new `Reboot(ctx, id)` on the `Manager` interface.
- **`internal/vm/lifecycle.go`** — `LibvirtManager.Reboot` looks up the live domain, refuses with typed `vm_not_running` when state is not `DOMAIN_RUNNING`, then calls `dom.Reboot(0)`. Default flag (no `VIR_DOMAIN_REBOOT_GUEST_AGENT`) so it doesn't depend on qemu-guest-agent being installed in the guest.
- **`internal/vm/mock_manager.go`** — `Reboot` mirrors the state guard with a `RebootErr` injection hook for tests.
- **`internal/vm/quota_manager.go`** — wrapper forwards the new method.
- **`internal/api/handlers_vm.go`** — `RebootVM` handler; publishes `vm.reboot_requested` for audit.
- **`internal/api/router.go`** — `POST /vms/{vmID}/reboot`.
- **`internal/api/validation.go`** — `vm_not_running` already maps to 409 from 2.5.3, no change needed.

### CLI
- **`internal/cli/vm.go`** — `vmsmith vm reboot <id>`.

### Frontend / GUI
- **`web/src/api/client.ts`** — `vms.reboot(id)`.
- **`web/src/pages/VMDetail.jsx`** — `RefreshCw`-icon Reboot button visible only while running; distinct from the existing Restart button (which is stop+start). The tooltip spells out the difference.
[...]

PR #216, opened 2026-05-10
feat(images): bulk-delete by ids or tag across API, CLI, and GUI (2.3.6)

chinmay28:
## Summary

Operators managing many built/uploaded images currently have to delete them one at a time — every release-candidate sweep means N round-trips with no partial-failure visibility. This adds a single-request bulk-delete endpoint that mirrors the snapshot `bulk_delete` (2.3.5) shape: pass either an explicit `ids` list or a `tag` selector (case-insensitive), and get a per-target result back so a missing image surfaces as `resource_not_found` without aborting the rest. Roadmap row **2.3.6** marked ✅ Done.

This sits naturally alongside the recently shipped snapshot bulk-delete (#198 / 2.3.5) and rounds out section 2.3 "Bulk Operations" for resource cleanup.

## Behaviour

| Outcome | Status | Code |
|---|---|---|
| `ids` or `tag` selector applied | 200 with per-target results | — |
| Both `ids` and `tag` set | 400 | `invalid_bulk_request` |
| Neither set | 400 | `invalid_bulk_request` |
| Tag matches no images | 200, `results: []` | — |
| Some IDs missing | 200, mixed results | `resource_not_found` per missing item |

Tag matching uses the existing case-insensitive `storage.FilterImagesByTag` helper, so the same rules as `GET /images?tag=` apply.

### Backend
- **`internal/api/handlers_image.go`** — `BulkDeleteImages` validates exactly-one-of `ids`/`tag`, resolves the tag selector, then loops `DeleteImage` per target. Each success emits an `image.deleted` app event with `bulk=true` for audit (mirrors `snapshot.deleted` w/ `bulk=true` from 2.3.5).
- **`internal/api/router.go`** — `POST /images/bulk_delete` registered before the `/images/{imageID}` routes so chi's literal-path-wins rule applies.
- **`docs/openapi.yaml`** — new path documented; `BulkDeleteImagesRequest`, `BulkDeleteImagesResponse`, `BulkDeleteImageResult` schemas added.

### CLI
- **`internal/cli/image.go`** — `vmsmith image delete` now accepts an optional positional `<image-id>` and `--tag <tag>` (mutually exclusive). The tag path enumerates matches, deletes one by one, and prints per-image OK/FAIL — same shape as `vmsmith snapshot delete --prefix`.

### Frontend
- **`web/src/api/client.ts`** — `images.bulkDelete({ids?, tag?})`.
- **`web/src/pages/ImageList.jsx`** — per-row + select-all checkboxes, a "Delete selected" button in a sticky action bar, and a dismissible "X of Y succeeded" result banner. Selection state is pruned on page/filter churn so it never references rows that scrolled away.

### Tests
- **API integration (6 cases):** `TestBulkDeleteImages_ByIDs`, `_ByTag` (case-insensitive), `_PartialFailure` (`resource_not_found` surfacing), `_EmptyRequestRejected`, `_BothIDsAndTagRejected`, `_TagNoMatchEmptyResponse`.
- **CLI (4 cases):** `TestCLI_ImageDelete_NoArgsErrors`, `_BothIDAndTagErrors`, `_TagDeletesAllMatching`, `_TagNoMatchPrintsMessage`.
- **Mock server:** `/images/bulk_delete` route honours the same 400 contract.
- **Playwright + JSDOM mock GUI:** `bulk delete selected images` and `bulk delete via select-all images` round-trip through the new UI.

### Docs
- **`CLAUDE.md`** — REST quick reference and CLI command summary updated.
- **`docs/ROADMAP.md`** — new row **2.3.6** added under section 2.3 Bulk Operations, marked ✅ Done.

## Test plan

- [x] `CGO_ENABLED=1 go test -tags libvirt_dlopen -count=1 -short -skip 'TestStreamEvents_ReplayOverflowReturnsGone|TestBusSlowSubscriberIsDropped' ./...` — all green (the two skipped tests are pre-existing flakes on `main`, called out in #205/#207/#209/#211/#212/#213/#215)

## Notes

- **Why a `tag` selector instead of a `prefix` like snapshots?** Image IDs are opaque (`img-<unix-nano>`), but image *tags* are the operator's organisational tool — they already drive `?tag=` filtering in the GET path and the GUI tag-filter chips. Operators sweep release cohorts ("delete every image tagged `rc-2026-05`"), not name-prefixed sets. Tag selection re-uses the existing `FilterImagesByTag` helper so the matching rules are consistent across `GET /images?tag=` and `POST /images/bulk_delete`.
- **`bulk=true` event attribute** mirrors snapshot 2.3.5 so audit consumers can distinguish single-item deletes from batch sweeps without scanning timestamps.
- **Path order in the router** — `r.Post("/bulk_delete", ...)` is registered before the `{imageID}` routes; chi's trie matches literal paths first regardless of registration order, but I kept the explicit ordering for clarity.
- **No conflict with #215 (VM reboot — `RebootVM` handler in `handlers_vm.go`) or #213 (VM list sort — `ListVMs` handler).** This PR touches `handlers_image.go` and `ImageList.jsx`, both untouched by the open lifecycle work.

https://claude.ai/code/session_01GHzUKRt1iZNKumAis2oAbB

---

---
_Generated by [Claude Code](https://claude.ai/code/session_01GHzUKRt1iZNKumAis2oAbB)_


chinmay28 (commented):
LGTM (can't formally approve as the author). Faithful mirror of the 2.3.5 snapshot bulk_delete pattern, applied to images.

Verified locally:
- `CGO_ENABLED=1 go test -tags libvirt_dlopen -short -run "TestBulkDeleteImages|TestCLI_ImageDelete_NoArgsErrors|TestCLI_ImageDelete_BothIDAndTagErrors|TestCLI_ImageDelete_TagDeletesAllMatching|TestCLI_ImageDelete_TagNoMatchPrintsMessage" ./internal/api/... ./internal/cli/...` — all 10 tests pass

Architecture / contract notes:
- Route registration order in `internal/api/router.go` puts `/bulk_delete` before `/{imageID}` (chi's literal-path-wins rule would handle either ordering, but the explicit ordering is more readable).
- `image.deleted` event with empty `vmID` matches the existing single-delete handler at `handlers_image.go:149`. Adding `bulk=true` as an attribute matches snapshot 2.3.5's `bulk=true` audit shape — good consistency for log/webhook consumers.
- Tag selector reuses `storage.FilterImagesByTag` (case-insensitive), so the matching rules are identical to `GET /images?tag=`. That's the right call.
- `tag` selector justification (vs `prefix` like snapshots) is well-argued in the PR description — image IDs are opaque, tags are the operator's organizational tool. Agree.

Minor (non-blocking) thoughts:
- No de-duplication on `ids` input — `{"ids": ["img-1", "img-1"]}` will attempt to delete twice; the second attempt returns `resource_not_found`. Snapshot bulk_delete probably behaves the same way; not worth diverging.
- Potential TOCTOU race in the tag path: another caller could delete an image between `ListImages` → `DeleteImage`, which surfaces as a per-target `resource_not_found` rather than aborting. That's the right behavior for a bulk endpoint.
- The GUI's `useEffect` that prunes `selected` when `imageList` changes calls `setSelected(new Set())` unconditionally when `imageList` is empty even if `selected` was already empty — minor, just an unnecessary re-render. Easy fix if you touch this again.
- Request middleware logs the body of bulk-delete requests (per the existing logger middleware that captures up to 4096 bytes for POST/PUT). For tag selectors that's fine, but be aware that ID lists end up in structured log output.

---

PR #220, opened 2026-05-11
feat(vms): free-text search filter across API, CLI, and GUI (2.2.13)

chinmay28:
## Summary

`GET /api/v1/vms` already supports `?tag=` (2.2.4) and `?status=` (5.4.2), but operators with more than a handful of VMs still have to scroll the list to find a specific machine. This adds a `?search=<text>` query param that does a case-insensitive substring match across **name, description, and tags** so the GUI search box, CLI `--search`, and any scripted client all use the same predicate. Roadmap row **2.2.13** marked ✅ Done.

ID is intentionally **excluded** from the match — opaque `vm-<unix-nano>` strings rarely match a useful operator query and would produce noisy false positives when the substring happens to look numeric.

This rounds out section **2.2 VM Tags & Metadata** so the filter surface and the metadata fields stay symmetric.

## Behaviour

| Outcome | Status | Notes |
|---|---|---|
| `?search=web` (substring in name) | 200 | matches `web-prod-01`, `web-staging` |
| `?search=customer` (substring in description) | 200 | matches VMs with that word in `description` |
| `?search=team-` (substring in tag) | 200 | matches VMs with any tag containing `team-` |
| `?search=WEB` (uppercase needle) | 200 | case-insensitive — matches `Web-Prod-01` |
| `?search=%20%20alpha%20%20` | 200 | whitespace-trimmed — matches `alpha` |
| `?search=web&status=running&tag=prod` | 200 | filters compose additively (intersection) |

Filters compose additively — `?search=web&tag=prod&status=running` narrows to running web-* VMs tagged `prod`. The match is case-insensitive: the haystack is lowercased, and the API/CLI handlers lowercase + trim the needle before calling the helper.

### Backend (API + shared predicate)
- **`pkg/types/vm_search.go`** — new `VMMatchesSearch(vm, query)` predicate. Lowercase substring scan over name + description + tags. Empty query matches every VM; nil VM never matches; caller is responsible for lowercasing the needle (locked in by a unit test so future callers don't accidentally regress the case-insensitive contract).
- **`internal/api/handlers_vm.go`** — extends the existing tag/status filter loop to also evaluate `?search=` after trimming + lowercasing the value. Lives next to the tag/status checks so the three filters share one pass over the VM list.
- **`docs/openapi.yaml`** — adds the `SearchFilter` parameter under `/vms` with a description noting the substring contract, the case-insensitive guarantee, and the deliberate ID exclusion.

### CLI
- **`internal/cli/vm.go`** — `vmsmith vm list --search <text>` flag, applied with the same trim + lowercase + `VMMatchesSearch` pipeline as the API. Combines with the existing `--tag` and `--status` flags.

### Frontend
- **`web/src/pages/VMList.jsx`** — adds a debounced (250 ms) search input with a `Search` icon and an `X` clear-button next to the existing tag chips. The committed query is round-tripped through the URL via `?search=`, so a filtered view is shareable. The fetch dependency includes `searchFilter` so the list refreshes immediately when the debounce settles.
- **`web/src/api/client.ts`** — `vms.list({...})` now accepts `search`. Regenerated `web/src/api/generated/schema.d.ts` via `npm --prefix web run generate:api`.
- **`tests/web/mock-server.js`** — mirrors the same predicate so Playwright assertions can actually reduce the visible VM list when the user types.

### Tests
- **Unit (`pkg/types/vm_search_test.go`, 6 cases):** empty query / nil VM / name substring / description substring / tag substring / case-handling contract / empty description.
- **API integration (`internal/api/api_test.go`, 7 new cases):** `TestListVMs_FilterBySearch_MatchesName`, `_MatchesDescription`, `_MatchesTag`, `_IsCaseInsensitive`, `_NoMatch`, `_CombinesWithStatus`, `_CombinesWithTag`, `_TrimsWhitespace`.
- **CLI (`internal/cli/commands_test.go`, 5 new cases):** `TestCLI_VMList_FilterBySearch_MatchesName`, `_MatchesDescription`, `_MatchesTag`, `_NoMatch`, `_CombinesWithStatus`.
- **Playwright (`tests/web/gui.spec.js`, 2 new cases):** `search input filters the VM list and round-trips through the URL` (asserts the input filters down + URL gets `search=web` + clearing restores the list and removes the param), `search input matches no VMs when query has no hits`.

### Docs
- **`CLAUDE.md`** — REST API quick reference for `GET /vms` extended to mention `?search=`, the case-insensitive contract, the whitespace-trim, the ID exclusion, and the additive composition with `tag` / `status`. CLI line mirrors via `vmsmith vm list --search`.
- **`docs/ROADMAP.md`** — new row **2.2.13** under section 2.2, marked ✅ Done.

## Test plan

- [x] `CGO_ENABLED=1 go test -tags libvirt_dlopen -count=1 -short -skip 'TestStreamEvents_ReplayOverflowReturnsGone|TestBusSlowSubscriberIsDropped' ./...` — all green (the two skipped tests are pre-existing flakes flagged in #205/#207/#209/#211/#212/#213/#215/#216/#217/#218/#219)
- [x] `npm --prefix web run generate:api` — schema regenerates with the new query param

## Notes

- **Why not also match against VM ID?** IDs are opaque `vm-<unix-nano>` strings — operators rarely type them from memory, and the all-numeric tail produces noisy false positives when the search term looks numeric (e.g. searching for `2026` matches every VM created this year). If a use-case shows up, an `?include_id=true` opt-in is a small follow-up.
- **Why a shared `pkg/types` helper?** The CLI's `vm list` filters in-process against the manager rather than hitting the REST API (it shares the same code path as the daemon), so without a shared predicate the API and CLI could drift on edge cases like trim/lowercase semantics. Keeping the matcher in `pkg/types` next to `SortVMs` (5.4.2) follows the same pattern that 5.4.2 used for the sort comparator.
- **Why debounced in the GUI?** A keystroke-per-fetch pattern fans out one request per character; 250 ms is the sweet spot — long enough to coalesce a fast typist, short enough to feel live. The committed query is what drives `useFetch`; the input box keeps its own state so backspacing doesn't cause a jarring partial reset.
- **No conflict with the recent slices.** All of #205/#207/#209/#211/#212/#213/#215/#216/#217/#218/#219 are merged; this PR is built on the post-merge `main`. The bulk-action bar from #219 and the tag/status filters from 2.2.4/5.4.2 are untouched; the new search box sits above the tag chips.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Hg67uYdZke9neDj69ACnhe)_


chinmay28 (commented):
LGTM — tight scope, predicate placement and API/CLI/GUI symmetry are all consistent with the existing tag/status filter shape from 2.2.4 / 5.4.2. (Can't formally approve since I'm also the author; leaving as a review-comment.)

- `CGO_ENABLED=1 go test -tags libvirt_dlopen -count=1 -short -skip 'TestStreamEvents_ReplayOverflowReturnsGone|TestBusSlowSubscriberIsDropped' ./pkg/types/... ./internal/api/... ./internal/cli/... ./internal/vm/...` — all pass, including the 6 new `pkg/types.VMMatchesSearch` cases, 8 new API integration cases, and 5 new CLI cases.

- `pkg/types/vm_search.go` keeps the matcher next to `SortVMs` so the API/CLI handlers share a single contract. The `TestVMMatchesSearch_CallerLowercasesQuery` test locks in the "caller lowercases the needle" invariant — good guard against a future caller regressing case-insensitivity.
- API/CLI both lowercase + trim the needle before delegating to the shared helper, then run the predicate inside the existing tag/status filter loop. One pass over the VM list; filters compose additively as the docs claim.
- Excluding ID from the haystack is the right call — `vm-<unix-nano>` strings produce false positives on numeric substrings and operators rarely type them.
- Frontend: debounced (250 ms) `<input>` + URL round-trip via `?search=`; `useFetch` re-fetches when the committed query settles. Same pattern the tag/status chips established.
- Roadmap row 2.2.13 added under section 2.2 alongside the metadata fields.

No conflicts with PR #221/#222/#223 — disjoint file sets.

---

chinmay28 (commented):
## Review

Clean vertical slice for roadmap row **2.2.13**. Follows the pattern established by 5.4.2 (sort) and 2.2.4 (tag) — a shared `pkg/types` predicate, API + CLI both routing through it, mock-server mirroring it for Playwright, and the GUI debouncing the input. The roadmap row is detailed and the implementation matches the contract exactly.

### Verified locally

- `CGO_ENABLED=1 go test -tags libvirt_dlopen -count=1 -short -skip 'TestStreamEvents_ReplayOverflowReturnsGone|TestBusSlowSubscriberIsDropped' ./...` — all packages green, including the 7 new API integration cases, 5 new CLI cases, and 6 new unit cases for `VMMatchesSearch`.

### Things I liked

- `TestVMMatchesSearch_CallerLowercasesQuery` is exactly the right test — locks in the caller-lowercase-needle contract so a future caller that passes a raw `?search=` value silently regressing the case-insensitive promise will fail loudly.
- ID intentionally excluded with a documented rationale (`vm-<unix-nano>` numeric noise) — matches `CLAUDE.md`'s description of the ID scheme. Sensible.
- Filter composition is a single pass over the VM list — `tag` + `status` + `search` evaluated in sequence rather than three list passes. Nice.
- GUI's URL round-trip drops the `?search=` param when cleared (rather than setting it to empty), keeping the URL clean. The dependency array on `useFetch` correctly uses `searchFilter` (the committed value), not `searchInput` (the typed-but-not-yet-debounced value).
- Page reset (`setPage(1)`) is wired to `searchFilter` changes — important because changing the filter while sitting on page 3 would otherwise drop the user into a possibly-empty page.

### Nits (non-blocking)

- `VMMatchesSearch` lowercases the haystack on every call (3+ allocations per VM per request). For the current scale this is fine; if the VM count ever climbs into the tens of thousands, caching `nameLower` / `descLower` / `tagsLower` on the VM struct would help.
- The `vm.Description != ""` short-circuit is a minor micro-optimisation — `strings.Contains("", q)` is already false for non-empty `q`, so it's defensible either way.

---

---

PR #221, opened 2026-05-11
feat(images): sortable list across API, CLI, and GUI (5.4.5)

chinmay28:
## Summary

5.4.2 added a whitelisted sort to `GET /api/v1/vms` (id / name / created_at / state) plus `vm list --sort --order` and VMList dropdowns. The image list was left behind — `GET /api/v1/images` accepts only `?tag=` + pagination, and the CLI silently hardcoded created_at-asc.

This rounds out section **5.4 Pagination & Filtering** so images share the same sort surface VMs already do, and operators with hundreds of images can flip to "biggest first" or alphabetical the same way. Roadmap row **5.4.5** marked ✅ Done.

## Behaviour

| Request | Status | Notes |
|---|---|---|
| `GET /images?sort=name&order=asc` | 200 | case-insensitive name compare; id tiebreak |
| `GET /images?sort=size&order=desc` | 200 | biggest first (size_bytes); id tiebreak |
| `GET /images?sort=created_at&order=desc` | 200 | newest first; id tiebreak |
| `GET /images?sort=name&page=2&per_page=2` | 200 | filter → sort → paginate, deterministic across requests |
| `GET /images?tag=prod&sort=name&order=desc` | 200 | `?tag=` applies first, then sort, then pagination |
| `GET /images?sort=format` | 400 | `invalid_sort` |
| `GET /images?order=sideways` | 400 | `invalid_order` |

`?tag=` filter runs before sort so `X-Total-Count` reflects the filtered population. All comparators tiebreak on `id` so two independent fetches of the same page return the same set — same contract `SortVMs` locks in.

### Backend (API + shared sort)
- **`pkg/types/image_sort.go`** — new `SortImages(imgs, field, order)` predicate. Whitelisted constants `ImageSortID` / `ImageSortName` / `ImageSortSize` / `ImageSortCreatedAt`. Mirrors `SortVMs` shape and id-tiebreak contract.
- **`internal/api/image_sort.go`** — `parseImageSort` mirroring `parseVMSort`: typed 400 `invalid_sort` listing the supported fields and `invalid_order` for anything other than asc/desc.
- **`internal/api/handlers_image.go`** — `ListImages` now parses sort, applies it after `?tag=` filter, before pagination. Same control flow shape as `ListVMs`.

### CLI
- **`internal/cli/image.go`** — `vmsmith image list --sort <id|name|size|created_at> --order <asc|desc>` flags. The old hardcoded created_at-asc default is replaced with id-asc so CLI and API agree on a single ordering contract (matches what 5.4.2 did for VMs).

### Frontend
- **`web/src/pages/ImageList.jsx`** — sort dropdowns next to the existing tag chips. State is round-tripped through the URL (`?sort=&order=`) so a sorted view is shareable; non-default values are written into the URL, defaults are dropped. `useFetch` dependency includes sort/order so the list re-fetches when either changes.
- **`web/src/api/client.ts`** — `images.list({...})` now accepts `sort` + `order`. Regenerated `web/src/api/generated/schema.d.ts` via `npm --prefix web run generate:api`.
- **`tests/web/mock-server.js`** — mirrors the same validation (400 on bad sort/order) plus the same comparator so Playwright assertions actually reorder.

### Tests
- **Unit (`pkg/types/image_sort_test.go`, 6 cases):** id-asc default / name case-insensitive / size desc with id tiebreak / created_at desc with id tiebreak / unknown field falls back to id / stable pagination on equal-name input.
- **API integration (`internal/api/api_test.go`, 7 new cases):** `TestListImages_SortByName`, `_SortBySizeDesc`, `_SortByCreatedAtDesc`, `_SortCombinesWithTagFilter`, `_SortPaginationDeterministic` (page 1 + page 2 hand-off), `_RejectsInvalidSort`, `_RejectsInvalidOrder`. Adds a `seedStoredImageFull` helper that controls size + created_at without needing qemu-img.
- **CLI (`internal/cli/commands_test.go`, 4 new cases):** `TestCLI_ImageList_SortByName`, `_SortBySizeDesc`, `_RejectsInvalidSort`, `_RejectsInvalidOrder`.
- **Playwright (`tests/web/gui.spec.js`, 1 new case):** `sort controls reorder the image list and round-trip through the URL` — default id-asc, switch to name-asc, switch to size-desc, URL captures `sort=size&order=desc`.

### Docs
- **`CLAUDE.md`** — REST API quick reference for `GET /images` extended to document `?sort=` + `?order=`, the id-tiebreak determinism guarantee, and the CLI mirror.
- **`docs/openapi.yaml`** — new `ImageSort` parameter (reuses the existing `SortOrder` parameter), referenced from `/images`.
- **`docs/ROADMAP.md`** — new row **5.4.5** under section 5.4, marked ✅ Done.

## Test plan

- [x] `CGO_ENABLED=1 go test -tags libvirt_dlopen -count=1 -short -skip 'TestStreamEvents_ReplayOverflowReturnsGone|TestBusSlowSubscriberIsDropped' ./...` — all green (same two pre-existing flakes that #205/.../#220 skip)
- [x] `npm --prefix web run generate:api` — schema regenerates with the new query params

## Notes

- **Default sort changed from created_at-asc → id-asc in CLI.** The previous CLI default was created_at-asc (no API counterpart). 5.4.2 normalised VMs on id-asc — both the API default and the CLI default — and this PR does the same for images so the two layers agree. Users who relied on the implicit created_at-asc ordering can ask for it explicitly with `--sort created_at`.
- **Why size as a sort field?** Asked for repeatedly by operators looking to free host disk — "what are the biggest images I haven't touched recently". `ImageSortName` covers alphabetical and `ImageSortCreatedAt` covers chronological; `ImageSortSize` is the third axis specific to images. `ImageSortState` doesn't have an obvious image counterpart so the enum is shorter than the VM one.
- **No conflict with PR #220 (free-text search for VMs).** This PR touches images only; it does not modify `pkg/types/vm_*.go`, `web/src/pages/VMList.jsx`, or any of the VM handlers. The two PRs are independently mergeable.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Eh4KBd3N37bM4RnYRpmPNo)_


chinmay28 (commented):
LGTM — direct mirror of the 5.4.2 VM-sort pattern, plus the image-specific `size` axis. (Can't formally approve since I'm also the author; leaving as a review-comment.)

- `CGO_ENABLED=1 go test -tags libvirt_dlopen -count=1 -short -skip 'TestStreamEvents_ReplayOverflowReturnsGone|TestBusSlowSubscriberIsDropped' ./pkg/types/... ./internal/api/... ./internal/cli/... ./internal/vm/...` — all pass, including the 6 new `pkg/types.SortImages` cases, 7 new API integration cases, and 4 new CLI cases.

- `pkg/types/image_sort.go` mirrors `vm_sort.go` (whitelisted constants + `sort.SliceStable` + id tiebreak). `ImageSortSize` reads `SizeBytes` directly — that's the persisted field on `types.Image`.
- `internal/api/image_sort.go` matches `parseVMSort` shape — typed 400 `invalid_sort` / `invalid_order`, defaults to id-asc, lowercase + trim.
- `ListImages` runs `?tag=` filter first, then sort, then pagination — correct ordering so `X-Total-Count` reflects the filtered population.
- CLI default flipped from implicit `created_at`-asc to id-asc — agreed call: the 5.4.2 normalisation made id-asc the cross-layer default for VMs and this keeps images symmetric. Users wanting the prior behaviour ask explicitly via `--sort created_at`.
- Frontend ImageList round-trips sort/order through URL query params (defaults dropped) — matches the VMList behaviour 5.4.2 established.
- Roadmap row 5.4.5 added under section 5.4.

No conflicts with PR #220/#222/#223 — disjoint file sets.

---

chinmay28 (commented):
## Review

Faithful port of the 5.4.2 VM-sort pattern to images, including the size-as-third-axis convenience for "biggest first" disk-space audits. Implementation mirrors `SortVMs` shape-for-shape — same `sort.SliceStable` skeleton, same id-tiebreak contract, same `parseXxxSort` surface emitting 400 `invalid_sort` / `invalid_order`. Roadmap row 5.4.5 is well-motivated.

### Verified locally

- `CGO_ENABLED=1 go test -tags libvirt_dlopen -count=1 -short -skip 'TestStreamEvents_ReplayOverflowReturnsGone|TestBusSlowSubscriberIsDropped' ./...` — all packages green, including the 6 new unit cases (`pkg/types/image_sort_test.go`), 7 API integration cases (`TestListImages_Sort*`), and 4 CLI cases (`TestCLI_ImageList_Sort*`).

### Things I checked

- The `seedStoredImageFull` helper sidesteps `qemu-img` calls so the size + created_at axes can be tested deterministically without shelling out — clean way to control the test fixtures.
- `?tag=` filter runs **before** sort which is **before** pagination — `X-Total-Count` reflects the filtered population, matching the existing tag+pagination contract (and the new test `TestListImages_SortCombinesWithTagFilter` locks that in).
- Equal-size tiebreak in descending order flips the ID direction (img-2 before img-1 when both are 2048 bytes, descending size). That's the same behaviour `SortVMs` has and the right call — deterministic for any single query, and reversing the tiebreak alongside the order avoids the jarring "size goes down, then ID jumps backwards" jump at the equal-size boundary.
- The CLI's old hardcoded `created_at-asc` default is replaced by `id-asc` to match the API contract. Documented in the PR's "Notes" section and the roadmap entry — agreed it's the right call so the two layers don't drift.
- `setPage(1)` on sort/order change in the GUI prevents the user from landing on an empty page after re-sorting a multi-page list.

### Nits (non-blocking)

- The `cmp` helper in `tests/web/mock-server.js` uses `(a.created_at || "").localeCompare(b.created_at || "")` — works because the seeded fixtures are ISO-8601 strings, but is technically a lexicographic compare on the string form. Fine for the test fixtures; a future addition of fractional-second precision in the seeder would still work since ISO-8601 lexical order matches chronological order.
- `internal/cli/image.go` reuses the `switch ... default: return fmt.Errorf("invalid --sort %q: ...")` inline-validation block that 5.4.2 added to `internal/cli/vm.go`. If a fourth list endpoint grows the same shape, extracting a small `validateListSort(field, order, allowed []string)` helper would be worth it — but not yet.

---

---

PR #201, opened 2026-05-07
feat(templates): editable metadata + tag-filter list (2.2.10)

chinmay28:
## Summary

Closes the last "metadata edit" gap in roadmap section 2.2 by mirroring the description / tag-filter story already shipped for VMs / images / snapshots / port forwards onto **reusable VM templates**. Operators can now correct typos in a template's description, relabel templates as policies evolve, and find templates by tag without scrolling. Image / resources / name / network attachments stay immutable post-create — delete-and-recreate is the right answer for those (consistent with `Snapshot.Description` PATCH being deliberately omitted in #185).

PATCH semantics match the established VM convention exactly:

| input | meaning |
|---|---|
| empty `description` | no change |
| `null`/missing `tags` | no change |
| `"tags": []` | explicitly clear |

## What changed

**Backend**

- `pkg/types/template.go` — adds `TemplateUpdateSpec` with the documented PATCH semantics.
- `internal/storage/image.go` — `Manager.UpdateTemplate` with **no-op detection**: `UpdatedAt` only advances when something actually changed, so an empty patch leaves the bbolt record byte-identical.
- `internal/api/handlers_template.go`
  - `PATCH /api/v1/templates/{id}` — 200 with the patched record, 400 `invalid_spec` for over-long descriptions (`maxTemplateDescriptionLength = 1024`), 404 on unknown ID. Emits `template.updated` on the events bus when any field changed (no-op patches don't emit, mirroring the suggestion in PR #181's review).
  - `GET /api/v1/templates?tag=<tag>` — case-insensitive filter applied **before** pagination so `X-Total-Count` reflects the filtered population.
  - Tag normalization preserves "explicit clear" intent when `normalizeTags` collapses an all-blank slice to `nil` (same belt-and-suspenders pattern used in #181 for images).
- `internal/api/router.go` — registers `Patch("/{templateID}", s.withRequestBodyLimit(s.UpdateTemplate))`.

**CLI**

- `vmsmith template edit <id> [--description] [--tag ...] [--clear-tags]`
  - `--tag` and `--clear-tags` are mutually exclusive
  - At least one flag is required
  - `--clear-tags` is the explicit affordance the #181 review asked for, instead of the awkward `--tag ""` workaround
- `vmsmith template list --tag <tag>` — case-insensitive client-side filter for the in-process direct-CLI path.

**GUI**

[...]

PR #222, opened 2026-05-11
feat(webhooks): editable webhook config across API, CLI, and GUI (4.2.19)

chinmay28:
## Summary

Mirrors the 2.2.10 / 2.2.11 / 2.2.12 metadata-edit pattern for webhook receivers. The webhook subsystem ships create / list / get / delete / test, but every editable field — URL, signing secret, event-type filter, active flag — used to require **delete and re-create**, losing the `id`, the persisted delivery history, and any external bookkeeping pinned to that id.

This rounds out section **4.2 Event System & Notifications** so the webhook config surface and the rest of the resource-metadata edit surfaces (templates, snapshots, port forwards) stay symmetric. Roadmap row **4.2.19** marked ✅ Done.

## Behaviour

| Request | Status | Notes |
|---|---|---|
| `PATCH /webhooks/{id}` with `{"url": "https://new.example/x"}` | 200 | URL replaced, worker bounced |
| `PATCH /webhooks/{id}` with `{"secret": "rotated"}` | 200 | server-side rotation; response never echoes the secret |
| `PATCH /webhooks/{id}` with `{"event_types": ["vm.started"]}` | 200 | filter list replaced |
| `PATCH /webhooks/{id}` with `{"event_types": []}` | 200 | filter cleared → matches every event |
| `PATCH /webhooks/{id}` with `{"active": false}` | 200 | worker stopped, registration retained |
| `PATCH /webhooks/{id}` with `{}` | 400 | `noop_update` |
| `PATCH /webhooks/{id}` with `{"url": "ftp://nope"}` | 400 | `invalid_url` |
| `PATCH /webhooks/{id}` with `{"secret": ""}` | 400 | `missing_secret` (secrets can be rotated, never cleared) |
| `PATCH /webhooks/wh-missing` | 404 | `resource_not_found` |

Pointer-typed fields express "no change" via JSON-key absence. `event_types: []` is an **explicit** clear (subscribe-all). On any field-level change the in-memory delivery worker is **Unregister-then-Register**ed so the next event uses the new URL / secret / filter / active state without a daemon restart.

### Backend
- **`pkg/types/webhook.go`** — new `WebhookUpdateSpec{URL,Secret,EventTypes,Active *…}`. All pointers; semantics documented inline (omit = no change; `event_types: []` clears; empty `secret` rejected).
- **`internal/api/handlers_webhook.go`** — `UpdateWebhook` validates each field, persists the merged record, bounces the worker via `Unregister(id)` + `Register(updated)`. Includes a `stringSlicesEqual` helper so a "no-op" PATCH (every supplied value already matches the stored record) returns 200 without bouncing the worker.
- **`internal/api/router.go`** — `r.Patch("/{webhookID}", s.withRequestBodyLimit(s.UpdateWebhook))`.
- **`docs/openapi.yaml`** — new `PATCH /webhooks/{webhookID}` path + `WebhookUpdateSpec` schema.

### CLI
- **`internal/cli/webhook.go`** — `vmsmith webhook edit <id> [--url --secret --event-types --clear-event-types --active]`. Flag rules:
  - `--event-types` and `--clear-event-types` are mutually exclusive
  - `--active` is honored only when explicitly set (uses `cmd.Flags().Lookup().Changed`)
  - At least one editable flag must be present, else "no fields to update"

### Frontend
- **`web/src/api/client.ts`** — adds `webhooks.update(id, spec)` typed against the regenerated OpenAPI schema.
- **`web/src/api/generated/schema.d.ts`** — regenerated via `npm --prefix web run generate:api`.
- **`web/src/pages/Settings.jsx`** — adds an `EditWebhookModal` driven by a Pencil-icon button alongside Test / Delete. Pre-fills the current URL + event-type filter; secret field is a password input with placeholder "Leave blank to keep the current secret"; a "Subscribe to every event" checkbox clears the filter; an "Deliveries enabled" checkbox toggles `active`. Submit builds a minimal PATCH body — only fields the user actually changed are sent.

### Tests
- **API (`internal/api/handlers_webhook_test.go`, 12 new cases):** `TestUpdateWebhook_SetsURL`, `_RotatesSecret`, `_SetsEventTypes`, `_ClearsEventTypes`, `_TogglesActive`, `_RejectsEmptyBody`, `_RejectsInvalidURL`, `_RejectsEmptySecret`, `_NotFound`, `_NoOpWhenValueEqualsCurrent`, `_RejectsBadJSON`, `_ReactivatesStoppedWorker`. Each asserts the persisted record, the redacted response, and `Register` / `Unregister` calls on the fake manager.
- **CLI (`internal/cli/commands_test.go`, 8 new cases):** `TestCLI_WebhookEdit_SetsURL`, `_RotatesSecret`, `_TogglesActive`, `_SetsEventTypes`, `_ClearsEventTypes`, `_RejectsConflictingEventTypeFlags`, `_RejectsNoFlags`, `_PropagatesDaemonError`. Backed by an `httptest.Server`-driven `fakeWebhookDaemon` that captures the PATCH body and lets each test assert on what the CLI actually serialised.
- **Mock server (`tests/web/mock-server.js`):** new PATCH route mirroring the same validation (400 `noop_update` / `invalid_url` / `missing_secret`) so Playwright assertions actually exercise the round-trip.
- **GUI runner (`tests/web/run-gui-tests.js`, 2 new cases):** `edit webhook URL and event-type filter via PATCH`, `edit webhook can clear filter to subscribe-all`.
- **Playwright spec (`tests/web/gui.spec.js`):** mirror cases registered against the live mock server.

### Docs
- **`CLAUDE.md`** — REST API quick reference for `PATCH /webhooks/{id}` documenting pointer semantics, the `event_types: []` clear convention, the secret-rotation rule, and the worker-bounce on success. CLI mirror documented in the same row.
- **`docs/ROADMAP.md`** — new row **4.2.19** under section 4.2, marked ✅ Done.

## Test plan

- [x] `CGO_ENABLED=1 go test -tags libvirt_dlopen -count=1 -short -skip 'TestStreamEvents_ReplayOverflowReturnsGone|TestBusSlowSubscriberIsDropped' ./...` — all green (same two pre-existing flakes that #205/…/#221 skip)
- [x] `npm --prefix web run generate:api` — schema regenerates with the new PATCH path
- [x] `node tests/web/run-gui-tests.js` — 43 passed, 0 failed (includes the 2 new edit-webhook scenarios)

## Notes

- **Why not allow clearing the secret?** Empty `secret` would defeat receiver-side HMAC verification — every delivery would still carry `X-VMSmith-Signature: sha256=…`, but the signed payload would be `HMAC(key="", body)`, which is computable by anyone. Better to fail-fast with `missing_secret` than ship deliveries the receiver should reject.
- **Why bounce the worker on every change?** The `webhooks.Manager` worker captures the webhook record by value in `newWorker(m, *wh)`. There's no live "reload my config" path on the worker today, so the safe thing is the Unregister-then-Register dance. Cheap (the queue is per-worker and bounded), and the next event sees the new config. A future "hot reload" patch would just push the new record through a channel, but that's out of scope.
- **Why does `_NoOpWhenValueEqualsCurrent` not bounce the worker?** Each field-equality check guards the `changed = true` set. If every supplied value already matches what's persisted, we skip the `PutWebhook` + manager bounce and return 200 with the redacted record. Saves a worker restart cycle for clients that PATCH the same value defensively (e.g., a settings sync loop).
- **No conflict with PR #220 (VM search) or PR #221 (image sort).** This PR touches `pkg/types/webhook.go`, `internal/api/handlers_webhook*.go`, `internal/api/router.go`'s webhook subtree, `internal/cli/webhook.go`, `web/src/pages/Settings.jsx`, and the webhook block in `tests/web/mock-server.js` / `run-gui-tests.js`. The two open PRs target VMs and images respectively — no overlap.

---
_Generated by [Claude Code](https://claude.ai/code/session_01LZYebFyHPchccDaMimoVzo)_


chinmay28 (commented):
LGTM — clean mirror of the 2.2.10 / 2.2.11 / 2.2.12 metadata-edit pattern, adapted for the webhook subsystem's worker bounce. (Can't formally approve since I'm also the author; leaving as a review-comment.)

- `CGO_ENABLED=1 go test -tags libvirt_dlopen -count=1 -short -skip 'TestStreamEvents_ReplayOverflowReturnsGone|TestBusSlowSubscriberIsDropped' ./pkg/types/... ./internal/api/... ./internal/cli/... ./internal/vm/...` — all pass, including the 12 new `handlers_webhook_test.go` API cases and 8 new CLI cases.

- `WebhookUpdateSpec` uses pointer fields for "no change" — matches the 2.2.10 template-edit and 2.2.11 image-edit conventions. The `*[]string` for `event_types` makes the `nil = no change` vs `[] = clear` distinction explicit on the wire.
- `UpdateWebhook` validates each field before applying, computes `changed` field-by-field, and only persists+bounces the worker when something actually changed (`_NoOpWhenValueEqualsCurrent` test confirms). Worker bounce via `Unregister`+`Register` is the right approach given the worker captures the record by value in `newWorker`.
- Secret-empty is rejected with `missing_secret` — correct call. Empty secret would still produce a `X-VMSmith-Signature` header but with HMAC-SHA256 of the empty key, which is essentially "any party can forge".
- URL scheme validation matches `CreateWebhook` (http/https only).
- CLI: `webhook edit` uses `cmd.Flags().Lookup("...").Changed` to distinguish "not provided" from "passed as default" — this is the right idiom for cobra bool flags. Mutual-exclusion check between `--event-types` and `--clear-event-types` runs after parse so the diagnostic is clean.
- Frontend `EditWebhookModal` builds a minimal PATCH body — only fields the user actually touches are serialised. The "Subscribe to every event" checkbox correctly serialises `event_types: []` only when the webhook currently has a filter, avoiding a no-op PATCH that the API would reject.
- Roadmap row 4.2.19 added under section 4.2 — fills out the webhook surface alongside register / list / test / delete.

No conflicts with PR #220/#221/#223.

---

chinmay28 (commented):
## Review

Solid metadata-edit slice that closes the operational gap the PR description calls out (rotate a leaked secret / fix a typo'd URL / pause during an incident without losing the id or the persisted delivery history). Lines up with roadmap row **4.2.19** and follows the 2.2.10 / 2.2.11 / 2.2.12 metadata-edit shape — pointer-typed `WebhookUpdateSpec`, nil-key = no change, `event_types: []` is the explicit clear, secret-can-rotate-but-not-empty.

### Verified locally

- `CGO_ENABLED=1 go test -tags libvirt_dlopen -count=1 -short -skip 'TestStreamEvents_ReplayOverflowReturnsGone|TestBusSlowSubscriberIsDropped' ./...` — all packages green (12 new API cases, 8 new CLI cases)

### Things I checked

- **Worker bounce semantics:** `Unregister` is idempotent, and `Register` with Active=false is intentionally a no-op (comment at handlers_webhook.go:270-273 spells it out). So toggling `active=false` does correctly stop deliveries; toggling back to true re-starts the worker. `TestUpdateWebhook_ReactivatesStoppedWorker` locks that round-trip in.
- **No-op detection:** the `changed` flag gates both the `PutWebhook` write and the worker bounce, so a defensive `PATCH {"url": "<same value>"}` doesn't churn the worker. `TestUpdateWebhook_NoOpWhenValueEqualsCurrent` verifies this — nice.
- **Secret cannot be cleared:** `missing_secret` on empty trimmed input is the right call given HMAC verification would otherwise pass with `HMAC(key="", body)`. The PR description's reasoning is correct.
- **Redacted responses:** both the no-op early return and the success path call `redactWebhook(&updated)` — secret never leaks back to the client.
- **CLI's `--active` flag uses `f.Changed`** rather than the raw bool value. Without that the default `true` would silently re-enable any webhook the user edited for an unrelated field. Good catch on the implementer's part.
- **`--event-types` vs `--clear-event-types` mutually exclusive** at the CLI; same intent the JSON shape carries (nil = no change vs `[]` = clear).
- **404 for unknown id, 400 for empty body, 413 on oversized body** all wired correctly via the existing `requireWebhookStore` / `isRequestTooLarge` helpers.

### Nits (non-blocking)

- `stringSlicesEqual` is order-sensitive but `normalizeEventTypes` is order-preserving (insertion order, not sorted). So a `PATCH {"event_types": ["b", "a"]}` against a stored `["a", "b"]` will count as a real change and bounce the worker even though the filter set is semantically the same. Sorting the slice inside `normalizeEventTypes` would canonicalise the comparison; equally `slices.Equal` from stdlib could replace the helper (Go 1.22+; PR #201 had the same observation about a local copy of this helper). Either way, no production impact — just a spurious worker bounce on reorderings.
- `handlers_webhook.go:225` accepts the URL on a `HasPrefix("http://"|"https://")` check. This is the same minimal scheme check `CreateWebhook` uses (line 88), so the contract is consistent. A future tightening (full `url.Parse` + reject userinfo + reject fragments) could live in a shared validator.

---

---

PR #227, opened 2026-05-12
feat(images): free-text search filter across API, CLI, and GUI (5.4.9)

chinmay28:
## Summary

`GET /api/v1/images` already supports `?tag=` plus the new `?sort=` / `?order=` from #221. Operators with many images still had to scroll the list to find a specific one. This adds a `?search=` query param that does a case-insensitive substring match across **name, description, and tags** — the same shape `?search=` already takes on VMs (#220, 2.2.13) and events (#225, 4.2.20). Roadmap row **5.4.9** marked ✅ Done.

ID is intentionally **excluded** from the match — opaque `img-<unix-nano>` strings rarely match a useful operator query and would produce noisy false positives when the substring happens to look numeric.

## Behaviour

| Request | Status | Notes |
|---|---|---|
| `?search=rocky` (name substring) | 200 | matches `rocky9-base`, `rocky9-rc` |
| `?search=hardened` (description substring) | 200 | matches images with `hardened` in description |
| `?search=storage` (tag substring) | 200 | matches images with any tag containing `storage` |
| `?search=ROCKY` (uppercase needle) | 200 | case-insensitive |
| `?search=%20%20alpha%20%20` | 200 | whitespace-trimmed |
| `?search=rocky&tag=prod&sort=name` | 200 | filters compose additively (intersection); X-Total-Count reflects the post-search count |

The handler lowercases + trims the needle once before delegating to the shared predicate. Caller-lowercases-needle invariant is locked in by a unit test so future call sites can't silently regress.

### Backend (API + shared predicate)
- **`pkg/types/image_search.go`** — new `ImageMatchesSearch(img, query)` predicate mirroring `VMMatchesSearch`. Empty query matches every non-nil image; nil image never matches.
- **`internal/api/handlers_image.go`** — `ListImages` now applies `?search=` after `?tag=` and before sort+pagination so `X-Total-Count` reflects the post-search population.
- **`docs/openapi.yaml`** — documents the new `search` query parameter on `/images`.

### CLI
- **`internal/cli/image.go`** — `vmsmith image list --search <q>` with the same trim + lowercase + `ImageMatchesSearch` pipeline.

### Frontend
- **`web/src/pages/ImageList.jsx`** — debounced (250 ms) search input with a `Search` icon and `X` clear-button. Committed query round-trips through the URL via `?search=`. Page resets to 1 on filter change.
- **`web/src/api/client.ts`** — `images.list({...})` now accepts `search`. Regenerated `web/src/api/generated/schema.d.ts` via `npm --prefix web run generate:api`.
- **`tests/web/mock-server.js`** — mirrors the same predicate so Playwright reductions are real.

### Tests
- **Unit (`pkg/types/image_search_test.go`, 8 cases):** empty / nil / name / description / tag / caller-lowercases contract / empty-description / no-match.
- **API integration (`internal/api/api_test.go`, 8 cases):** name / description / tag / case-insensitive / trim-whitespace / no-match / composes-with-`?tag=` / composes-with-`?sort=`.
- **CLI (`internal/cli/commands_test.go`, 5 cases):** name / description / tag / no-match / composes-with-`--tag`.
- **Playwright (`tests/web/gui.spec.js`, 2 cases):** `search input filters the image list and round-trips through the URL`; `search input matches no images when query has no hits`.

### Docs
- **`CLAUDE.md`** — REST API quick reference for `GET /images` extended to document `?search=`, the case-insensitive contract, the whitespace-trim, the ID exclusion, and additive composition with `tag` / `sort` / `order` / pagination. CLI mirror documented in the same row.
- **`docs/ROADMAP.md`** — new row **5.4.9** under section 5.4, marked ✅ Done.

## Test plan

- [x] `CGO_ENABLED=1 go test -tags libvirt_dlopen -count=1 -short -skip 'TestStreamEvents_ReplayOverflowReturnsGone|TestBusSlowSubscriberIsDropped' ./...` — all green (same two pre-existing flakes that #205…#226 skip)
- [x] `npm --prefix web run generate:api` — schema regenerates with the new `search` query param

## Notes

- **Why not also match against image ID?** IDs are opaque `img-` strings — operators rarely type them from memory, and the all-numeric tail produces noisy false positives when the search term looks numeric. Same rationale as 2.2.13 (VMs) and 4.2.20 (events).
- **Why a shared `pkg/types` helper?** Keeps the predicate next to `SortImages` so the API handler and CLI both share a single contract. Without it the two layers could drift on edge cases like trim/lowercase semantics.
- **Why debounced in the GUI?** A keystroke-per-fetch pattern fans out one request per character. 250 ms is the sweet spot — long enough to coalesce a fast typist, short enough to feel live. `searchInput` keeps the live `<input>` state; `searchFilter` is the committed query that drives `useFetch`.

https://claude.ai/code/session_011ijdyLa5dAEfxPdN9YiQtU

---
_Generated by [Claude Code](https://claude.ai/code/session_011ijdyLa5dAEfxPdN9YiQtU)_


chinmay28 (commented):
## Review: LGTM ✅ (cannot self-approve)

Clean mirror of the established `?search=` pattern (VMs 2.2.13, events 4.2.20) — extends it to images with the additional `tags` haystack slot, which matches the data model.

### Verified locally
- `CGO_ENABLED=1 go test -tags libvirt_dlopen -count=1 -short -skip 'TestStreamEvents_ReplayOverflowReturnsGone|TestBusSlowSubscriberIsDropped' ./...` — all green
- CI: both Backend (1.22) and Frontend mock GUI jobs passing

### What's right
- `ImageMatchesSearch` lives next to `SortImages` in `pkg/types`, keeping the predicate shared between API and CLI — same pattern as snapshot/event/VM search.
- Order of filter application in `ListImages` (tag → search → sort → pagination) is consistent with the architecture's "filters compose additively" contract; `X-Total-Count` correctly reflects post-search population (`TestListImages_FilterBySearch_CombinesWithSort` locks it in).
- Excluding ID from the haystack is the right call — same rationale as 2.2.13/4.2.20; the `Caller_Lowercases` test locks the contract in.
- CLI mirrors the pattern (lowercase + trim before predicate, then sort, then paginate). `--tag` + `--search` composition tested.
- Frontend: 250 ms debounce, URL round-trip via `?search=`, page-reset on filter change, empty-state surfaces the needle. Consistent with VMList/Dashboard search UX.

### Minor (non-blocking)
- `web/src/api/client.ts:178` — the `as any` cast on the query params is technically unnecessary now that the generated schema knows `search` (visible in the schema.d.ts diff). It's preexisting boilerplate from when `tag` was first added though, so consistent with neighbors — fine to leave.

Roadmap row 5.4.9 ✅ Done is accurate. Merge-ready from my side.

---

PR #230, opened 2026-05-13
feat(templates): free-text search filter across API, CLI, and GUI (5.4.12)

chinmay28:
## Summary

`GET /api/v1/templates` already supports `?sort=` / `?order=` from #224 (5.4.7) and `?tag=` from 2.2.10. Operators with many templates still had to scroll the Create-VM dropdown to find a specific one. This adds a `?search=` query param that does a case-insensitive substring match across **name, description, and tags** — the same shape `?search=` already takes on VMs (#220, 2.2.13), events (#225, 4.2.20), images (#227, 5.4.9), snapshots (#228, 5.4.10), and port forwards (#229, 5.4.11). Roadmap row **5.4.12** marked ✅ Done.

ID, `image`, `default_user`, and the network attachment fields are intentionally **excluded** from the haystack — they describe what the template produces, not the template itself, and including them produces noisy false positives (e.g. an `image=rocky9.qcow2` template matching every search containing "rocky"). Same rationale as 2.2.13 / 4.2.20 / 5.4.9 / 5.4.10 / 5.4.11 exclude ID.

## Behaviour

| Request | Status | Notes |
|---|---|---|
| `?search=rocky` (name substring) | 200 | matches templates whose name contains `rocky` |
| `?search=hardened` (description substring) | 200 | matches templates whose description contains `hardened` |
| `?search=storage` (tag substring) | 200 | matches templates carrying a tag containing `storage` |
| `?search=ROCKY` (uppercase needle) | 200 | case-insensitive |
| `?search=%20%20alpha%20%20` | 200 | whitespace-trimmed |
| `?search=needle-not-present` | 200 | empty list, no error |
| `?search=rocky&tag=prod` | 200 | filters compose additively (intersection) |
| `?search=rocky&sort=name&order=asc` | 200 | composes with sort + pagination |

The handler lowercases + trims the needle once before delegating to the shared predicate.

### Backend (API + shared predicate)
- **`pkg/types/template_search.go`** — new `TemplateMatchesSearch(tpl, query)` predicate mirroring `SnapshotMatchesSearch` / `ImageMatchesSearch` / `VMMatchesSearch`. Empty query matches every non-nil template; nil template never matches.
- **`internal/api/handlers_template.go`** — `ListTemplates` now applies `?search=` after `?tag=` filter and before sort + pagination so `X-Total-Count` reflects the post-search population.
- **`docs/openapi.yaml`** — documents the new `search` query parameter on `/templates`.

### CLI
[...]

PR #229, opened 2026-05-12
feat(network): free-text search filter for port forwards (5.4.11)

chinmay28:
## Summary

`GET /api/v1/vms/{vmID}/ports` already supports `?sort=` / `?order=` from #226 (5.4.8). Operators with many port-forward rules on a single VM still had to scroll the list to find a specific one. This adds a `?search=` query param that does a case-insensitive substring match across **description, protocol, host_port, guest_port, and guest_ip** — the same shape `?search=` already takes on VMs (#220, 2.2.13), events (#225, 4.2.20), images (#227, 5.4.9), and snapshots (#228, 5.4.10). Roadmap row **5.4.11** marked ✅ Done.

The rule ID (`{vmID}/{host}`) and `vm_id` are intentionally **excluded** from the haystack — the ID literally embeds the host port (already covered) and `vm_id` is the URL scope, not a useful operator query. Same rationale as 2.2.13 / 4.2.20 / 5.4.9 / 5.4.10.

## Behaviour

| Request | Status | Notes |
|---|---|---|
| `?search=jumpbox` (description substring) | 200 | matches rules whose description contains `jumpbox` |
| `?search=udp` (protocol substring) | 200 | matches every UDP rule |
| `?search=8080` (host port substring) | 200 | matches host_port 8080 or any port whose decimal contains 8080 |
| `?search=192.168.100` (guest IP substring) | 200 | matches every rule on that subnet |
| `?search=JUMPBOX` (uppercase needle) | 200 | case-insensitive |
| `?search=%20%20web%20%20` | 200 | whitespace-trimmed |
| `?search=needle-not-present` | 200 | empty list, no error |
| `?search=web&sort=host_port` | 200 | filters compose additively |

The handler lowercases + trims the needle once before delegating to the shared predicate.

### Backend (API + shared predicate)
- **`pkg/types/portforward_search.go`** — new `PortForwardMatchesSearch(pf, query)` predicate mirroring `SnapshotMatchesSearch` / `ImageMatchesSearch` / `VMMatchesSearch`. Empty query matches every non-nil rule; nil rule never matches.
- **`internal/api/handlers_network.go`** — `ListPorts` now applies `?search=` after fetch and before sort, mirroring the snapshot/image handler shape.
- **`docs/openapi.yaml`** — documents the new `search` query parameter on `/vms/{vmID}/ports`.

### CLI
[...]

PR #228, opened 2026-05-12
feat(snapshots): free-text search filter across API, CLI, and GUI (5.4.10)

chinmay28:
## Summary

`GET /api/v1/vms/{vmID}/snapshots` already supports `?sort=` / `?order=` from #223 (5.4.6). Operators with many snapshots on a single VM still had to scroll the list to find a specific one. This adds a `?search=` query param that does a case-insensitive substring match across **name and description** — the same shape `?search=` already takes on VMs (#220, 2.2.13), events (#225, 4.2.20), and images (#227, 5.4.9). Roadmap row **5.4.10** marked ✅ Done.

The snapshot ID (`<vmID>/<name>`) and `vm_id` are intentionally **excluded** from the haystack — IDs are redundant with the name match (the ID literally embeds the name) plus `vm_id` is the URL scope, not a useful operator query. Same rationale as 2.2.13 / 4.2.20 / 5.4.9.

## Behaviour

| Request | Status | Notes |
|---|---|---|
| `?search=upgrade` (name substring) | 200 | matches `pre-upgrade`, `upgrade-rc` |
| `?search=hardening` (description substring) | 200 | matches snapshots whose description contains `hardening` |
| `?search=UPGRADE` (uppercase needle) | 200 | case-insensitive |
| `?search=%20%20alpha%20%20` | 200 | whitespace-trimmed |
| `?search=needle-not-present` | 200 | empty list, no error |
| `?search=upgrade&sort=name` | 200 | filters compose additively; `X-Total-Count` reflects the post-search count |

The handler lowercases + trims the needle once before delegating to the shared predicate. The caller-lowercases-needle invariant is locked in by a unit test so future call sites can't silently regress.

### Backend (API + shared predicate)
- **`pkg/types/snapshot_search.go`** — new `SnapshotMatchesSearch(snap, query)` predicate mirroring `VMMatchesSearch` / `ImageMatchesSearch`. Empty query matches every non-nil snapshot; nil snapshot never matches.
- **`internal/api/handlers_snapshot.go`** — `ListSnapshots` now applies `?search=` after fetch and before sort + pagination so `X-Total-Count` reflects the post-search population.
- **`docs/openapi.yaml`** — documents the new `search` query parameter on `/vms/{vmID}/snapshots`.

### CLI
- **`internal/cli/snapshot.go`** — `vmsmith snapshot list <vm-id> --search` with the same trim + lowercase + `SnapshotMatchesSearch` pipeline.

### Frontend
[...]

PR #231, opened 2026-05-13
feat(logs): free-text search filter across API and GUI (5.4.13)

chinmay28:
## Summary

`GET /api/v1/logs` already accepts `level`, `source`, `since`, and pagination filters. With a live daemon, the log ring buffer can quickly fill with hundreds of entries that all share the same level/source — making "find that specific request" a scroll-and-squint problem. This adds a `?search=` query param that does a case-insensitive substring match across each entry's **message, source, level, and every value in the structured fields map** — the same shape `?search=` already takes on VMs (#220, 2.2.13), events (#225, 4.2.20), images (#227, 5.4.9), snapshots (#228, 5.4.10), port forwards (#229, 5.4.11), and templates (#230, 5.4.12). Roadmap row **5.4.13** marked ✅ Done.

Field **keys** are intentionally **excluded** from the haystack — they are a small repeating vocabulary (`vm_id`, `method`, `error`, …) and including them produces noisy matches against operator-supplied values that happen to share a substring with a key name. Timestamps are also excluded — operators use the existing `since` filter for time scoping.

## Behaviour

| Request | Status | Notes |
|---|---|---|
| `?search=listening` (message substring) | 200 | matches entries whose message contains `listening` |
| `?search=daemon` (source substring) | 200 | matches every entry whose source contains `daemon` |
| `?search=warn` (level substring) | 200 | matches every warn-level entry |
| `?search=vm-12345` (field-value substring) | 200 | matches every entry carrying that vm id in any field value |
| `?search=DAEMON` (uppercase needle) | 200 | case-insensitive |
| `?search=%20%20listening%20%20` | 200 | whitespace-trimmed |
| `?search=listening&level=warn` | 200 | filters compose additively (intersection) |

The handler lowercases + trims the needle once before delegating to the shared predicate.

### Backend (predicate + API)
- **`internal/logger/search.go`** — new `EntryMatchesSearch(entry, query)` predicate next to `Entry`. The predicate lives in `internal/logger` (not `pkg/types`) because `Entry` itself is private to the logger package.
- **`internal/api/handlers_logs.go`** — `GetLogs` now lowercases + trims `?search=` and applies the predicate after the `source` filter and before pagination so `X-Total-Count` reflects the post-search population. Adds `strings` to the import set.
- **`docs/openapi.yaml`** — documents the new `search` query parameter on `/logs` (including the field-key exclusion rationale and additive composition contract).

### Frontend
- **`web/src/pages/LogViewer.jsx`** — debounced (250 ms) search input above the log table with a `Search` icon and `X` clear-button. The committed query round-trips through the URL via `?search=` so the filtered view is shareable / bookmarkable / survives a refresh. When no entries match, the empty-state surfaces the needle (`No log entries match "..."`) so it's clear the list is filtered.
- **`web/src/api/client.ts`** — `logs.list({...})` now accepts `search`. Regenerated `web/src/api/generated/schema.d.ts` via `npm --prefix web run generate:api`.
- **`tests/web/mock-server.js`** — mirrors the same predicate (including the field-key exclusion) so Playwright reductions exercise the real filter shape.

### CLI

Deferred. There is no existing `vmsmith logs` command and adding a daemon-API-backed list command from scratch would expand scope (it would need its own `--level` / `--source` / `--since` / pagination flag set). The API surface is the binding contract here; the GUI is the primary consumer of the in-memory ring buffer. Tracked as a follow-up.

### Tests
- **Unit (`internal/logger/search_test.go`, 9 cases):** empty / message / source / level / field-value / field-keys-excluded / caller-lowercases contract / no-match / empty-fields-handled.
- **API integration (`internal/api/api_test.go`, 7 cases):** message substring / case-insensitive / whitespace-trim / no-match / field-value-matches / composes-with-level / total-count-reflects-filtered.
- **Playwright (`tests/web/gui.spec.js`, 4 cases):** `search input filters the log table and round-trips through the URL`; `search input matches against structured field values`; `search shows empty-state hint when no entries match`; `search clear button restores the unfiltered view`.

### Docs
- **`CLAUDE.md`** — REST API quick reference for `GET /logs` extended to document `?search=`, the case-insensitive contract, the whitespace-trim, the field-key exclusion, and additive composition with `level` / `source` / `since` / pagination.
- **`docs/ROADMAP.md`** — new row **5.4.13** under section 5.4, marked ✅ Done.

## Test plan

- [x] `CGO_ENABLED=1 go test -tags libvirt_dlopen -count=1 -short -skip 'TestStreamEvents_ReplayOverflowReturnsGone|TestBusSlowSubscriberIsDropped' ./...` — all green (api, cli, config, console, daemon, events, logger, metrics, network, storage, store, vm, webhooks, pkg/types, pkg/version)
- [x] `npm --prefix web run generate:api` — schema regenerates with the new `search` query param
- [ ] GitHub Actions to confirm Playwright (cached Chromium version mismatch prevented local run; CI installs the matching version)

## Notes

- **Why exclude field *keys* from the haystack?** Keys are a small, repeating vocabulary that VMSmith emits across thousands of log lines: `vm_id`, `method`, `path`, `status_code`, `error`, `duration_ms`, `addr`. A search for `"method"` would otherwise match nearly every API request entry regardless of the value the operator actually cares about. Values, by contrast, are operator-supplied (or daemon-supplied) substrings that carry the signal — `vm-12345`, `POST`, `/api/v1/vms`, `iptables not available` — exactly what an operator types into a "find this log line" box.
- **Why exclude timestamps?** They are already scoped via the existing `since` filter, and a substring match on `"2026-05-13"` would match every entry from today, producing no signal.
- **Why does the predicate live in `internal/logger` instead of `pkg/types`?** `Entry` is package-private to `internal/logger`; the API handler imports the logger package already, so co-locating the predicate keeps it close to the type it operates on. The shared "across API and CLI" predicate convention in `pkg/types/` exists for the wire types that both layers consume; `Entry` doesn't fit that shape.
- **Closes the 5.4 search-filter sextet** alongside VMs (2.2.13), events (4.2.20), images (5.4.9), snapshots (5.4.10), port forwards (5.4.11), and templates (5.4.12). Every list endpoint VMSmith ships now has a symmetric `?search=` filter with consistent trim / case / composition semantics.

---
_Generated by [Claude Code](https://claude.ai/code/session_018L9rvWxxxyJVuTSyfHsuKL)_


chinmay28 (commented):
Reviewed end-to-end and verified locally (cannot APPROVE because GitHub treats me as the PR author).

**Architecture & roadmap alignment:** Closes the 5.4 search-filter sextet. The predicate signature (`EntryMatchesSearch(entry, lowercased)`) mirrors `VMMatchesSearch`, `ImageMatchesSearch`, `SnapshotMatchesSearch`, `PortForwardMatchesSearch`, `TemplateMatchesSearch`, `EventMatchesSearch` — caller-trims/lowercases contract is consistent. Locating the predicate in `internal/logger` (instead of `pkg/types`) is the right call since `Entry` is package-private — the rationale is documented inline.

**Design choices look correct:**
- Field **keys** excluded from haystack — agree, `vm_id` / `method` / `error` would generate noise.
- Timestamps excluded — agree, already covered by `since`.
- Filter applied after `source` and before pagination — `X-Total-Count` correctly reflects post-search count.
- In-place slice filter (`entries[:0]`) avoids extra allocation.

**Verified locally:**
- `CGO_ENABLED=1 go test -tags libvirt_dlopen -count=1 -short ./...` — all packages green (api, cli, config, console, daemon, events, logger, metrics, network, storage, store, vm, webhooks, pkg/types, pkg/version)
- `npm --prefix web run build` — clean
- New unit tests (9) and integration tests (7) pass; all positive/negative paths covered including caller-lowercases contract.

**Frontend UX:** 250 ms debounce, URL round-trip via `?search=`, empty-state surfaces the needle, clear button — consistent with the other search filters. Mock-server mirrors the field-key exclusion so Playwright exercises the real shape.

**Docs:** OpenAPI, CLAUDE.md quick-ref, and ROADMAP row 5.4.13 all updated and accurate.

LGTM — recommend merge.

---

PR #233, opened 2026-05-13
feat(webhooks): free-text search filter across API, CLI, and GUI (5.4.14)

chinmay28:
## Summary

Closes the symmetric "list + search" surface on the final list endpoint. Every other list — VMs (`2.2.13`), events (`4.2.20`), images (`5.4.9`), snapshots (`5.4.10`), port forwards (`5.4.11`), templates (`5.4.12`), logs (`5.4.13`) — already shipped a case-insensitive substring filter. Webhooks were the last hold-out, leaving operators with a manual scroll through the Settings page on noisy clusters.

Roadmap row **5.4.14** marked ✅ Done.

### API
- **`pkg/types/webhook_search.go`** — new `WebhookMatchesSearch(wh, lowercased)` predicate. Lowercased haystack; caller is responsible for lowercasing the needle once, mirroring the existing 5.4.x contracts.
- **`internal/api/handlers_webhook.go`** — `ListWebhooks` lowercases + trims the needle once after the existing chronological sort, applies the predicate while building the redacted response slice. Empty/absent `?search=` short-circuits to the unfiltered set; non-matching entries are skipped before redaction so the response shape is identical to the unfiltered path.

### CLI
- **`internal/cli/webhook.go`** — `vmsmith webhook list --search <q>` trims + lowercases + URL-encodes the needle and forwards it as `?search=`. Empty values omit the param entirely so the daemon returns every webhook. Help text describes the haystack + exclusions.

### Frontend
- **`web/src/pages/Settings.jsx`** — adds a 250 ms-debounced search input above the webhook table with a `Search` icon and `X` clear button. The committed `searchFilter` drives the `useFetch` dependency; the live `searchInput` is what the user types. Filter round-trips via `?search=` on the URL so a bookmark replays it. The empty state copy branches on whether a search was active (`"No webhooks match your search"`) vs. the existing genuine-empty `"No webhooks registered"`.
- **`web/src/api/client.ts`** — `webhooks.list({ search })` forwards the new query param via the regenerated OpenAPI types.

### Haystack scope
Intentionally **excluded** from the match (lock-in via dedicated unit tests):

| Field | Why |
|---|---|
| `secret` | Never reaches the API anyway (redacted on every read). Including it would also be a small information leak — a needle that hits would confirm a substring of the secret. |
| `id` | Opaque `wh-<unix-nano>` strings — nobody types those from memory; matching produces noisy numeric false positives. |
| `last_error` | Per-attempt operator noise that changes between retries. The status badge / `last_status` is the right place to surface failure detail, not a free-text search predicate. |

### Tests

- **Unit (`pkg/types/webhook_search_test.go`, 9 cases):** `TestWebhookMatchesSearch_EmptyQueryMatchesAll`, `_NilWebhookNeverMatches`, `_URLSubstring`, `_EventTypeSubstring`, `_CallerLowercasesQuery`, `_NoMatchReturnsFalse`, `_SecretNotInHaystack`, `_IDNotInHaystack`, `_LastErrorNotInHaystack`, `_EmptyEventTypesHandled`.
- **API integration (`internal/api/handlers_webhook_test.go`, 7 cases):** `TestListWebhooks_FilterBySearch_MatchesURL`, `_MatchesEventType`, `_IsCaseInsensitive`, `_TrimsWhitespace`, `_EmptyQueryReturnsAll`, `_NoMatchReturnsEmpty`, `_SecretAndLastErrorNotInHaystack`. All assert the filter response retains the standard redaction.
- **CLI (`internal/cli/commands_test.go`, 4 cases):** `TestCLI_WebhookList_ForwardsSearchParam`, `_TrimsAndLowercasesSearch`, `_EmptySearchOmitsParam`, `_NoMatchPrintsEmptyMessage`.
- **Mock server (`tests/web/mock-server.js`):** `/webhooks` GET handler mirrors the same predicate, including the field exclusions, so Playwright assertions match production behaviour.
- **Playwright (`tests/web/gui.spec.js`, 2 scenarios):** `search input filters the webhook list and round-trips through the URL` (URL substring, URL round-trip, clear button, event-type substring); `search shows empty state when no webhooks match`.
- **JSDOM (`tests/web/run-gui-tests.js`, 1 scenario):** `webhook search input filters the list and matches URL or event types` — covers all three behaviours plus URL round-trip in a single mock-server flow.

### Docs
- **`docs/openapi.yaml`** documents the new `search` query parameter on `GET /webhooks` with the explicit exclusion list.
- **`CLAUDE.md`** REST quick-reference for `GET /webhooks` gains the search filter description.
- **`docs/ROADMAP.md`** adds row **5.4.14** under section 5.4, marked ✅ Done. Acknowledges this rounds out the symmetric search surface across every list endpoint.

## Behaviour

| Request | Status | Notes |
|---|---|---|
| `GET /webhooks?search=audit` | 200 | URL substring match |
| `GET /webhooks?search=vm.started` | 200 | event-type substring match |
| `GET /webhooks?search=CAMELCASE` | 200 | case-insensitive (lowercases needle) |
| `GET /webhooks?search=%20%20audit%20%20` | 200 | leading/trailing whitespace trimmed |
| `GET /webhooks?search=` | 200 | empty needle returns full set |
| `GET /webhooks?search=needle-not-anywhere` | 200 | empty array on no match |
| `GET /webhooks?search=<secret-substring>` | 200 | empty array — secret is not in the haystack |

## Test plan

- [x] `CGO_ENABLED=1 go test -tags libvirt_dlopen -count=1 -short -skip 'TestStreamEvents_ReplayOverflowReturnsGone|TestBusSlowSubscriberIsDropped' ./...` — all green
- [x] `CGO_ENABLED=1 go test -tags libvirt_dlopen -count=1 -race -short … ./pkg/types/... ./internal/api/... ./internal/cli/...` — race-clean
- [x] `npm --prefix web run generate:api` — regenerates with new `search` param on `/webhooks`
- [x] `gofmt -l` over touched files — clean
- [ ] GitHub Actions confirms backend + frontend jobs

## Notes

- **Why no CLI sort flag?** This slice intentionally mirrors only the `--search` half of the 5.4.x pattern. Webhook list ordering is currently deterministic by `created_at`; adding a sort surface is a separate concern (and the list is typically too small to justify the API breadth).
- **Symmetric pattern.** The implementation is byte-for-byte parallel to 5.4.9 (images) and 5.4.12 (templates) — same predicate shape, same `lower(trim(?search=))` handler pipeline, same 250 ms debounce + URL round-trip in the UI. This keeps future maintenance straightforward: one mental model for every list filter.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Q8guXGsJ5RYwvSQkXE4Nmp)_


chinmay28 (commented):
Nice piece of work — the contract is symmetric with the rest of the 5.4.x search slice, the secret/ID/last_error exclusions are well-justified and well-tested, and the `WebhookMatchesSearch` predicate is clean.

There's one bug that's reproducible locally and is what's failing CI's "Frontend build and mock GUI tests" job.

### Bug: undefined `q` in `tests/web/mock-server.js`

In `tests/web/mock-server.js` the new `GET /api/v1/webhooks` block references an undefined `q`:

```js
if (p === "/api/v1/webhooks" && method === "GET") {
    const needle = (q.get("search") || "").trim().toLowerCase();   // <-- q is not defined
    ...
}
```

Every other handler in this file resolves the query string via `url.searchParams.get(...)` (e.g. lines 140–142, 393–402). Because `q` is undefined, the very first `GET /webhooks` issued by the Settings page throws `ReferenceError: q is not defined` inside the async handler, which crashes the mock server. That is why the existing `settings page lists, creates, tests, and deletes webhooks` test now times out waiting for `[data-testid^="webhook-row-"]`, and why the next test in the file fails with `net::ERR_CONNECTION_REFUSED at http://localhost:4173/` — the server is dead by then.

Fix is a one-liner:

```js
const needle = (url.searchParams.get("search") || "").trim().toLowerCase();
```

I confirmed locally:
- on `main` the GUI suite is `40 passed, 0 failed`.
- on this branch the suite fails on the existing webhooks test, then cascades.

### Minor nit (non-blocking)

- `web/src/api/client.ts` uses `as any` to coerce the query object:
  ```ts
  unwrap(apiClient.GET('/webhooks', { params: { query: { search: search || undefined } as any } }))
  ```
  The regenerated `schema.d.ts` already types `search?: string`, so the `as any` shouldn't be needed here. Other 5.4.x list calls don't carry one.

Once the mock-server typo is fixed and the GUI suite is green, this is good to merge — everything else (predicate, handler wiring, CLI flag, Settings UI, OpenAPI doc, ROADMAP entry, test coverage) lines up with the established 5.4.x pattern.

---

PR #234, opened 2026-05-13
feat(webhooks): bulk-delete API + CLI + GUI (2.3.10)

chinmay28:
## Summary

Closes the symmetric bulk-delete surface across every list endpoint. Webhooks were the only list endpoint without a bulk-delete companion — snapshots ship `prefix`, images ship `tag`, port forwards ship `protocol`, templates ship `tag` (`2.3.9`). On noisy clusters retiring a cohort of webhooks meant enumerating IDs or running N DELETE round trips.

Roadmap row **2.3.10** marked ✅ Done.

### API

- **`internal/api/handlers_webhook.go`** — new `BulkDeleteWebhooks` handler. Accepts `{"ids": [...]}` or `{"event_type": "..."}` (exactly one); validation matches the image / template bulk shapes byte-for-byte (`invalid_bulk_request` on both/neither, `invalid_request_body` / `request_too_large` on bad JSON / oversize body). Returns `{"results": [{id, success, code?, message?}]}` so partial failures surface in a single response. Each successful target also unregisters from the in-memory runtime manager so the next event isn't delivered to a tombstoned webhook.
- **`internal/api/router.go`** — registers `POST /api/v1/webhooks/bulk_delete` with the standard `withRequestBodyLimit` wrapper.

### Event-type selector semantics

A webhook matches `event_type=X` iff its `event_types` filter list contains X as an exact (trimmed) element. **Catch-all webhooks (empty `event_types`) are deliberately NOT swept by the categorical selector**: they fire on every event including X, but the operator intent for "retire all subscribers to vm.deleted" is explicit-membership cleanup, not blanket removal. Catch-alls must be deleted by id. This is locked in by a dedicated test (`_ByEventType` seeds a catch-all webhook and asserts it survives).

### CLI

- **`internal/cli/webhook.go`** — `vmsmith webhook delete --event-type <type>` (mutually exclusive with positional `<id>`); single-positional `vmsmith webhook delete <id>` is preserved unchanged. The bulk path posts to `/webhooks/bulk_delete` and prints per-target `OK …` / `FAIL …`, returning a non-zero exit when any target failed.

### Frontend

- **`web/src/api/client.ts`** — typed `webhooks.bulkDelete({ ids | event_type })` wrapper against the regenerated OpenAPI schema.
- **`web/src/pages/Settings.jsx`** — per-row + select-all checkboxes, a "Delete selected" button, and an "X of Y succeeded" result banner. Mirrors the ImageList bulk-delete UI so the two surfaces share a mental model.

### Tests

- **API integration (8 cases)**: `TestBulkDeleteWebhooks_ByIDs`, `_ByEventType` (covers catch-all exclusion + survival of vm.started-only subscriber), `_PartialFailure`, `_EmptyRequestRejected`, `_BothIDsAndEventTypeRejected`, `_EventTypeNoMatchEmptyResponse`, `_BadJSON`, `_DisabledWhenStoreUnset`.
- **CLI (7 cases)**: `TestCLI_WebhookDelete_NoArgsErrors`, `_BothIDAndEventTypeErrors`, `_SinglePositional`, `_EventTypeBulk` (asserts the new payload shape on the wire), `_EventTypeTrimsWhitespace`, `_EventTypeNoMatchPrintsMessage`, `_PartialFailureSurfaces`.
- **Mock server** (`tests/web/mock-server.js`): `/webhooks/bulk_delete` route enforcing the same 400 contract and catch-all exclusion.
- **Playwright** (`tests/web/gui.spec.js`, 3 scenarios): bulk-delete selected webhooks, select-all-then-sweep, "Delete selected" stays disabled with no selection.
- **JSDOM mock-GUI** (`tests/web/run-gui-tests.js`, 2 scenarios): bulk-delete selected + select-all sweep.

### Docs

- **`docs/openapi.yaml`** documents the new path + `BulkDeleteWebhooksRequest` / `BulkDeleteWebhooksResponse` / `BulkDeleteWebhookResult` schemas.
- **`CLAUDE.md`** REST quick reference gains a `POST /webhooks/bulk_delete` row.
- **`docs/ROADMAP.md`** adds row **2.3.10** marked ✅ Done.

## Behaviour

| Request | Status | Notes |
|---|---|---|
| `{"ids": ["wh-1", "wh-2"]}` | 200 | per-target results |
| `{"event_type": "vm.deleted"}` | 200 | matches webhooks whose `event_types` contains `vm.deleted` exactly |
| `{"event_type": "vm.deleted"}` with a catch-all webhook present | 200 | catch-all survives — sweep is explicit-membership only |
| `{"ids": ["wh-real", "wh-missing"]}` | 200 | partial failure: real=success, missing=`resource_not_found` |
| `{}` | 400 | `invalid_bulk_request` |
| `{"ids": ["wh-1"], "event_type": "x"}` | 400 | `invalid_bulk_request` |
| `{"event_type": "no.such.type"}` | 200 | `{results: []}` — empty match is not an error |
| `not-json` | 400 | `invalid_request_body` |

## Test plan

- [x] `CGO_ENABLED=1 go test -tags libvirt_dlopen -count=1 -short -skip 'TestStreamEvents_ReplayOverflowReturnsGone|TestBusSlowSubscriberIsDropped' ./...` — all green (api, cli, config, console, daemon, events, logger, metrics, network, storage, store, vm, webhooks, pkg/types, pkg/version)
- [x] `npm --prefix web run generate:api` — schema regenerates with the new `BulkDeleteWebhooks*` shapes
- [x] `gofmt -l` over touched files — clean
- [x] Manual smoke test against the mock-server: ids selector, event_type selector, catch-all exclusion, both-or-neither rejection — all behave as specified
- [ ] GitHub Actions confirms backend + frontend jobs

## Notes

- **Why `event_type` and not `tag`?** Webhooks don't have tags. The natural categorical for "retire a cohort" is the event filter list — operators typically want to bulk-cleanup "everything still subscribed to a deprecated event" after a rotation. `event_type` matches what they think.
- **Mirrors `/templates/bulk_delete` (2.3.9) exactly.** Same exactly-one-of contract, same per-target result shape, same 400 codes, same redaction-preserving response. This keeps the API surface predictable and lets future client SDKs share the bulk-delete generic shape.
- **Webhooks DELETE doesn't emit events.** Mirrored here: the bulk path also doesn't publish `webhook.deleted` (which doesn't exist as an event type). The single DELETE has always been silent; the bulk path stays consistent. Separate work if we want webhook lifecycle events.

---
_Generated by [Claude Code](https://claude.ai/code/session_01JPfVqh52LcxpC1tYxoxWmM)_


chinmay28 (commented):
## LGTM — clean closure of the bulk-delete symmetry gap (commenting since I can't self-approve)

This delivers the last missing bulk-delete companion (webhooks were the only list endpoint without one) and follows the established shape from `/images/bulk_delete` (2.3.6), `/snapshots/bulk_delete` (2.3.5), `/ports/bulk_delete` (2.3.7), and `/templates/bulk_delete` (2.3.9) byte-for-byte. Maps roadmap row **2.3.10** correctly.

### Verified locally on the branch
- `CGO_ENABLED=1 go test -tags libvirt_dlopen -count=1 -short -skip 'TestStreamEvents_ReplayOverflowReturnsGone|TestBusSlowSubscriberIsDropped' ./...` — all packages green (`internal/api 6.4s`, `internal/cli 4.7s`, `internal/webhooks 0.3s`, plus everything else).
- Targeted run of `TestBulkDeleteWebhooks_*` (8 cases) and `TestCLI_WebhookDelete_*` (7 cases) — all pass.
- CI on the PR is green on both backend + frontend jobs.

### What's right
- **Catch-all exclusion** is the correct design call and it's deliberately tested (`TestBulkDeleteWebhooks_ByEventType` seeds `wh-catchall` and asserts it survives). Sweeping catch-alls on `event_type=X` would silently kill global-listener pipelines the operator probably didn't mean to retire. Docs (handler godoc, `webhookExplicitlySubscribed` comment, OpenAPI description, `CLAUDE.md`, PR body) all say the same thing — anyone re-reading this in six months will know why.
- **Selector choice (`event_type` vs `tag`).** Webhooks have no tags; `event_types` is the only categorical operators see, and "retire every subscriber to a deprecated event" is the natural cleanup workflow. The PR body articulates this trade-off explicitly.
- **Validation matches sibling endpoints exactly.** `invalid_bulk_request` on both/neither; `invalid_request_body` on bad JSON; `request_too_large` on oversize body; `resource_not_found` per-target. SDK consumers and the test layer can share the same generic shape.
- **Runtime manager bounce.** `s.webhookManager.Unregister(id)` is called per success so the next event isn't delivered to a tombstoned webhook — same pattern the single `DELETE` already uses.
- **CLI single-positional path is preserved.** `vmsmith webhook delete <id>` still hits the existing `DELETE /webhooks/{id}` (HTTP 204) — only the new `--event-type` flag routes through the bulk endpoint. The "exactly one of" / "mutually exclusive" check is at the CLI layer too, so users get fast feedback before any network hop.
- **GUI mirrors ImageList semantically.** Per-row + select-all checkboxes, indeterminate state, "X of Y succeeded" banner, dismiss button — same affordances the operator already knows from Images.
- **Webhooks DELETE doesn't emit events** is consistent across single and bulk paths; the PR body calls this out so the absence isn't misread as a bug.

### Test coverage
- API: 8 cases including `_DisabledWhenStoreUnset` (503), `_BadJSON`, `_PartialFailure`, `_EventTypeNoMatchEmptyResponse`, plus the catch-all survival assertion.
- CLI: 7 cases covering no-args, both-set rejection, single-positional, event-type bulk wire shape, whitespace trim, no-match message, and partial-failure exit code.
- Mock-server: `/webhooks/bulk_delete` enforces the same 400 contract + catch-all exclusion.
- Playwright (3 scenarios): selected, select-all sweep, disabled-when-empty.
- JSDOM (2 scenarios): selected + select-all.

### Minor / non-blocking
- The unused `successes` accumulator at the bottom of `runWebhookBulkDeleteByEventType` is harmless (silenced with `_ = successes`) but could just be deleted next time the file is touched.
- Three pre-existing gofmt issues in the tree are not introduced here (`internal/config/config.go`, `internal/console/store.go`, etc. need a sweep — separate cleanup from this PR).

Approving in spirit (cannot self-approve as author).

---
_Reviewed via Claude Code_

---

PR #259, opened 2026-05-18
feat(vms): ?image= base-image filter across API, CLI, and GUI (5.4.22)

chinmay28:
## Summary

Extends `GET /api/v1/vms` with `?image=<name>` case-insensitive exact-match on `spec.image`, mirrored through `vmsmith vm list --image <name>` and a debounced "Filter by image…" input on the VMList page (URL round-trip via `?image=`). Whitespace is trimmed before comparison; empty value disables the filter.

Closes the operator query *"show me every VM built from `rocky9.qcow2`"* that the existing free-text `?search=` filter would also match fuzzily against any VM whose description or tag happens to contain the substring, and that `?tag=` cannot answer without pre-tagging every VM by its base image. Composes additively with the existing `?tag=` / `?status=` / `?search=` filters and the sort + pagination params so `X-Total-Count` reflects the post-filter population.

### API
- `internal/api/handlers_vm.go`: `ListVMs` trims + lowercases the `?image=` value once and applies `strings.EqualFold(vm.Spec.Image, imageFilter)` inside the existing filter loop alongside `status` / `tag` / `search` so the post-filter / pre-pagination `X-Total-Count` stays correct.

### CLI
- `internal/cli/vm.go`: adds `--image` to `vmListCmd` with the same trim + lowercase + predicate pipeline so the in-process CLI path matches the daemon exactly.

### Frontend
- `web/src/pages/VMList.jsx`: adds a 250 ms-debounced text input alongside the existing search box with an `X` clear button and URL round-trip via `?image=`; the page also resets to page 1 whenever the image filter changes so the operator never lands beyond the post-filter page count.
- `web/src/api/client.ts`: extends `vms.list({ image })` to forward the new query parameter against the regenerated OpenAPI schema.

### Docs
- `docs/openapi.yaml`: adds new `ImageFilter` parameter component and references it on `GET /vms`.
- `CLAUDE.md`: REST quick reference for `GET /vms` extended.
- `docs/ROADMAP.md`: 5.4.22 entry added and marked ✅ Done.

### Tests
- `tests/web/mock-server.js`: mirrors the predicate so Playwright can validate every branch.

## Coverage

- **API integration (7)** — `TestListVMs_FilterByImage_ExactMatch`, `_IsCaseInsensitive`, `_TrimsWhitespace`, `_EmptyIsNoOp`, `_NoMatch`, `_ComposesWithStatus`, `_ComposesWithSearch`, `_TotalCountReflectsFiltered`.
[...]

PR #258, opened 2026-05-18
feat(vms): auto_start/locked tristate filters across API, CLI, and GUI (5.4.21)

chinmay28:
## Summary

Extends `GET /api/v1/vms` with `?auto_start=true|false` and `?locked=true|false` tristate filters on the daemon-managed lifecycle flags (2.5.1 / 2.5.2), mirrored through `vmsmith vm list --auto-start --locked` and two new Any/Yes/No dropdowns on the VMList page (URL round-trip via `?auto_start=` / `?locked=`).

Each filter is tristate:
- absent / whitespace-only → filter disabled (returns every VM)
- `true` / `false` (case-insensitive, plus `1` / `0` aliases) → exact-match predicate against `Spec.AutoStart` / `Spec.Locked`
- anything else → 400 `invalid_auto_start` / `invalid_locked`

Composes additively with the existing `?tag=` / `?status=` / `?search=` filters and the sort + pagination surface so `X-Total-Count` reflects the post-filter population.

Closes the operator query "which VMs auto-start at daemon boot" and "which VMs are delete-protected" that the existing `?status=` filter (5.4.2) and `?tag=` filter cannot answer without round-tripping every VM through `GET /vms/{id}`. The per-flag dropdowns also serve as a discoverability surface for the 2.5.1 (auto-start) and 2.5.2 (locked) features they expose.

### API
- `internal/api/parse_tristate.go`: shared `parseTristateBoolParam` helper (whitespace-trim + lowercase + `true`/`false`/`1`/`0` accepted, anything else returns the typed `*types.APIError`).
- `internal/api/handlers_vm.go`: `ListVMs` parses both params up-front (400 fail-fast) and threads `(value, set)` predicates into the existing filter loop alongside `status` / `tag` / `search` so the post-filter `X-Total-Count` stays correct.

### CLI
- `internal/cli/parse_tristate.go`: mirrors the parser shape for the CLI (`parseCLITristateBool`).
- `internal/cli/vm.go`: adds `--auto-start` / `--locked` flags to `vmListCmd` with the same predicates inside the in-process filter loop.

### Frontend
- `web/src/api/client.ts`: `vms.list({ autoStart, locked })` forwards the typed query params against the regenerated OpenAPI schema.
[...]

PR #262, opened 2026-05-18
feat(vms): ?default_user= filter across API, CLI, and GUI (5.4.23)

chinmay28:
## Summary

Extends `GET /api/v1/vms` with `?default_user=` case-insensitive exact-match against `Spec.DefaultUser`, mirrored through `vmsmith vm list --default-user` and a debounced "Filter by default user…" input on the VMList page (URL round-trip via `?default_user=`). Whitespace is trimmed before comparison; empty value disables the filter.

Closes the operator query *"which VMs SSH in as `ubuntu` vs `root` vs `ec2-user`"* that the existing `?search=` filter cannot answer cleanly — `search` would also match any VM whose name, description, or tag substring happens to mention the username. Composes additively with the existing `?tag=` / `?status=` / `?search=` filters and the sort + pagination params so `X-Total-Count` reflects the post-filter population.

### API
- `internal/api/handlers_vm.go`: `ListVMs` trims + lowercases the `?default_user=` value once and applies `strings.EqualFold(vm.Spec.DefaultUser, defaultUserFilter)` inside the existing filter loop alongside `status` / `tag` / `search` so the post-filter / pre-pagination `X-Total-Count` stays correct.

### CLI
- `internal/cli/vm.go`: adds `--default-user` to `vmListCmd` with the same trim + lowercase + predicate pipeline so the in-process CLI path matches the daemon exactly.

### Frontend
- `web/src/pages/VMList.jsx`: adds a 250 ms-debounced text input next to the existing search box with an `X` clear button and URL round-trip via `?default_user=`; the page resets to page 1 whenever the filter changes.
- `web/src/api/client.ts`: extends `vms.list({ defaultUser })` to forward the new query parameter against the regenerated OpenAPI schema.

### Docs
- `docs/openapi.yaml`: adds new `DefaultUserFilter` parameter component and references it on `GET /vms`.
- `CLAUDE.md`: REST quick reference for `GET /vms` extended.
- `docs/ROADMAP.md`: 5.4.23 entry added and marked ✅ Done.

### Tests
- `tests/web/mock-server.js`: mirrors the predicate so the JSDOM/Playwright suites can exercise the new query path.

## Coverage

- **API integration (7)** — `TestListVMs_FilterByDefaultUser_ExactMatch`, `_IsCaseInsensitive`, `_TrimsWhitespace`, `_EmptyIsNoOp`, `_NoMatch`, `_ComposesWithStatus`, `_TotalCountReflectsFiltered`.
- **CLI (5)** — `TestCLI_VMList_FilterByDefaultUser_ExactMatch`, `_IsCaseInsensitive`, `_EmptyOmitsFilter`, `_NoMatch`, `_ComposesWithStatus`.
- **Mock server** — `?default_user=` predicate mirrored (case-insensitive exact-match against `spec.default_user`).

## Test plan

- [x] `CGO_ENABLED=1 go build ./...` clean (with `libvirt-dev` installed)
- [x] `CGO_ENABLED=1 go test -race -count=1 ./...` — every package passes
- [x] `npm run generate:api` produces the new `default_user` query parameter on the VMs route
- [x] Mock server end-to-end: `curl ?default_user=ubuntu` returns only ubuntu VMs; `?default_user=UBUNTU` matches case-insensitively; empty value returns every VM
- [ ] CI runs the new tests against the mock server

---
_Generated by [Claude Code](https://claude.ai/code/session_01UBtg4jpEKW63GHvFxp147d)_


chinmay28 (commented):
## Review — LGTM ✅ (cannot self-approve)

**Build & test:**
- `CGO_ENABLED=1 go test -count=1 ./internal/api/ ./internal/cli/ ./pkg/types/` — every package passes
- All 7 API + 5 CLI default_user-filter cases pass with verbose output (exact / case-insensitive / whitespace-trimmed / empty-no-op / no-match / composes-with-status / X-Total-Count).

**Architecture alignment:**
- Filter predicate is added inside the existing single-pass filter loop in `handlers_vm.go` and `cli/vm.go`, sharing the same trim+lowercase contract as `tag` / `status` / `search`.
- CLI mirrors the daemon predicate exactly — `vm list --default-user` and the over-HTTP path agree.
- Reuses `strings.EqualFold` against the already-lowercased filter (same minor stylistic note as #264 — functionally identical).
- Frontend page-reset side effect includes `defaultUserFilter`; URL round-trip via `?default_user=`; debounced input mirrors the `?search=` UX exactly.
- OpenAPI gets a proper `DefaultUserFilter` component, mirroring the existing `SearchFilter` / `StatusFilter` shape — cleaner than inlining the parameter at the path level (as #264 / #265 did).

**Roadmap alignment:** 5.4.23 entry is plausibly placed between 5.4.20 and the in-flight 5.4.21–5.4.25. Skipping numbers is fine since the merge order isn't deterministic.

Would have approved if I weren't the PR author.

---

chinmay28:
LGTM (cannot submit a formal approval as the PR author).

Tested locally on the PR head:

- `CGO_ENABLED=1 go build ./...` clean
- `CGO_ENABLED=1 go test -count=1 ./internal/api/... ./internal/cli/... ./pkg/types/...` — all pass
- Targeted `TestListVMs_FilterByDefaultUser_*` and `TestCLI_VMList_FilterByDefaultUser_*` — all 7 + 5 cases pass

Architectural review:

- `strings.EqualFold(vm.Spec.DefaultUser, defaultUserFilter)` wired into the existing filter loop preserves the post-filter `X-Total-Count` contract.
- The query semantics ("which VMs SSH in as `ubuntu` vs `root` vs `ec2-user`") are a real operator pain that `?search=` cannot answer precisely, so the filter has clear motivation.
- API + CLI + GUI surfaces all trim + lowercase once at the boundary; the debounced VMList input, URL round-trip via `?default_user=`, page-reset on change, and `X` clear-button match the established 5.4.x VMList UX.
- Mock server's predicate falls back to a top-level `default_user` if `spec.default_user` is missing — fine for the JSDOM/Playwright suites, defensive and not relied on by the real daemon. No concern.

Note: this PR overlaps with #258 (`?auto_start=` / `?locked=`) and #259 (`?image=`) on `internal/api/handlers_vm.go` and `internal/cli/vm.go`. Whichever lands second will need a trivial rebase, but the semantics are independent.

---

chinmay28 (commented):
Requesting changes (leaving as a COMMENT review since the API blocks self-REQUEST_CHANGES).

The mechanical implementation is solid — same trim+lowercase+EqualFold predicate slotted alongside status/tag/search, matching the 5.4.x family. Tests pass cleanly locally (`CGO_ENABLED=1 go build ./...` ✅, 7 API + 5 CLI cases all green). But there's a semantic gap I think needs addressing before this lands.

**Issue: `?default_user=root` won't match the most common case**

`internal/vm/lifecycle.go:112` (existing code, not this PR) is explicit:

> `// DefaultUser intentionally left empty here: empty means "use root".`

CLAUDE.md says the same:

> VMs use `root` by default; override per-VM with `VMSpec.DefaultUser` / `--default-user`

So in production, a "root SSH" VM has `Spec.DefaultUser == ""`, not `"root"`. The predicate `strings.EqualFold(vm.Spec.DefaultUser, defaultUserFilter)` will therefore *not* match those VMs when the operator queries `?default_user=root`. The PR description explicitly calls out *"which VMs SSH in as ubuntu vs **root** vs ec2-user"* as the motivating query, but the implementation can only answer it for VMs explicitly created with `--default-user root` (an unusual configuration).

The test suite hides this by seeding `DefaultUser: "root"` directly — `_FilterByDefaultUser_ExactMatch` and `_ComposesWithStatus` both do this — so the existing coverage doesn't exercise the "empty means root" branch at all. Worth adding a regression case either way.

**Suggested fix**

Normalize the comparison to treat empty `Spec.DefaultUser` as `"root"` in the predicate. Something like:

```go
effectiveUser := vm.Spec.DefaultUser
if effectiveUser == "" {
    effectiveUser = "root"
}
if defaultUserFilter != "" && !strings.EqualFold(effectiveUser, defaultUserFilter) {
    continue
}
```

That makes the filter answer the operator's mental model ("which VMs SSH in as X"), matches the runtime semantics already documented in `lifecycle.go:112`, and is symmetric across `cli/vm.go` and `handlers_vm.go` + mock-server. Add one API and one CLI test that seeds `Spec.DefaultUser: ""` and asserts `?default_user=root` matches.

**Also: needs rebase**

This branch was last merged with main at `4315f80`. Current `origin/main` (78a6849) has picked up PRs #266 / #263. The diff vs current main shows what looks like a revert of those docs additions — please rebase / merge latest main before merging so they don't get clobbered.

Happy to re-review once the empty-vs-root semantics are addressed.

---

chinmay28:
Addressed the review comment about empty `Spec.DefaultUser` being treated as `root`:

- `internal/api/handlers_vm.go`: predicate now normalises empty `Spec.DefaultUser` → `root` before the case-insensitive compare, so `?default_user=root` matches both explicit-root and unset VMs.
- `internal/cli/vm.go`: same normalisation in the in-process CLI path.
- `tests/web/mock-server.js`: mirror updated.
- New regression tests: `TestListVMs_FilterByDefaultUser_RootMatchesEmpty` (API) and `TestCLI_VMList_FilterByDefaultUser_RootMatchesEmpty` (CLI) seed an empty-DefaultUser VM, an explicit-root VM, and an ubuntu VM, then assert `?default_user=root` returns the first two.
- ROADMAP 5.4.23 entry and OpenAPI `DefaultUserFilter` description updated to call out the "empty = root" semantic.

Also merged latest `main` into the branch (resolving the conflict with PR #258 / #259 / #268). `CGO_ENABLED=1 go test ./internal/api/ ./internal/cli/` clean locally.

---

PR #265, opened 2026-05-19
feat(network): ?protocol=tcp|udp filter on port-forward list (5.4.25)

chinmay28:
## Summary

Extends `GET /api/v1/vms/{vmID}/ports` with `?protocol=<tcp|udp>` case-insensitive exact-match against the rule's transport protocol, mirrored through the CLI (`vmsmith port list <vm-id> --protocol <tcp|udp>`) and a new "Any protocol / TCP / UDP" dropdown on the VMDetail port-forwards card (URL round-trip via `?port_protocol=`).

Whitespace is trimmed before comparison; empty value disables the filter; anything other than tcp/udp returns 400 `invalid_protocol`. Mirrors the bulk_delete `protocol` selector that already lets operators wipe every rule of a given protocol so the list and bulk-action surfaces agree. Composes additively with the existing `?tag=` / `?search=` / `?sort=` / `?order=` and pagination params so `X-Total-Count` reflects the post-filter population.

Closes the operator query *"show me every UDP rule on this VM"* / *"audit the TCP listeners on the bastion"* that the existing `?search=` filter would cover fuzzily — substring-matching `tcp` would also match a `tcp-proxy` description on a UDP rule.

### API
- `internal/api/handlers_network.go`: `ListPorts` parses `?protocol=` once (trimmed + lowercased), validates against `types.ProtocolTCP` / `types.ProtocolUDP`, and applies the predicate between `?tag=` and `?search=` so the post-filter / pre-pagination `X-Total-Count` stays correct. Invalid values fail fast with 400 `invalid_protocol`.

### CLI
- `internal/cli/network.go`: adds `--protocol` to `port list` with the same trim + lowercase + whitelist pipeline. Invalid value returns `invalid --protocol "<value>": must be 'tcp' or 'udp'`.

### Frontend
- `web/src/pages/VMDetail.jsx`: adds a third `<select>` (`port-protocol-filter`) next to the existing sort dropdowns with "Any protocol / TCP / UDP" options, round-tripped through the URL via `?port_protocol=`; the page snaps back to page 1 whenever the protocol filter changes.
- `web/src/api/client.ts`: extends `ports.list({ protocol })` to forward the new query param against the regenerated OpenAPI schema.

### Docs
- `docs/openapi.yaml`: documents the new `protocol` query parameter and the 400 response.
- `CLAUDE.md`: REST quick reference for `GET /vms/{id}/ports` extended.
- `docs/ROADMAP.md`: 5.4.25 entry added and marked ✅ Done.

[...]

PR #266, opened 2026-05-19
docs(console): document proxy design and deployment guardrails

march26bot:
## Summary\n- document the shipped console ticket flow and intended websocket proxy design\n- add production guidance to keep raw VNC loopback-only and avoid firewall/proxy exposure\n- refresh deployment docs to reflect built-in API-key auth availability\n\n## Testing\n- docs-only change (direct inspection)\n\nCloses roadmap item 5.1.12.

PR #263, opened 2026-05-19
web: clarify webhook search copy

march26bot:
$## Summary\n- update Settings webhook search copy to mention description support\n- align the empty-state hint with the actual search haystack\n- assert the updated placeholder in the existing Playwright webhook search test\n\n## Testing\n- npx playwright test tests/web/gui.spec.js -g "search input filters the webhook list and round-trips through the URL|search shows empty state when no webhooks match" *(fails in this environment because the expected nav shell test id is missing before the test reaches the changed assertion)*

PR #268, opened 2026-05-19
feat(webhooks): ?event_type= filter across API, CLI, and GUI (5.4.26)

chinmay28:
## Summary

Extends `GET /api/v1/webhooks` with `?event_type=<type>` — a case-insensitive exact-match against entries in the webhook's `event_types` filter list, mirrored through `vmsmith webhook list --event-type` and a debounced "Filter by event type…" input on the Settings page (URL round-trip via `?event_type=`).

**Catch-all webhooks (empty `event_types`) are NOT matched**, mirroring the existing `bulk_delete` `event_type` selector semantics so the list and bulk-action surfaces agree on "explicit-membership" semantics: this is **"which webhooks listen for `vm.created`"**, not **"which webhooks will fire for this event"**. The same pattern PR #265 used for ports (5.4.25), which mirrored the bulk_delete `protocol` selector.

Closes the operator query *"audit every webhook subscribed to `vm.deleted` before retiring the event"* that the existing `?search=` filter (5.4.14) would also fuzzy-match against a description that happens to mention the event name. Composes additively with the existing `?tag=` / `?search=` / `?sort=` / `?order=` / pagination params so `X-Total-Count` reflects the post-filter population.

### API
- `internal/api/handlers_webhook.go`: `ListWebhooks` parses `?event_type=` (trim + lowercase) and applies a new `webhookSubscribedToEventType` predicate between `?tag=` and `?search=` so the post-filter / pre-pagination `X-Total-Count` stays correct. The helper is case-insensitive (unlike the bulk_delete-specific `webhookExplicitlySubscribed`, which is left untouched to preserve its existing contract) but shares the **explicit-membership / catch-all-excluded** semantics.

### CLI
- `internal/cli/webhook.go`: adds `--event-type` to `webhook list` with the same trim + lowercase + URL-encode pipeline. Empty values omit the param so the daemon returns every webhook.

### Frontend
- `web/src/pages/Settings.jsx`: adds a debounced 250 ms text input next to the existing search box with an `X` clear button and URL round-trip via `?event_type=`; the page resets to page 1 whenever the filter changes; empty-state copy branched for the new filter ("No webhooks explicitly subscribe to …").
- `web/src/api/client.ts`: extends `webhooks.list({ eventType })` to forward the new query parameter against the regenerated OpenAPI schema.

### Docs
- `docs/openapi.yaml`: documents the new `event_type` query parameter on `GET /webhooks`.
- `CLAUDE.md`: REST quick reference for `GET /webhooks` extended.
- `docs/ROADMAP.md`: 5.4.26 entry added and marked ✅ Done.

### Tests
- `tests/web/mock-server.js`: mirrors the predicate including the catch-all exclusion.

## Coverage

- **API integration (9)** — `TestListWebhooks_FilterByEventType_ExactMatch`, `_CaseInsensitive`, `_WhitespaceTrimmed`, `_EmptyIsNoOp`, `_NoMatch`, `_CatchAllExcluded`, `_ComposesWithTag`, `_ComposesWithSearch`, `_TotalCountReflectsFiltered`.
- **CLI (4)** — `TestCLI_WebhookList_ForwardsEventTypeFilter`, `_TrimsAndLowercasesEventType`, `_EmptyEventTypeOmitsParam`, `_EventTypeComposesWithSearch`.
- **Playwright (1)** — `event-type filter narrows the webhook list and round-trips through the URL` (also covers case-insensitive matching, catch-all exclusion, clear-button restore, URL round-trip, and empty-state copy).
- **JSDOM/mock-GUI (1)** — `webhook event-type filter narrows the list and excludes catch-alls` in `run-gui-tests.js`.
- **Mock server** — `?event_type=` predicate mirrored (case-insensitive exact-match against `event_types` with catch-all exclusion).

## Test plan

- [x] `CGO_ENABLED=1 go build ./...` clean (with `libvirt-dev` installed)
- [x] `CGO_ENABLED=1 go test -count=1 ./internal/api/ ./internal/cli/ ./pkg/types/` — all pass
- [x] `npm run generate:api` produces the new `event_type` query parameter
- [x] Mock server end-to-end via curl: `?event_type=vm.created` returns only explicit subscribers; `?event_type=VM.CREATED` matches case-insensitively; catch-all webhook is excluded; `X-Total-Count` reflects the post-filter population
- [ ] CI runs the new Playwright case against the mock server

---
_Generated by [Claude Code](https://claude.ai/code/session_012GcA8EiigXikwBrb6FFj9t)_


chinmay28 (commented):
LGTM (cannot self-approve via API — leaving COMMENT).

**Verified locally**
- `CGO_ENABLED=1 go build ./...` ✅
- `go test ./internal/api/ ./internal/cli/` ✅
- Targeted: `TestListWebhooks_FilterByEventType_*` (9 cases) and `TestCLI_WebhookList_*EventType*` (4 cases) all green.

**Architecture fit**
- "Explicit-membership / catch-all-excluded" semantics is the right choice and matches the existing `bulk_delete` `event_type` selector contract (documented in CLAUDE.md). The PR description and helper godoc both call out the operator intent crisply — *"which webhooks listen for vm.created"* vs *"which webhooks will fire for this event"*.
- Filter ordering (`tag → event_type → search → sort → pagination`) preserves the post-filter `X-Total-Count` contract.
- Frontend mirrors the rest of the 5.4.x family — debounced 250 ms input, URL round-trip via `?event_type=`, page reset on filter change, dedicated empty-state copy ("No webhooks explicitly subscribe to…").

**Minor: code duplication with `webhookExplicitlySubscribed`**

The PR adds `webhookSubscribedToEventType` that's almost identical to the existing `webhookExplicitlySubscribed` (line 586) — the only behavioral difference is case-sensitivity. The author explicitly chose to keep them separate to preserve the existing bulk_delete contract, which is reasonable. Long-term consider factoring into one helper with a `caseInsensitive bool` parameter, or — easier — making bulk_delete case-insensitive too (operator typo tolerance is probably the right behaviour everywhere). Not blocking for this PR.

**Note: needs rebase before merge**
Branch is based on an older main; current `origin/main` has #266 / #263. Diff vs current main shows what looks like a revert of those docs additions (ARCHITECTURE.md, PRODUCTION_DEPLOYMENT.md, Settings.jsx webhook search placeholder). Pure merge drift — please rebase or merge latest main before merging.

---

PR #271, opened 2026-05-20
feat(snapshots): ?since=/?until= time-range filter (5.4.28)

chinmay28:
## Summary

Extends `GET /api/v1/vms/{vmID}/snapshots` with `?since=<rfc3339>` and `?until=<rfc3339>` inclusive bounds on `created_at`, mirrored through `vmsmith snapshot list <vm-id> --since --until` and two new `datetime-local` inputs on the VMDetail snapshots card (URL round-trip via `?snap_since=` / `?snap_until=`).

Whitespace is trimmed before parsing; empty disables the bound; invalid values return 400 `invalid_since` / `invalid_until`. A snapshot with a zero `created_at` is filtered OUT whenever any bound is set — operators querying a time window don't want unbounded entries silently included. Composes additively with the existing `?tag=` / `?search=` filters so `X-Total-Count` reflects the post-filter population.

Closes the operator queries *"show me every snapshot taken before the bad deploy"* / *"audit the nightly snapshots from last week"* that the existing `?search=` filter cannot answer (timestamps don't appear in the search haystack) and that snapshot-name-prefix matching only addresses fuzzily (only if the naming convention happens to encode the timestamp).

### API
- `internal/api/parse_time_range.go`: shared `parseTimeRangeParam` helper (whitespace-trim + RFC3339 / RFC3339Nano, anything else returns the typed `*types.APIError` so the handler can pass it through `writeAPIError`) plus the `snapshotInTimeRange` predicate (inclusive bounds; zero-time filtered out when any bound is set).
- `internal/api/handlers_snapshot.go`: `ListSnapshots` parses both params up-front (400 fail-fast) and applies the predicate between `?tag=` and `?search=` so the post-filter `X-Total-Count` stays correct.

### CLI
- `internal/cli/parse_time_range.go`: mirrors the parser shape for the CLI (`parseCLITimeRange` returns a usage error that name-checks the offending flag) plus `snapshotInCLITimeRange`.
- `internal/cli/snapshot.go`: adds `--since` / `--until` to `snap list` with the same predicates inside the in-process filter loop.

### Frontend
- `web/src/pages/VMDetail.jsx`: adds two `datetime-local` inputs (`snap-list-since` / `snap-list-until`) under the snapshot search box with a "Clear range" button when any bound is set; URL round-trip via `?snap_since=` / `?snap_until=`; values are committed immediately (no debounce — date pickers already debounce naturally).
- `web/src/api/client.ts`: extends `snapshots.list({ since, until })` to forward the new query parameters against the regenerated OpenAPI schema.

### Docs
- `docs/openapi.yaml`: documents both `since` / `until` parameters with their RFC3339, inclusive, zero-time-excluded semantics.
- `CLAUDE.md`: REST quick reference for `GET /vms/{id}/snapshots` extended accordingly.
- `docs/ROADMAP.md`: 5.4.28 entry added and marked ✅ Done.

### Tests
- `tests/web/mock-server.js`: mirrors the predicate including the 400 paths and zero-`created_at` exclusion so Playwright can validate every branch. Seed snapshots upgraded to distinct timestamps so the GUI tests can exercise the time-range filter.

## Coverage

- **Unit (8 + 8)** — `TestParseTimeRangeParam` covers empty / whitespace / RFC3339 / RFC3339Nano / trimmed / invalid-text / invalid-date / invalid-epoch; `TestSnapshotInTimeRange` covers no-bounds / in-range / before / after / equal-since / equal-until / zero-with-bounds-excluded / zero-without-bounds-included.
- **API integration (8)** — `TestListSnapshots_FilterBySince`, `_FilterByUntil`, `_FilterBySinceAndUntil`, `_FilterBySince_Inclusive`, `_FilterByInvalidSince`, `_FilterByInvalidUntil`, `_FilterBySince_EmptyIsNoOp`, `_FilterByTimeRange_ExcludesZeroCreatedAt`, `_FilterBySince_ComposesWithTagAndSearch`.
- **CLI (6)** — `TestCLI_SnapshotList_FilterBySince`, `_FilterByUntil`, `_FilterBySinceAndUntil`, `_RejectsInvalidSince`, `_RejectsInvalidUntil`, `_EmptySinceOmitsFilter`.
- **Playwright (2)** — `snapshot ?since= filter narrows the list by created_at and round-trips through the URL`, `snapshot ?until= filter narrows the list by created_at upper bound`.

## Test plan

- [x] `CGO_ENABLED=1 go build ./...` clean
- [x] `CGO_ENABLED=1 go test -race -count=1 ./internal/api/ ./internal/cli/ ./pkg/types/` — all pass
- [x] `npm run generate:api` produces the new `since` / `until` query parameters on the snapshots route
- [ ] CI runs the new Playwright cases against the mock server

---

---
_Generated by [Claude Code](https://claude.ai/code/session_014GUNZwDCXzY67W8iwyXP7f)_


chinmay28 (commented):
LGTM (cannot self-approve via API — leaving as COMMENT).

**Verified locally:**
- `CGO_ENABLED=1 go build ./...` ✅
- `CGO_ENABLED=1 go vet ./...` ✅
- `CGO_ENABLED=1 go test -race -count=1 ./internal/api/ ./internal/cli/ ./pkg/types/` — all pass
- Targeted: `TestParseTimeRangeParam` (9), `TestSnapshotInTimeRange` (8), `TestListSnapshots_FilterBy(Since|Until|SinceAndUntil|*)` (8), `TestCLI_SnapshotList_(FilterBySince|FilterByUntil|FilterBySinceAndUntil|RejectsInvalidSince|RejectsInvalidUntil|EmptySinceOmitsFilter)` (6) — all green
- `npm run generate:api && npm run build` — frontend builds clean

**Architecture fit:**
- `parseTimeRangeParam` is colocated with the existing `parsePagination` / `parseTristateBoolParam` helpers in `internal/api/`; the CLI mirror `parseCLITimeRange` keeps the in-process and over-HTTP paths symmetric. Same factoring shape that 5.4.21 used.
- Filter ordering in `ListSnapshots` (`tag → since/until → search → sort → paginate`) preserves the post-filter `X-Total-Count` contract.
- 400 `invalid_since` / `invalid_until` mirrors the existing event-handler `invalid_since` error code (`internal/api/handlers_events.go:71`), so operators get a consistent vocabulary across surfaces.
- The "zero `created_at` filtered out when bounds set" semantic is the conservative choice — silently including unbounded entries in a time-range query would surprise operators.
- `datetime-local` UI inputs commit immediately (no debounce — date pickers already debounce naturally) and round-trip via `?snap_since=` / `?snap_until=`.

Would have approved if I weren't the PR author.

---

PR #275, opened 2026-05-21
docs(console): document proxy boundary and loopback note

march26bot:
## Summary
- document the browser-console ticket flow and proxy trust boundary
- add a console-specific security checklist to the architecture docs
- add an explicit loopback firewall note to the production deployment guide

## Testing
- python assertion script verifying the new doc anchors/strings

PR #278, opened 2026-05-21
docs(console): tighten security checklist wording

march26bot:
$## Summary\n- rename the console docs block to an explicit security checklist\n- call out loopback-only VNC and wss:// expectations more directly\n- clarify the roadmap checkoff text for the console docs slice\n\n## Testing\n- docs-only change\n


chinmay28 (approved):
Docs-only, merges cleanly into current `main` (verified). The reframing of `Security notes` → `Security checklist`, the explicit loopback bullet, and the `wss://` reminder are all reasonable, and the ROADMAP 5.1.12 checkoff wording is more concrete than the previous version.

**Heads-up (non-blocking):** since PR #275 landed, `docs/ARCHITECTURE.md` already grows a `Security checklist` block inside the new `Console Ticketing and Proxy Design` section near line 284. With this PR there will be two adjacent "Security checklist" headings in the doc — one in the top-half console-architecture section and one in §9.1 "Console Access Foundation" — that overlap on "VNC stays on loopback" and "tickets are short-lived / single-use". Not a merge conflict, but you might want a follow-up to either merge §9.1's Security checklist into the canonical one above or cross-link them so operators don't have to read both to get the full picture. The `wss://` bullet you're adding is genuinely new and worth keeping wherever the consolidation lands.

PR description also has a stray `$` and literal `\n` escapes at the top (`$## Summary\n- …`) — looks like a HEREDOC mis-render. Doesn't affect the diff, just makes the PR thread harder to skim.

---

PR #282, opened 2026-05-22
feat(events): server-side filters on the SSE event stream (5.4.33)

chinmay28:
## Summary

Extends `GET /api/v1/events/stream` with the same predicate set already shipped on `GET /events`: `?vm_id=`, `?type=`, `?type_prefix=`, `?source=`, `?severity=`, `?actor=`, `?resource_id=`, `?search=`. Filters apply to **both** replayed events (after the `Last-Event-ID` / `?since=<seq>` cursor) **and** live events arriving on the bus so the SSE stream and the paginated list endpoint agree on membership byte-for-byte.

Closes the operator queries *"tail just vm-1's events"* / *"stream every snapshot.* event family"* that previously required every consumer to discard cross-tenant noise client-side.

## Semantics

- Case-sensitive exact-match on `vm_id`, `type`, `source`, `severity`, `actor`, `resource_id` — mirrors the store-backed `EventFilter` byte-for-byte.
- Case-insensitive prefix on `type_prefix`; case-insensitive substring on `search` (delegated to the existing `EventMatchesSearch` haystack).
- Whitespace trimmed; empty disables the corresponding predicate. The pre-existing 410 `event_stream_replay_window_exceeded` short-circuit and the seq-based `?since=` replay cursor are unchanged — the filter set is purely additive.
- Non-matching events are dropped server-side before the SSE frame is written, so the wire cost scales with the post-filter population.

## Surface

- **API:** `internal/api/handlers_events.go` parses the filter once at the top of `StreamEvents` and gates both the replay loop and the live bus loop on the new predicate.
- **CLI:** `internal/cli/events.go` `streamEventsOnce` forwards every set field as a query param alongside the existing `since=` replay cursor; client-side `filterEvents` stays in place as defense-in-depth.
- **GUI:** `web/src/hooks/useEventStream.js` accepts an optional `filter` prop, stashes it behind a ref so reconnects always pick up the freshest value, and extends `buildStreamUrl` to append the supported keys. No call-site changes — Dashboard / VMList still get the unfiltered stream.
- **Shared predicate:** `pkg/types/event_filter.go` introduces `EventStreamFilter` + `EventMatchesStreamFilter` so the handler logic stays one source of truth.

## Test plan

- [x] `go test -count=1 -race ./internal/api/... ./internal/cli/... ./pkg/types/...`
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `gofmt -l` clean on new/modified files
- [x] `npm run build` (vite production build) clean
- [ ] golangci-lint (CI)
- [ ] Playwright mock-GUI tests (CI)

## Coverage added

- **Unit (`pkg/types`):** 8 `EventMatchesStreamFilter` cases — empty-matches-everything, nil-never-matches, VMID case-sensitive, per-field exact-match (Type/Source/Severity/Actor/ResourceID), TypePrefix case-insensitive, Search delegation, all-fields-must-match composition, and `HasAny` zero + per-field positive cases.
- **API integration (`internal/api`):** 4 cases — `TestStreamEvents_FilterReplay_VMID`, `_TypePrefix`, `_Search`, `TestStreamEvents_FilterLive_VMID` (the live case exercises the full bus pipeline + filter under publish/race conditions).
- **CLI (`internal/cli`):** 2 cases — `TestCLI_EventsFollow_ForwardsFiltersAsQueryParams` pins the wire contract for every supported field; `_EmptyFilterOmitsQueryParams` confirms backward compatibility for callers that don't filter.

## Docs

- `CLAUDE.md` REST quick reference adds a dedicated `GET /events/stream` row with the new filter set + reiterates the 410 / replay-cursor contract.
- `docs/ROADMAP.md` adds row 5.4.33 with ✅ Done.

https://claude.ai/code/session_01WLoWBxv21CtQwyTgbo8ieG

---
_Generated by [Claude Code](https://claude.ai/code/session_01WLoWBxv21CtQwyTgbo8ieG)_


chinmay28 (commented):
Independent re-verification (GitHub blocks self-approval, so leaving this as a COMMENT instead of an APPROVE — but everything checks out).

## What I verified

**Code shape**
- `pkg/types/event_filter.go` introduces `EventStreamFilter` + `EventMatchesStreamFilter` in `pkg/types`, which is the right home — both the API handler and any future webhook fan-out filter could share the same predicate. The `HasAny()` short-circuit on the no-filter fast path means the existing zero-cost SSE path is preserved for callers that don't filter.
- Field semantics match the store-backed `EventFilter` byte-for-byte (case-sensitive exact-match on `VMID`/`Type`/`Source`/`Severity`/`Actor`/`ResourceID`; case-insensitive prefix on `TypePrefix`; delegation to `EventMatchesSearch` on `Search`). The handler call site lowercases + trims `type_prefix` / `search` before populating the struct, which matches `ListEvents` exactly — so the SSE stream and the list endpoint really do agree on membership.
- The filter is applied **both** in the replay loop and the live bus loop. The replay-overflow 410 short-circuit is computed pre-filter, which is the correct call — a faraway client should still get a 410 fail-fast even if their filter would have dropped the noise.
- `useEventStream` filter handling: the `useRef` pattern is right; reconnects pick up the freshest filter without tearing down the EventSource on every prop change. Existing callers in `Dashboard.jsx` / `VMList.jsx` pass no `filter` and are unaffected (the `if (filter)` guard in `buildStreamUrl` handles undefined cleanly).
- CLI `streamEventsOnce` forwards every set field as a query param; the client-side `filterEvents` predicate stays in place as defense-in-depth, which is documented in the source.

**Build + tests**
- `CGO_ENABLED=1 go test -race -count=1 ./pkg/types/... ./internal/api/... ./internal/cli/... ./internal/store/... ./internal/logger/... ./internal/network/... ./internal/config/... ./internal/webhooks/...` — all green (8.5s + 11.4s + 9.7s + others)
- `go vet ./...` — clean
- `gofmt -l pkg/types/event_filter.go pkg/types/event_filter_test.go internal/api/handlers_events.go internal/cli/events.go` — clean on the new/modified files
- `cd web && npm run build` — clean Vite production build

**Coverage**
- 8 `EventMatchesStreamFilter` unit cases cover empty / nil / VMID case-sensitivity / per-field exact / TypePrefix / Search delegation / all-fields-must-match composition / `HasAny` zero+per-field.
- 4 SSE API integration cases (`TestStreamEvents_FilterReplay_VMID` / `_TypePrefix` / `_Search`, plus `TestStreamEvents_FilterLive_VMID` exercising the bus pipeline under publish/race). The live test is structured the right way — interleaving noise + signal on a 50ms tick and asserting that vm-drop never crosses the wire.
- 2 CLI cases pin the wire contract (`_ForwardsFiltersAsQueryParams` exhaustively checks every supported key; `_EmptyFilterOmitsQueryParams` preserves pre-5.4.33 wire behavior).

## Architecture / roadmap fit

Matches roadmap item 5.4.33 and follows the same pattern as 4.2.23 / 4.2.24 / 4.2.25 (actor / resource_id / type_prefix filters on the list endpoint). The CLAUDE.md REST quick reference adds a dedicated `GET /events/stream` row documenting the new predicate set and reiterating the 410 / replay-cursor contract — that's the right shape to keep the AI-readable surface honest.

## Non-blocking notes

1. No `useEventStream` consumer in the GUI passes a `filter` yet. That's fine — this PR is the primitive; future Activity-page work can wire `?vm_id=` / `?type_prefix=` into the live stream so the dashboard scoping carries through to SSE. Worth tracking as a follow-up roadmap row.
2. The `useEventStream` filter ref means changing the `filter` prop at runtime won't tear down + re-open the EventSource — the new filter only takes effect on the next reconnect (which the consumer would need to force). The PR docstring is explicit about this. For Dashboard / VMList future use, a callback to force-reconnect-on-filter-change would close the gap, but it's not blocking since no consumer needs that today.

LGTM — ready for a second reviewer's APPROVE.

---

chinmay28 (commented):
Re-verification (GitHub blocks self-approval — leaving this as COMMENT, but everything checks out).

**Code shape**
- `pkg/types/event_filter.go` introduces `EventStreamFilter` + `EventMatchesStreamFilter` in the right home (shared types). The `HasAny()` short-circuit preserves the zero-cost SSE path for unfiltered callers.
- Field-level parsing in `StreamEvents` matches `ListEvents` byte-for-byte: `TrimSpace` on every field, `ToLower` on `type_prefix` + `search` to mirror the caller-pre-lowers contract that `EventMatchesSearch` documents. So the SSE stream and `/events` agree on membership.
- Filter is applied in both the replay loop and the live bus loop. The replay-overflow 410 short-circuit is computed pre-filter — correct.
- `useEventStream` filter handling: `useRef` is the right shape — reconnects pick up the freshest filter without tearing down the `EventSource` on every prop change. Existing call sites (Dashboard / VMList) don't pass `filter`, so they're unaffected via the `if (filter)` guard in `buildStreamUrl`.
- CLI `streamEventsOnce` forwards every set field as a query param while keeping the client-side `filterEvents` predicate as defense-in-depth.

**Build + tests (local)**
- `CGO_ENABLED=1 go test -race -count=1 ./pkg/types/... ./internal/api/... ./internal/cli/...` — all green
- `go vet ./...` — clean
- `gofmt -l` clean on every new/modified Go file

**Coverage**
- 8 `EventMatchesStreamFilter` unit cases (empty / nil / VMID case-sensitivity / per-field exact / TypePrefix case-insensitive / Search delegation / all-fields-AND / `HasAny`).
- 4 SSE API integration cases — `_FilterReplay_VMID`, `_TypePrefix`, `_Search`, plus `_FilterLive_VMID` which exercises the bus pipeline under publish/race.
- 2 CLI cases pinning the wire contract (`_ForwardsFiltersAsQueryParams` exhaustively, `_EmptyFilterOmitsQueryParams` to preserve pre-5.4.33 behaviour).

**Architecture / roadmap fit**
- Roadmap row 5.4.33 — matches the pattern of 4.2.23 / 4.2.24 / 4.2.25 (actor / resource_id / type_prefix on the list endpoint) and closes the per-stream parity gap. CLAUDE.md REST quick reference updated.

**Non-blocking follow-ups**
1. No `useEventStream` consumer in the GUI passes a `filter` yet — fine, this PR is the primitive; an Activity-page follow-up can wire `?vm_id=` / `?type_prefix=` into the live stream.
2. The `useEventStream` filter ref means changing `filter` at runtime won't reconnect — the docstring is explicit, so the contract is clear, but a follow-up callback to force-reconnect-on-filter-change would close the gap if a Dashboard scope-toggle ever needs it.

LGTM — ready for a second reviewer's APPROVE.

---

chinmay28 (commented):
(Posting as a comment review since GitHub disallows self-approval.) **LGTM** — would approve otherwise.

Clean extension of `GET /events/stream` with the same predicate set that `GET /events` already exposes. Verified the moving pieces:

**Semantics match the existing store filter byte-for-byte.** `pkg/types/event_filter.go::EventMatchesStreamFilter` does case-sensitive exact-match on `VMID` / `Type` / `Source` / `Severity` / `Actor` / `ResourceID`, case-insensitive prefix on `TypePrefix` (caller-lowercased), and delegates `Search` to the shared `EventMatchesSearch` haystack. This mirrors `internal/store/bolt.go::ListEventsFiltered` exactly, so a client picking between `/events` and `/events/stream` sees identical membership. `HasAny()` short-circuits the no-filter fast path cleanly.

**Handler integration is correct.** `StreamEvents` parses the filter once at the top, applies it inside both the replay loop (after the 410 overflow check — pre-filter is the right place for that count) and the live bus loop. Non-matching events are dropped before the SSE frame is written, so the wire cost scales with post-filter population as advertised.

**CLI and GUI surfaces are consistent.** The CLI `streamEventsOnce` forwards every filter field as a query param while keeping the in-process predicate as defense-in-depth. The React hook captures `filter` behind a ref so reconnects pick up the freshest value without tearing down the EventSource on every prop change — nice touch.

**Tests.** Verified locally — `go test -count=1 -race ./...` passes after installing libvirt-dev; the 8 unit + 4 SSE integration + 2 CLI cases in this PR all hit. CI also green.

**Minor follow-up (non-blocking):** `docs/openapi.yaml` doesn't yet describe `GET /events/stream` (only the `/vms/{vmID}/stats/stream` SSE endpoint is there). This isn't a regression — the endpoint was already undocumented before this PR — but adding the new filters widens that gap. Worth picking up in a follow-up so Swagger UI surfaces them.

---

PR #291, opened 2026-05-24
docs(console): flesh out proxy/security design

march26bot:
## Summary
- expand the architecture docs for the console ticket and websocket proxy flow
- document VNC vs serial console intent resolution and active-session lifecycle expectations
- add an explicit console security checklist and mark roadmap item 5.1.12 done

## Testing
- docs-only change; verified via direct diff inspection


chinmay28 (approved):
## Review: approved ✅

Docs-only change; merges cleanly against `main` (0 conflicts). I verified the new content against the actual codebase rather than just reading it.

### Accuracy
- The numbered ticket flow matches what's implemented today: step 1 (`POST /api/v1/vms/{id}/console/ticket` behind normal API auth, verify-running, single-use ticket) is exactly the `handlers_console.go` endpoint already in `main`.
- The `intent=vnc` (`<graphics type='vnc'>`) vs `intent=serial` (`<console type='pty'>`) split is consistent with roadmap 5.1.9, and the section is correctly framed as the *designed* shape for "remaining console work," so describing the not-yet-built serial path here is appropriate.
- The 5.1.12 roadmap note references `docs/PRODUCTION_DEPLOYMENT.md` keeping VNC loopback-only — confirmed that note is genuinely present in `main` (`PRODUCTION_DEPLOYMENT.md:150-152`), so the "Done" claim isn't overstated.

### Quality
The rewrite is a clear improvement: the third invariant (browser never reaches libvirt sockets directly), the VM-ID + caller-identity scoping rule, the "reject upgrade when VM not running / endpoint unresolvable" item, and the `wss://` guidance are all the right things to call out for an authenticated console proxy. No security concerns — it argues *for* keeping raw VNC off the network and redacting `ticket=`.

Nothing to run (no code). Looks good.

---

PR #293, opened 2026-05-24
feat(scheduler): recurring VM-action schedules (5.2)

chinmay28:
## Summary

Implements **Scheduled Operations** (roadmap §5.2) — recurring VM actions (snapshot / start / stop / restart) on a cron timer — as a full vertical slice on top of the existing schedule type/store foundation (5.2.1 / 5.2.2, already ✅). Closes roadmap items **5.2.3, 5.2.4, 5.2.5, 5.2.6, 5.2.7, 5.2.8, 5.2.9, 5.2.12** ✅ (5.2.10 VMDetail embed and the real-QEMU portion of 5.2.11 are deferred with honest notes in the roadmap).

### Engine (`internal/scheduler/`)
- One `robfig/cron/v3` instance **per distinct timezone** (`WithSeconds` + `WithLocation`); schedules register/deregister on CRUD.
- Bounded **worker pool** (overflow → `queue_full` skip) + **per-schedule concurrency** semaphore (overlap → `concurrent_run` skip).
- Action registry: `snapshot` (auto-named `auto-<schedule>-<ts>`, honors `retention_count`, trims only its own auto snapshots), `start`/`stop`/`restart` with typed skips (`vm_not_found`, `vm_already_running`, `vm_already_stopped`).
- Per-attempt **timeout** + **retry/backoff** for transient errors (skips never retried).
- **Startup catch-up** (`skip` / `run_once` / `run_all`, capped at `max_catch_up`); `last_tick` cursor advanced every `tick_interval_seconds`. Runs in the background so a long replay never blocks daemon startup.
- **Tag-selector fan-out** resolved against the live VM set at fire time; one run per matched VM. Empty target = all VMs.
- Emits `schedule.created/updated/deleted/fired/fire_succeeded/fire_failed/fire_skipped/catch_up_replayed` with `actor: "scheduler"`.

### API (`internal/api/handlers_schedule.go`)
`POST/GET/GET{id}/PATCH/DELETE /api/v1/schedules`, `GET {id}/runs`, `POST {id}/run-now`. Decoupled from the engine via `ScheduleController` / `ScheduleStore` interfaces (mirrors the webhook subsystem so `api` never imports `scheduler`). Cron/timezone/target/catch-up validation; list `vm_id`/`action`/`enabled`/`search` filters + `sort`/`order` + pagination with `X-Total-Count`. Returns 503 `schedules_disabled` when the subsystem is off.

### CLI (`internal/cli/schedule.go`)
`vmsmith schedule create | list | show | edit | delete | run-now`.

### GUI (`web/src/pages/Schedules.jsx`)
List with inline enabled toggle, next-fire, last-result chip, run-now; create/edit modal with cron preset chips (hourly / daily 02:00 / weekly Sun); per-row recent-runs expander; URL-round-tripped search/action/enabled/sort filters. New nav item + route + `client.ts` `schedules` group.

### Wiring / config / docs
- Daemon wires the engine (start after the bus, stop before it); `schedules.*` config block + `vmsmith.yaml.example`.
- CI runs the new `./internal/scheduler/...` + `./pkg/types/...` unit tests.
- New `docs/SCHEDULES.md`; CLAUDE.md, `docs/openapi.yaml`, and ROADMAP updated.

## Test plan
- [x] Unit (`internal/scheduler`): action dispatch + skip reasons, snapshot retention trim, retry-then-success / retry-exhausted, concurrent_run + queue_full skip, tag fan-out, all-VMs, register/unregister, catch-up policies, missed-fire math, ValidateSpec/NextFireTime.
- [x] API integration (`internal/api`): create success + 8 validation cases, list filters/sort/400s, get/update/delete 404 + noop, run-now + run history, 503-when-disabled.
- [x] CLI (`internal/cli`): create/list/edit/delete/run-now + client-side validation.
- [x] Playwright mock-GUI (`tests/web/gui.spec.js`): 9 Schedules scenarios — runs in CI (Chromium download blocked in the authoring environment).
- [x] `gofmt`, `go vet ./...`, CI-pinned `golangci-lint v1.64.8`, full `go test -race` suite, and `make web` all green locally.

https://claude.ai/code/session_01DgZu2LM86BUPvNemGaWYFn

---

---
_Generated by [Claude Code](https://claude.ai/code/session_01DgZu2LM86BUPvNemGaWYFn)_


internal/scheduler/engine.go
```diff
@@ -0,0 +1,685 @@
...
+// skip instead of blocking the cron goroutine.
+func (e *Engine) enqueue(j job) {
+	e.mu.Lock()
+	ch := e.jobs
+	e.mu.Unlock()
+	if ch == nil {
+		return
+	}
+	select {
+	case ch <- j:
```
chinmay28: **Shutdown race: `send on closed channel` panic.**

`enqueue` reads `ch := e.jobs` under the lock, *releases* the lock, then does this non-blocking send. Meanwhile `Stop()` (lines 290-309) acquires the lock, sets `e.jobs = nil`, releases, then `close(jobs)`. If a cron FuncJob goroutine is parked between the unlock at line 341 and this send, `Stop()` can close the very channel `ch` points at → `panic: send on closed channel` (and a data race).

This is reachable because `Stop()` only calls `c.Stop()` on each cron, which **does not wait for in-flight job goroutines**. I confirmed against `robfig/cron@v3.0.1`: `startJob` runs each entry in `go func()` tracked by `jobWaiter`, and `Cron.Stop()` returns a `context.Context` that's only `Done()` after `jobWaiter.Wait()` — but the engine discards that return value. So a cron tick that fires concurrently with daemon shutdown can be inside `enqueue` when the channel closes.

The race detector won't flag it in the current tests because none stop the engine at the instant a cron FuncJob is mid-enqueue, but on a busy daemon with second-granularity schedules the window is real.

Two clean fixes:

1. **Send under the lock.** Since `Stop()` also takes `e.mu` before nil-ing/closing, doing the non-blocking select-send while holding `e.mu` makes them mutually exclusive — `enqueue` either wins before the close or observes `e.jobs == nil`. Keep the (blocking) I/O of `recordQueueFull` *outside* the lock:
   ```go
   e.mu.Lock()
   ch := e.jobs
   full := false
   if ch != nil {
       select {
       case ch <- j:
       default:
           full = true
       }
   }
   e.mu.Unlock()
   if ch != nil && full {
       e.recordQueueFull(j)
   }
   ```
2. **Or** wait for the cron job goroutines before closing: capture `ctxs := append(ctxs, c.Stop())` for each cron in `Stop()`, then `for _, c := range ctxs { <-c.Done() }` before `close(jobs)`.

Either resolves the panic; option 1 is the smaller change.

---

chinmay28 (commented):
## Review: strong work, one shutdown race to fix before merge

(Posting as a comment — GitHub blocks self-approval on your own PR. Verdict below.)

I reviewed the engine, actions, catch-up, API decoupling, daemon wiring, config, and types, and ran `go test -race ./internal/scheduler/... ./internal/api/... ./internal/cli/... ./pkg/types/... ./internal/config/... ./internal/daemon/...` locally — **all green**, and it merges cleanly against `main` (0 conflicts). CI is also green.

This is a genuinely well-architected vertical slice for roadmap 5.2:
- **Decoupling is right.** `api` defines its own `ScheduleStore` / `ScheduleController` interfaces (mirroring the webhook `WebhookRegistrar` pattern), so `api` never imports `scheduler` — consistent with the established architecture.
- **Daemon ordering is correct:** engine starts *after* the event bus (so fire events are delivered) and stops *before* it (so in-flight fires can still publish). The `cfg.Schedules.Enabled` gate wires the REST surface either way and returns 503 `schedules_disabled` when off.
- **Concurrency model is thoughtful:** one cron per timezone, bounded worker pool with `queue_full` skip instead of blocking the cron goroutine, per-schedule semaphore for `concurrent_run`, per-attempt timeout + backoff, and skips that are never retried.
- **Snapshot retention is safe:** `trimSnapshots` only ever touches `auto-<name>-` prefixed snapshots, so operator-created snapshots can't be deleted; oldest-first sort tiebreaks on the timestamped name. 
- **Catch-up math is defensive:** `missedFires` guards zero/future `lastTick` and non-advancing schedules, and the cursor is captured *before* it's advanced. Running catch-up in the background so it never blocks daemon startup is the right call.

### Must-fix (1)
- **Shutdown race in `enqueue` vs `Stop()`** — see the inline comment on `engine.go:346`. A cron FuncJob mid-`enqueue` can race `close(jobs)` → `panic: send on closed channel`. Confirmed `c.Stop()` doesn't wait for in-flight job goroutines (its returned context is discarded). Concrete fix in the inline comment (send under `e.mu`).

### Minor / non-blocking
- **`RunNow` is synchronous.** `RunNowSchedule` → `engine.RunNow` → `e.fire(...)` runs inline on the request goroutine, so a `run-now` against a slow action (snapshot) or a large tag fan-out can block the HTTP request up to `ActionTimeout` per VM and hold the connection past typical proxy timeouts. Fine for v1, but consider firing async + returning 202, or documenting the blocking behavior.
- **Worker fires use `context.Background()`** (`worker()` at engine.go:332), so live cron fires ignore the daemon ctx and are bounded only by `ActionTimeout` + the `Stop()` drain. Catch-up correctly uses the daemon ctx. Not wrong, but means a shutdown waits for buffered jobs to drain rather than cancelling them — worth a note in `docs/SCHEDULES.md`.
- **`startAction` treats `paused` as `vm_already_running`** and skips. Arguably a paused VM should `resume`, but skipping is safe and defensible for v1.

Verdict: **approve once the `engine.go:346` race is addressed.** Everything else is solid and well-tested.

---

internal/scheduler/engine.go
chinmay28: Fixed in e488ff8 via option 1 (send under `e.mu`). `enqueue` now does the non-blocking select-send while holding `e.mu` — which `Stop()` also holds before niling/closing `e.jobs` — so the two are mutually exclusive: enqueue either wins before the close or observes `e.jobs == nil`. `recordQueueFull`'s blocking store/event I/O stays outside the lock.

Added `TestEnqueue_ConcurrentWithStop_NoPanic`, which hammers `enqueue` from 8 goroutines concurrently with `Stop()`. Verified it panics/data-races under `-race` on the previous implementation and passes reliably now (`-count=5 -race`).

---

PR #295, opened 2026-05-24
feat(templates): ?default_user= filter on GET /templates (5.4.38)

chinmay28:
## Summary

Adds a per-`default_user` filter to the template list across the full vertical slice (API, CLI, GUI), closing roadmap item **5.4.38** ✅.

- `GET /api/v1/templates?default_user=<value>` — a case-insensitive exact-match against the template's `default_user` field. Closes the operator query *"show me every template that provisions the `deploy` SSH user"*. Templates carry a `default_user` (2.4.1) that VMs already filter by (5.4.23), but the template list could only slice by `tag` / `image` / `search`. Applied after `?image=` and before `?since=`/`?until=`/`?search=` so it composes additively and `X-Total-Count` reflects the post-filter population.
- **No empty-means-root fallback** (intentional divergence from the VM filter): a template's empty `default_user` means "use the image's built-in user", so an empty stored value never matches a non-empty query.
- CLI: `vmsmith template list --default-user <value>`.
- GUI: a 250 ms-debounced "Filter by default user" input on the TemplateList page that round-trips through `?default_user=` in the URL, with a tailored empty-state.

Whitespace-trimmed; empty disables the filter.

## Files

- `internal/api/handlers_template.go` — `?default_user=` predicate wired into `ListTemplates`
- `internal/cli/template.go` — `--default-user` flag on `template list`
- `web/src/pages/TemplateList.jsx`, `web/src/api/client.ts` — debounced GUI input + client param
- `docs/openapi.yaml` (+ regenerated `web/src/api/generated/schema.d.ts`), `CLAUDE.md`, `docs/ROADMAP.md`
- `tests/web/mock-server.js` — mirrors the filter; seeded templates carry distinct users (`ubuntu` / `root`)

## Test plan

- [x] API integration: 8 `TestListTemplates_FilterByDefaultUser_*` cases (exact / case-insensitive / trims-whitespace / empty-no-op / empty-stored-never-matches / no-match / composes-with-tag / total-count)
- [x] CLI: 4 `TestCLI_TemplateList_FilterByDefaultUser_*` cases (exact / case-insensitive / empty-omits / no-match)
- [x] Playwright: 3 scenarios (narrow + URL round-trip, case-insensitive, no-match)
- [x] `go build ./...`, `go vet ./...`, race-enabled api+cli+types tests green locally; `make web` builds
- [ ] Playwright mock GUI suite — runs in CI (Chromium download is blocked by this environment's network policy)

https://claude.ai/code/session_01LmjFuvzT1ZfazJsQEk4h8o

---
_Generated by [Claude Code](https://claude.ai/code/session_01LmjFuvzT1ZfazJsQEk4h8o)_


chinmay28 (commented):
✅ **Reviewed — looks good to merge.** (Posting as a comment rather than a formal approval since GitHub won't let me approve my own PR.)

Verified end-to-end — clean, well-scoped implementation of roadmap 5.4.38.

- **Backend** (`handlers_template.go`): the `?default_user=` predicate is applied after `?image=` and before the time-range/`?search=` filters, exactly as documented, using the same in-place `templates[:0]` filter pattern as its siblings. Case-insensitive via `EqualFold`; an empty stored `default_user` never matches a non-empty query (the documented divergence from the VM filter's empty-means-root fallback — correct, since a template's empty value means "use the image's built-in user").
- **CLI** (`template.go`): `--default-user` mirrors the API semantics.
- **Vertical slice consistency**: OpenAPI param, `mock-server.js`, and `client.ts` (`defaultUser → default_user`) all line up with the backend.
- **Tests**: ran `go test -race ./internal/api/ ./internal/cli/` — all 8 `TestListTemplates_FilterByDefaultUser_*` and 4 `TestCLI_TemplateList_FilterByDefaultUser_*` cases pass, and the full api+cli packages are green with no regressions. `go build ./...` clean.

Note: the Playwright GUI suite couldn't run locally (Chromium download blocked by this environment, as the PR notes) — verified the frontend builds and the mock-server filter logic is coherent instead.

---

PR #298, opened 2026-05-25
feat(schedules): show VM schedules on VM detail (5.2.10)

march26bot:
## Summary
- add a schedules card to VM detail showing direct VM schedules and matching tag-selector schedules
- add a VM-scoped create-schedule modal with the target VM prefilled/read-only
- mark roadmap item 5.2.10 done and cover the flow with Playwright

## Testing
- `npx playwright test tests/web/gui.spec.js --grep "vm detail shows matching schedules and can prefill a new VM schedule"`


chinmay28 (approved):
Clean implementation of roadmap 5.2.10. Verified:

- **Roadmap alignment** — the card shows direct (`vm_id == this.id`) and tag-selector-matched schedules, and "Add schedule" opens a VM-scoped modal with the target VM prefilled/read-only. Matches the 5.2.10 spec; the roadmap line flip to ✅ Done is accurate.
- **Code correctness** — all referenced identifiers are in scope (`Modal`/`Spinner`/`ErrorBanner` from `components/Shared`, `useEffect`, and `tags = safeArray(vm.tags)` defined before the filter). The `CreateVMScheduleModal` payload matches `CreateScheduleRequest` (name/action/vm_id/cron_spec/timezone/enabled/catch_up_policy/retention_count/max_concurrent), and the default cron `0 0 2 * * *` is a valid 6-field spec.
- **Build** — `npm run build` (vite) compiles cleanly.
- **Test wiring** — the new Playwright spec is coherent with the mock server: `sch-1` is seeded as `nightly-snapshot` on `vm-1`, and `POST /schedules` returns `sch-new-N` carrying the request's `vm_id`, so the created row renders on the same VM detail page.

⚠️ I could **not execute** the Playwright test here — Chromium download is blocked by this environment's network policy (`npx playwright install chromium` is a silent no-op, no browser cached), same constraint your other PRs note. The static/build verification above is what I could confirm; please make sure the GUI suite goes green in CI.

Two minor, non-blocking observations for a follow-up if you think they're worth it:
1. **All-VMs schedules aren't shown.** A schedule with no `vm_id` and empty `tag_selector` targets *every* VM (per the schedule semantics) and will fire against this VM, but the filter returns `false` when `selector.length === 0`, so it never appears on the card. The roadmap text only calls for vm_id/tag matches, so this is in-scope — just a slight gap in "what automation affects this VM."
2. **Client-side filtering over `perPage: 100`.** If a deployment grows past 100 schedules, matches beyond the first page won't render here. Fine for v1; a `vm_id`-scoped fetch would be the eventual fix.

LGTM.

---

PR #312, opened 2026-05-28
fix(webhooks): persist final HTTP failure status

march26bot:
## Summary
- persist the receiver's final HTTP status on webhook delivery failures instead of zeroing it out
- do the same for manual webhook test deliveries so Settings/CLI reflect the real HTTP failure
- update webhook manager tests to lock in the 500-status persistence contract

## Testing
- Not run locally: this environment does not have the Go toolchain installed
- Verified by direct code inspection against the existing webhook tests

PR #324, opened 2026-05-29
feat(schedules): ?timezone= filter on GET /schedules (5.4.55)

chinmay28:
## Summary

Closes roadmap item **5.4.55**. Extends `GET /api/v1/schedules` with `?timezone=<IANA>`, a case-sensitive exact-match against the schedule's stored `timezone` field. Closes the deployment-audit operator query *"show me every schedule that runs in `America/New_York`"* / *"which schedules were ever pinned to the legacy `Asia/Tokyo` zone"* — the schedule list already sliced by `?action=` / `?catch_up_policy=` / `?enabled=` (categorical) and `?vm_id=` / `?tag_selector=` (target) but had no symmetric way to scope by timezone, even though timezone determines fire-time semantics across DST transitions and across geographically-distributed operators.

IANA timezone names are case-sensitive (`America/New_York` ≠ `america/new_york`), so the filter is case-sensitive — mirrors the `?vm_id=` / `?actor=` / `?resource_id=` exact-match contracts; case-insensitive matching is the job of `?search=`. Whitespace-trimmed; empty disables; no default-fallback for empty stored values since the engine's effective default is host-dependent (`time.Local`) — operators querying by timezone must supply the literal value they stored. Applied between `?catch_up_policy=` and `?enabled=`; `X-Total-Count` reflects the post-filter population.

## Changes

- **API** — `internal/api/handlers_schedule.go` `ListSchedules` reads + trims `?timezone=` and applies the exact-match predicate inside the filter loop right after the catch-up check.
- **CLI** — `internal/cli/schedule.go` adds `vmsmith schedule list --timezone <tz>` with whitespace-trimmed forwarding (no lowercasing — IANA case-sensitivity).
- **GUI** — `web/src/pages/Schedules.jsx` adds a 250 ms-debounced "Timezone" text input alongside the existing catch-up dropdown, URL round-trip via `?timezone=`, page-reset on change. `web/src/api/client.ts` extends `schedules.list({ timezone })`.
- **Mock server** — `tests/web/mock-server.js` mirrors the exact-match filter and updates the seed `sch-2` timezone from `""` to `America/New_York` so the filter has a meaningful 2/1 spread (UTC × 2 vs America/New_York × 1).
- **Docs** — `docs/openapi.yaml` documents the new query parameter (regenerated `schema.d.ts`). `CLAUDE.md` REST quick reference + CLI map extended. `docs/ROADMAP.md` marks 5.4.55 done; `README.md` progress block regenerated to 201/221 (91.0%).

## Coverage

- **1 API integration case** — `TestListSchedules_FilterByTimezone` (baseline / exact match UTC + America/New_York + Asia/Tokyo / case-sensitive `utc` and `america/new_york` rejection / whitespace-trim / empty no-op / no-match / composes with action / composes with catch_up_policy / paginated total-count).
- **2 CLI cases** — `TestCLI_ScheduleList_ForwardsTimezone`, `TestCLI_ScheduleList_EmptyTimezoneOmitsParam`.
- **1 Playwright case** — `timezone filter narrows the schedule list and round-trips through the URL` (covers exact match, case-sensitivity miss, multi-row UTC match, and clear-restores-unfiltered).

## Drive-by fix

This PR also fixes a stale assertion in `TestTestWebhook_EndToEndFailure` (`internal/api/handlers_webhook_test.go`). The test was added before #312 changed `recordExternalDeliveryResult` to persist the receiver's actual HTTP status (instead of zeroing it on failure). The test still asserted `LastStatus = 0`, which now reliably fails against the post-#312 behavior. Updated to assert `LastStatus = 500`. Without this fix the integration suite was failing on `main` independently of this feature.

## Test plan

- [x] `go vet ./internal/api/ ./internal/cli/` clean
- [x] Race-enabled targeted tests green locally (`go test -race -run TestListSchedules_FilterByTimezone ./internal/api/`, `go test -race -run TestCLI_ScheduleList ./internal/cli/`)
- [x] CI's unit-test package set + integration test set (`./internal/api/`) green locally with `-race`
- [x] Mock-server sanity-checked via curl (UTC → sch-1/sch-3, America/New_York → sch-2, lowercase variant → empty, empty → all three)

https://claude.ai/code/session_01PTKhbvtQQBwSVLW8vzZ7uG

---
_Generated by [Claude Code](https://claude.ai/code/session_01PTKhbvtQQBwSVLW8vzZ7uG)_


chinmay28 (commented):
Reviewed end-to-end and verified locally — would approve, but GitHub blocks self-approval.

**Architecture alignment**
- The case-sensitive exact-match contract mirrors the existing `?vm_id=` / `?actor=` / `?resource_id=` predicates exactly, and the explicit decision *not* to apply a `?catch_up_policy=skip`-style empty-means-default fallback is the right call given the engine's `time.Local` host dependency.
- Filter is applied between `?catch_up_policy=` and `?enabled=`, matching the documented ordering in `CLAUDE.md` and `docs/openapi.yaml`. `X-Total-Count` stays post-filter correct.
- Roadmap item 5.4.55 ✅ Done; README progress block regenerated to 201/221; OpenAPI + generated TS schema updated.

**Verification**
- `go test -race -run TestListSchedules_FilterByTimezone ./internal/api/` — PASS
- `go test -race -run TestTestWebhook_EndToEndFailure ./internal/api/` — PASS (drive-by fix is correct: post-#312 `recordExternalDeliveryResult` now persists the receiver's actual status code, so asserting `LastStatus = 500` instead of `0` is the right contract)
- `go test -race ./internal/api/ ./internal/cli/ ./internal/webhooks/ ./internal/scheduler/ ./internal/store/ ./pkg/types/` — all green
- `go vet` clean on the modified files; `gofmt -l` clean on the PR's diff
- Mock-server filter mirrors the daemon: trim → exact-match against `s.timezone || ""`

**Drive-by fix justification**
`TestTestWebhook_EndToEndFailure` was reliably failing on `main` since #312 changed the persisted-status semantics. Fixing it in the same PR keeps CI green and is the minimum scope to land 5.4.55.

LGTM — ready to merge once CI is green.

---

chinmay28 (commented):
## Review (LGTM)

Clean implementation of the timezone filter on `GET /schedules`. Implementation follows the established schedule-list filter pattern exactly.

**Architecture alignment**
- Case-sensitive exact-match is the right contract — IANA timezone names are case-sensitive, and the PR description correctly cites the `?vm_id=` / `?actor=` / `?resource_id=` precedent. Case-insensitive matching belongs in `?search=`, which is unchanged.
- No empty-string fallback (unlike `?default_user=root` → linux runtime semantics) is the right call: the engine's empty timezone resolves to `time.Local`, which is host-dependent, so there's no stable value to back-fill against.
- Filter placed between `?catch_up_policy=` and `?enabled=`, composing additively with all existing filters; `X-Total-Count` stays post-filter correct.
- Full vertical slice covered (API handler, CLI flag, GUI debounced input, mock-server, OpenAPI schema regenerated).
- Drive-by webhook test fix is the same one in #327 / #329 — justified: `TestTestWebhook_EndToEndFailure` is broken on `main` after #312 (verified locally).

**Tests**
- Full Go test suite (`./internal/api/`, `./internal/cli/`, `./pkg/types/`) green locally on this branch.
- API integration test covers: baseline, exact-match across three timezones, case-sensitivity (lowercase variants must miss), whitespace-trim, empty-no-op, no-match, compose-with-action, compose-with-catch_up_policy, paginated total-count.
- CLI tests cover forward-trim and empty-omit semantics.
- Mock-server seed `sch-2` migration to `America/New_York` gives Playwright a meaningful 2/1 spread.

**Roadmap**
- 5.4.55 marked Done; README progress regenerated to 201/221.

No changes requested; ready to merge.

---

PR #327, opened 2026-05-30
feat(vm): first-class Windows guest support (workstation + server)

chinmay28:
Add an os_type guest-family selector to VMSpec so vmsmith can run Windows
10/11 and Windows Server 2019/2022/2025 alongside Linux, sharing the same
overlay-disk / NAT / DHCP-reservation / snapshot / lifecycle machinery. Only
the libvirt domain tuning and the first-boot provisioning datasource diverge.

Types & validation:
- VMSpec gains OSType (linux|windows), OSVariant, and write-only AdminPassword,
  plus ResolvedOSType()/IsWindows() helpers and KnownWindowsVariants.
- validateVMSpec adds invalid_os_type / invalid_os_variant (400) and a Windows
  resource floor (>=2048 MB RAM, >=32 GB disk, only when explicitly set).

Domain XML (OS-aware, Linux path unchanged):
- Windows guests get a SATA system disk (boots without virtio storage drivers),
  e1000e NIC (native driver), localtime clock, Hyper-V enlightenments +
  hypervclock timer, USB tablet, and QXL video. Empty device fields normalise
  to the historical Linux defaults so direct DomainParams callers/tests are
  unaffected.
- The virtio-win driver ISO is attached as an extra cdrom when
  storage.virtio_win_iso is set (auto-probes /usr/share/virtio-win/...).

Provisioning:
- createProvisioningISO dispatches on OS family; Windows gets a cloudbase-init
  NoCloud datasource (cidata ISO) with meta-data (instance-id/hostname/
  admin_pass) and a #ps1_sysnative PowerShell user-data that sets the
  Administrator password, enables RDP, and installs/enables OpenSSH with the
  injected key. AdminPassword is redacted from the stored/returned record.

CLI: vmsmith vm create --os --os-variant --admin-password.

Docs: docs/WINDOWS_GUESTS.md operator guide, README quick-start example,
openapi.yaml VMSpec schema, vmsmith.yaml.example, ROADMAP 5.6, CLAUDE.md.

Coverage: pkg/types OS-type helpers, domain Windows tuning + XML, Windows
user-data generation + PS quoting, API create/validation paths, CLI create.

https://claude.ai/code/session_01WLAnMffitMeqWCBjyhxPAf


chinmay28 (commented):
## Review (LGTM with one nit)

Large but well-contained Windows-guest feature. The os-family selector keeps the Linux code path unchanged while adding the Windows-specific tuning at the few points where it actually diverges (domain XML, provisioning datasource, NIC model). The structure is right.

**Architecture alignment**
- `OSType` + `ResolvedOSType()` + `IsWindows()` helpers in `pkg/types` cleanly preserve the "empty == linux" backwards-compat contract for pre-existing VMs.
- `DomainParams` empty-field normalisation in `GenerateDomainXML` keeps every direct caller (including tests) on the historical Linux defaults — good defensive design.
- Disk-bus split (SATA system disk + e1000e NIC) so Windows boots without virtio drivers preloaded; virtio-win ISO attaches when configured and is silently dropped otherwise. That's the right operator ergonomics.
- `AdminPassword` redaction (`spec.AdminPassword = ""` after `createProvisioningISO`) keeps the secret out of bbolt and the API response. Good — but worth a follow-up test that asserts `vm.Spec.AdminPassword == ""` post-create at the bbolt persistence boundary, not just at the in-memory layer.
- Cloudbase-init `#ps1_sysnative` user-data is idempotent and degrades gracefully (RDP enabled unconditionally; SSH only when a key is provided). `psSingleQuote` correctly doubles single quotes for PS string literals — tested.
- Drive-by webhook test fix is the same one in #324 / #329 (justified — broken on main after #312).

**Nit (non-blocking)**
- `validateOSType` does a case-sensitive comparison against literal `"linux"` / `"windows"`. The CLI normalises via `strings.ToLower`, but a raw JSON `POST` with `"os_type": "Windows"` (initial cap) will currently return 400 `invalid_os_type`. Either lowercase the value before the switch (mirrors the CLI) or document the case-sensitivity in `openapi.yaml`. Same applies to `IsKnownWindowsVariant` — the CLI lowercases the variant flag, but a JSON request with `"os_variant": "Windows-Server-2022"` will hit `invalid_os_variant`. Low-priority since the docs / examples consistently show lowercase, but worth a one-liner `strings.ToLower` in the API path to keep CLI and HTTP behaviour symmetric.

**Tests**
- Full Go test suite (`./internal/api/`, `./internal/cli/`, `./pkg/types/`, `./internal/vm/`) green locally on this branch.
- Coverage is thorough: OS-type helpers, Windows domain tuning (disk bus / clock / hyperv / tablet / qxl / e1000e), domain XML rendering with and without virtio-win, Linux defaults unchanged, validation (invalid os_type, unknown variant, RAM/disk floors, zero == default OK), Windows user-data (PS header, password set, quote escaping, SSH path, no-SSH path), CLI create flow.
- No mock-manager changes needed since the Windows fields flow through the existing `Create`/`Spec` paths.

**Roadmap & docs**
- `docs/WINDOWS_GUESTS.md` is the right home for the operator-facing setup story; README quick-start, openapi.yaml schema, install-deps scripts, and `vmsmith.yaml.example` all wired in.
- ROADMAP 5.6 marked, CLAUDE.md updated.

Ready to merge once the case-sensitivity nit above is either fixed or explicitly punted to a follow-up.

---

PR #328, opened 2026-05-30
Add template creation to the web GUI

chinmay28:
The Templates admin page could list, edit (description/tags), and delete
templates, but there was no way to create one from the GUI — the empty-state
even pointed users to the Create-VM modal, which only *applies* templates.

Add a "New Template" button + CreateTemplateModal to TemplateList that
collects name, base image (dropdown of imported images, free-text fallback
when none exist), cpus/ram_mb/disk_gb, default_user, description, and tags,
then POSTs to the already-existing /templates endpoint via the existing
templatesApi.create client method.

Also wire POST /api/v1/templates into the Playwright mock server (with
name/image-required and duplicate-name validation mirroring the daemon) and
add GUI tests covering the create flow and the submit-disabled guard.

https://claude.ai/code/session_014AUo6m3f9Ny64Xhb55TAaY


chinmay28 (commented):
## Review (LGTM)

Closes a real GUI gap — the Templates page could list / edit / delete but pointed users to the Create-VM modal for *creating* a template (which only *applies* templates), which was confusing.

**Architecture alignment**
- Reuses `templatesApi.create` from the existing client.ts surface — no new client wiring needed.
- `CreateTemplateModal` is self-contained: form-reset-on-open via the `[open]` effect, image dropdown when images are seeded with free-text fallback when not (good operator UX for fresh installs), submit gated on `name` + `image` non-empty, errors surfaced inline from the mutation result, optional fields (`default_user`, `description`, `tags`) only included in the POST body when non-empty.
- Mock-server addition mirrors the daemon contract correctly: `invalid_name` when missing, `invalid_image` when missing, duplicate-name check (case-insensitive). Returns 201 on success.
- The TemplateList empty-state now includes a "New Template" CTA (with its own `data-testid`) so the page recovers cleanly from a true-empty state.

**Tests**
- CI green (backend + frontend mock GUI). Frontend-only change — Go test impact is nil.
- Playwright covers both the happy-path create flow (name / image dropdown / cpus / ram / disk / default-user / description / tags) and the submit-disabled guard (name alone is not enough; image required).

**Nit (non-blocking)**
- `humanSize` uses base-1024 multipliers but labels them "GB" / "MB" instead of "GiB" / "MiB". This matches what the rest of the GUI already does (ImageList uses the same idiom), so consistent — but worth a one-line audit pass at some point. Not for this PR.

No changes requested; ready to merge.

---

PR #334, opened 2026-05-31
feat(vm): auto-generate Windows Administrator password (5.6.17)

chinmay28:
## Summary

Closes roadmap **5.6.17** — the natural follow-up to #327's Windows-guest landing and #333's `admin_password` write-only / redacted contract.

When a Windows VM is created without an explicit `admin_password`, the daemon now mints a 20-char password from `crypto/rand` that always satisfies the Windows complexity policy (one upper / lower / digit / symbol, Fisher-Yates shuffled). The password is injected into the cloudbase-init datasource, redacted from the stored bbolt record (same as today), and surfaced **exactly once** on the create response as `generated_admin_password` on the `VM` struct. `GET /vms/{id}` and `GET /vms` never carry the field — it is set on the response copy *after* `store.PutVM` so it cannot land in bbolt.

## Surfaces

- **API** (`POST /api/v1/vms`): one-shot `generated_admin_password` in the response body. Documented as create-response-only in the OpenAPI schema.
- **CLI** (`vmsmith vm create`): banner with the password and a "save it now" warning printed immediately after the VM ID. Suppressed when `admin_password` is supplied explicitly and for Linux guests.
- **GUI** (`VMList.jsx`): one-time-reveal modal (`GeneratedAdminPasswordModal`) with a copy-to-clipboard button after a successful create; the value is gone once dismissed.
- **Mock server** mirrors the auto-generate + transient-on-response shape so Playwright assertions see a realistic value while subsequent GETs on the stored record return nothing.

## Why this matters

`admin_password` has been write-only since 5.6.3 (#327): the operator could pass one on create, but if they didn't, the VM booted with a blank/random Administrator password that they could never recover. This closes that UX gap — operators get either the password they chose or one the daemon mints for them, with the same redaction guarantees in both paths. Rotation path documented in `docs/WINDOWS_GUESTS.md` — recreate the VM with an explicit `admin_password` (or omit it again to mint a fresh one); in-place reset is out of scope (would require re-running cloudbase-init on a live guest).

## Changes

- **types** — `pkg/types/vm.go`: `VM.GeneratedAdminPassword` field (`omitempty`-tagged, set only on the create response).
- **VM manager** — new `internal/vm/admin_password.go` (crypto/rand 20-char generator with the four Windows complexity classes from a shell-safe alphabet, Fisher-Yates shuffled). `LibvirtManager.Create` and `MockManager.Create` both call it when `spec.IsWindows() && spec.AdminPassword == ""`. The LibvirtManager sets `vm.GeneratedAdminPassword` *after* `store.PutVM` so the password never reaches bbolt; the MockManager mirrors the same shape (stored record carries no admin_password, response copy carries the generated password).
- **CLI** — `internal/cli/vm.go` prints a one-time banner with the password and a "save it now" warning when `result.GeneratedAdminPassword != ""`.
- **GUI** — `web/src/pages/VMList.jsx`: captures the create response and opens `GeneratedAdminPasswordModal` when the response carries the field. The modal renders the password in a `select-all` `<code>` block, offers a copy-to-clipboard button (with "Copied" feedback for 2s), and a "I've saved it" dismiss that clears the state.
- **Mock server** — `tests/web/mock-server.js`: `mockGenerateAdminPassword()` 16-char generator (same alphabet as the daemon) + the `wantsAutoPassword` check on `POST /vms`. The response is built from a clone of the stored record so subsequent GETs return nothing.
- **OpenAPI** — `docs/openapi.yaml` documents `VM.generated_admin_password` as create-response-only. Regenerated `web/src/api/generated/schema.d.ts`.
- **Docs** — `CLAUDE.md` REST quick-reference extended on `POST /vms`. `docs/WINDOWS_GUESTS.md` gains a dedicated "Auto-generated Administrator password" section covering the three surfaces and the recreate-to-rotate path. `docs/ROADMAP.md` marks 5.6.17 ✅ Done. `README.md` progress regenerated via `scripts/roadmap-progress.sh` to 211 / 240 (87.9%).

## Coverage

- **5 unit cases** — `internal/vm/admin_password_test.go`:
  - `TestGenerateAdminPassword_Length` (exact 20 chars).
  - `TestGenerateAdminPassword_SatisfiesComplexity` (50 iters × 4 class assertions each).
  - `TestGenerateAdminPassword_Uniqueness` (100 iters, dedupe via map).
  - `TestGenerateAdminPasswordN_RejectsTooShort` (length < 4 returns error).
  - `TestGenerateAdminPasswordN_MinLengthOK` (length 4: one from each class, no filler).
- **3 API integration cases** — `internal/api/api_test.go`:
  - `TestCreateVM_GeneratesAdminPasswordForWindowsWhenOmitted` — POST returns the password AND a follow-up GET returns neither `generated_admin_password` nor `spec.admin_password`.
  - `TestCreateVM_DoesNotGenerateAdminPasswordWhenSupplied` — explicit `admin_password` suppresses the field.
  - `TestCreateVM_DoesNotGenerateAdminPasswordForLinux` — Linux VM never carries the field.
- **3 CLI cases** — `internal/cli/commands_test.go`:
  - `TestCLI_VMCreate_WindowsGeneratesAdminPassword` — banner + "save it now" present.
  - `TestCLI_VMCreate_LinuxDoesNotShowPasswordBanner` — Linux create has no banner.
  - `TestCLI_VMCreate_WindowsWithExplicitPasswordDoesNotShowBanner` — explicit pw suppresses banner.
- **1 Playwright case** — `tests/web/gui.spec.js`: `creating a Windows VM without admin_password surfaces a one-time reveal` — selects `tmpl-3` (windows-2022), asserts modal visibility, password length ≥ 12, copy button toggles to "Copied", post-dismiss modal disappears + the new VM card appears.

## Test plan

- [x] `go vet ./...` clean
- [x] `gofmt -l` clean on the diff
- [x] Backend unit + API integration tests green locally with `-race -count=1` (CI's exact test set)
- [x] Mock-server smoke test via curl: `POST {os_type:windows}` returns `generated_admin_password`; subsequent `GET /vms/{id}` does NOT; Linux POST does NOT; explicit `admin_password` does NOT
- [x] Frontend `make web` builds cleanly (regenerated `schema.d.ts` typechecks against the new field)

---
_Generated by [Claude Code](https://claude.ai/code/session_01EdaphTmMHHp2yK6CRPAgrd)_


chinmay28 (commented):
LGTM — would approve if GitHub allowed self-approval. Reviewed end-to-end, ran tests, verified the architectural claims.

**Architecture / correctness**
- Redaction sequence in `internal/vm/lifecycle.go::Create` is provably safe — the password is set on `spec.AdminPassword`, baked into the provisioning ISO (line 222), redacted via `spec.AdminPassword = ""` (line 228) **before** `store.PutVM` (line 286), and only then mounted on the response copy as `vm.GeneratedAdminPassword` (line 297). The bbolt write at PutVM serializes the redacted spec; the in-memory mutation afterward never touches the store. Verified by `TestCreateVM_GeneratesAdminPasswordForWindowsWhenOmitted` which asserts the follow-up `GET /vms/{id}` returns neither `generated_admin_password` nor `spec.admin_password`.
- `startIPMonitor` → `applyStaticIPFallback` does not regenerate the provisioning ISO on the DHCP-fallback path; it only adds a DHCP host reservation and restarts the domain. The password baked into the original ISO survives that restart, so cloudbase-init still applies it on boot. No correctness regression there.
- `MockManager.Create` mirrors the libvirt manager's contract: `storedSpec.AdminPassword = ""` before `m.vms[id] = vm`, then the response copy gets the transient `GeneratedAdminPassword`. The API integration tests exercise this end-to-end against `httptest`.
- `VMManager` interface unchanged (no new method needed since generation is an internal create-time concern). Matches the `CLAUDE.md` "always go through the Manager interface" guidance.
- OpenAPI documents the field as create-response-only and the regenerated `schema.d.ts` carries the same description verbatim, so the GUI typecheck pins it.

**Crypto**
- `crypto/rand` throughout (no `math/rand` slip). `rand.Int(rand.Reader, big.NewInt(n))` for each draw eliminates modulo bias.
- 20-char default × union alphabet of ~78 chars ≈ 126 bits of entropy, well above the description's stated ~118-bit floor.
- Symbol set wisely excludes shell-mangling chars (`$`, backtick, `\`, `"`, `'`, space) — important since the CLI banner prints the value to a terminal.
- Length floor of 4 in `generateAdminPasswordN` is the right minimum to guarantee one char from each complexity class; `TestGenerateAdminPasswordN_RejectsTooShort` / `_MinLengthOK` pin both ends of the contract.
- Fisher-Yates shuffle correctly uses `randomIndex(i+1)` so position 0 can also receive a non-class-seed character.

**Roadmap fit**
- Natural follow-up to 5.6.18 (#333) which landed `admin_password` as write-only + redacted. 5.6.17 closes the "no password supplied" UX gap that ticket explicitly called out. Roadmap row marked ✅ Done with a detailed Notes column; `README.md` progress regenerated to 211 / 240 via `scripts/roadmap-progress.sh`.
- Rotation-by-recreate is the only sane path given the value is by-design non-persisted, and `docs/WINDOWS_GUESTS.md` calls that out explicitly along with the in-guest `net user Administrator` escape hatch.

**Testing (locally on the PR head)**
- `go test -race -count=1 ./internal/vm/ ./internal/api/ ./internal/cli/`: green.
- All 11 new tests pass with `-v -race`:
  - `TestGenerateAdminPassword_{Length,SatisfiesComplexity,Uniqueness}` plus the two N-variants
  - `TestCreateVM_{GeneratesAdminPasswordForWindowsWhenOmitted,DoesNotGenerateAdminPasswordWhenSupplied,DoesNotGenerateAdminPasswordForLinux}`
  - `TestCLI_VMCreate_{WindowsGeneratesAdminPassword,LinuxDoesNotShowPasswordBanner,WindowsWithExplicitPasswordDoesNotShowBanner}`
- `gofmt -l` clean across the touched files; `go vet ./...` clean.
- CI on the PR head SHA is green (both "Backend build and tests (1.22.5)" and "Frontend build and mock GUI tests" succeeded).
- The pre-existing race in `TestJanitorRemovesExpiredTicketsAutomatically` (`internal/console`) reproduces on `main` and is unrelated to this PR.

Nice piece of work — ready to merge.

---

PR #335, opened 2026-05-31
feat(web): surface Windows guest details in VM UI

march26bot:
## Summary
- add guest-OS badges to the VM list and VM detail pages
- extend the Create Machine modal with Windows OS/variant/password controls and client-side RAM/disk floors
- show an RDP connection hint on Windows VM detail pages when a 3389 port-forward exists

## Testing
- Not run: frontend build unavailable in this worktree because local web deps are not installed
- Added Playwright coverage for the new Windows UI affordances

## Roadmap
Implements roadmap item 5.6.6 (GUI surface for Windows guests).

PR #336, opened 2026-05-31
feat(vm): per-VM device overrides (5.6.15)

chinmay28:
## Summary

Closes roadmap **5.6.15** — a natural follow-up to the Windows-guest landing (#327) and the `os_type` immutability work (#333). Five optional override fields on `VMSpec` (`disk_bus`, `nic_model`, `machine`, `firmware`, `virtio_win_iso`) let advanced operators tune individual VMs — especially Windows guests once virtio drivers are installed in-guest — without changing the daemon-wide config. Each override falls back to the OS-family default when omitted, so existing VMs and unset specs render identically to before.

### Why this matters

`storage.virtio_win_iso` and the SATA/e1000e+pc-q35-6.2 defaults baked at create time were a fleet-wide knob. After installing the virtio-net/virtio-blk drivers from the attached ISO, an operator had no way to flip a single Windows guest over to virtio for throughput, or to pin a specific QEMU machine version, or to enable UEFI for a Windows 11 install — short of editing the daemon config + restarting it. This change makes each of those a per-VM field that respects the OS-family default when empty.

## Surfaces

- **API** (`POST /api/v1/vms`): five new optional properties on `VMSpec`. Validation rejects unknown values with stable 400 codes (`invalid_disk_bus`, `invalid_nic_model`, `invalid_machine`, `invalid_firmware`); `virtio_win_iso` is a free-form path (only NUL bytes rejected — the LibvirtManager logs a warning + falls back if missing on the host).
- **CLI** (`vmsmith vm create`): `--disk-bus`, `--nic-model`, `--machine`, `--firmware`, `--virtio-win-iso`, all lowercased + whitespace-trimmed before being forwarded.
- **GUI** (`VMList.jsx`): new "Device Tuning" subsection in the Advanced tab of the Create-VM modal — three enum selects (disk_bus / nic_model / firmware) plus machine + virtio_win_iso text inputs. Empty keys are stripped from the request body so the daemon resolves the OS-family default.

## Behaviour

| Field | Empty resolves to | Override values | Notes |
| --- | --- | --- | --- |
| `disk_bus` | linux=virtio, windows=sata | `virtio` / `sata` | Flips the disk target letter (`vda`/`sda`) and bumps the cloud-init cdrom slot (`sda`→`sdb` on SATA, `sdb`→`sda` on virtio) so the libvirt XML stays consistent. |
| `nic_model` | linux=virtio, windows=e1000e | `virtio` / `e1000e` | Applied to **every** `<interface>` entry — primary NAT + extras. Pairs with the `applyVirtioWin` Windows ISO so operators can install the drivers, then pin virtio. |
| `machine` | `pc-q35-6.2` | conservative `[A-Za-z0-9._-]+` | Shell metacharacters rejected before they can reach the domain template. |
| `firmware` | bios (no attribute) | `bios` / `uefi` / `ovmf` | `uefi` and `ovmf` both map to libvirt's `firmware='efi'` shorthand, which auto-picks the host's OVMF code/vars pair. Required for Windows 11. Does NOT enable Secure Boot or attach a virtual TPM — that's roadmap 5.6.9. |
| `virtio_win_iso` | `storage.virtio_win_iso` config / `/usr/share/virtio-win/virtio-win.iso` auto-probe | absolute path | Per-VM override. Missing override path logs a warning and falls back to the daemon config rather than failing the create. |

## Changes

- **types** — `pkg/types/vm.go`: 5 new fields + `ResolvedDiskBus()` / `ResolvedNICModel()` / `ResolvedFirmwareAttr()` / `IsValidMachineType()` helpers. Five new constants (`DiskBusVirtio`, `DiskBusSATA`, `NICModelVirtio`, `NICModelE1000e`, `FirmwareBIOS`, `FirmwareUEFI`, `FirmwareOVMF`).
- **domain** — `internal/vm/domain.go`: `DomainParams` gains `NICModel` + `FirmwareAttr`; `DomainParamsFromSpec` consumes the overrides; the XML template gains a conditional `<os firmware='...'>` attribute.
- **lifecycle** — `internal/vm/lifecycle.go`: new `virtioWinISOPathForSpec(spec)` checks the per-VM override before falling back to `virtioWinISOPath()`. `applyVirtioWin` switches to the spec-aware resolver.
- **API** — `internal/api/validation.go`: `validateDeviceOverrides` is called from `validateVMSpec` after the existing OS-type + clock-offset validation. Four new error codes wired into `statusForAPIError` so they return HTTP 400.
- **CLI** — `internal/cli/vm.go`: 5 new flags on `vmsmith vm create`. Values lowercased + trimmed (matching the `--os`/`--clock-offset` convention).
- **OpenAPI** — `docs/openapi.yaml`: 5 new properties under `VMSpec`. Regenerated `web/src/api/generated/schema.d.ts`.
- **GUI** — `web/src/pages/VMList.jsx`: "Device Tuning" subsection in the Advanced tab. `advancedCount` updated so the tab badge counts the new fields too.
- **Mock server** — `tests/web/mock-server.js`: mirrors the validation contract (400 paths) and persists the five fields on the stored record so Playwright assertions can re-fetch.
- **Docs** — `CLAUDE.md` REST quick reference extended on `POST /vms`. `docs/ROADMAP.md` marks 5.6.15 ✅ Done with the full implementation summary. `README.md` progress regenerated via `scripts/roadmap-progress.sh` to 212 / 240 (88.3%).

## Coverage

- **16 unit cases** — `pkg/types/vm_device_overrides_test.go`: `TestVMSpecResolvedDiskBus` × 6 (linux/windows defaults, mixed-case, whitespace, both-direction overrides), `TestVMSpecResolvedNICModel` × 5, `TestVMSpecResolvedFirmwareAttr` × 6 (empty→empty, bios→empty, uefi/ovmf→efi, mixed-case, whitespace), `TestIsValidMachineType` × 12 (shell metacharacter / quote / path-separator rejection alongside accepted alphabet).
- **10 domain-layer cases** — `internal/vm/domain_overrides_test.go`: `TestDomainParamsFromSpec_DeviceOverrides` × 9 (linux defaults, linux→sata, windows→virtio post-driver, uefi shorthand, ovmf alias, explicit bios attribute-less, custom machine, NIC override, mixed-case lowercase) plus `TestDomainParamsFromSpec_NICModelOverrideAppliedToEveryInterface` (regression: every `<interface>` carries the override, not just the first).
- **5 API integration cases** — `internal/api/api_test.go`: `TestCreateVM_DeviceOverrides_Accepted`, `_EmptyResolvesToDefaults`, `TestCreateVM_InvalidDiskBus`, `_InvalidNICModel`, `_InvalidFirmware`, `_InvalidMachine`.
- **1 CLI case** — `internal/cli/commands_test.go`: `TestCLI_VMCreate_DeviceOverrides` covers the five flags round-tripping through the mock manager with lowercasing.
- **1 Playwright case** — `tests/web/gui.spec.js`: `device tuning overrides round-trip through the Advanced tab` — fills all five inputs on a new VM, then re-fetches via `/api/v1/vms` to confirm every field persisted on `spec`.

## Test plan

- [x] `go vet ./...` clean
- [x] `gofmt -l` clean on the diff
- [x] Backend unit + API integration tests green locally
- [x] `make web` builds cleanly (regenerated `schema.d.ts` typechecks against the new fields)
- [ ] GitHub Actions CI passes (backend + frontend mock GUI)

---
_Generated by [Claude Code](https://claude.ai/code/session_013qxjXQ2KzpAe6yUaMrmLA6)_


chinmay28 (commented):
Solid overall — the type / validation / CLI / GUI plumbing all hang together and the unit tests for `DomainParamsFromSpec` and the resolver helpers are thorough. But I found one **real bug** that needs to be fixed before this lands. (Same author as the PR, so leaving this as a COMMENT — please treat as "needs changes".)

### 🐛 Bug: `--machine` override silently overwritten by `detectMachineType` in every lifecycle entry point

`DomainParamsFromSpec` correctly resolves `spec.Machine → params.Machine` (verified by `TestDomainParamsFromSpec_DeviceOverrides/custom machine type`), but **all three call sites in `internal/vm/lifecycle.go` then unconditionally overwrite it on the very next line**:

- `internal/vm/lifecycle.go:231-232` (Create):
  ```go
  params := DomainParamsFromSpec(spec, diskPath, cloudInitISO, m.cfg.Network.Name, natMAC)
  params.Machine = detectMachineType(m.conn)   // <-- nukes spec.Machine
  ```
- `internal/vm/lifecycle.go:374-375` (Clone) — same pattern.
- `internal/vm/lifecycle.go:900-902` (Update — domain redefine) — same pattern.

So the per-VM `--machine pc-q35-rhel9.6.0` an operator passes through the CLI / GUI is parsed, validated, stored in bbolt, asserted in `TestCreateVM_DeviceOverrides_Accepted`, and then **silently ignored** at the point where it actually has to land in the libvirt XML. The other four overrides (`disk_bus`, `nic_model`, `firmware`, `virtio_win_iso`) are not affected because nothing post-`DomainParamsFromSpec` touches them in the lifecycle layer.

**Why the existing tests miss this:**
- `domain_overrides_test.go` calls `DomainParamsFromSpec` directly, not via `LibvirtManager.Create` / `Clone` / `Update`.
- The API tests (`TestCreateVM_DeviceOverrides_Accepted`) use `testServer(t)` which is wired to `MockManager`, not `LibvirtManager`, so they only round-trip what's stored in bbolt — they never reach the codepath that calls `detectMachineType`.

**Suggested fix** (one option — minimal):
```go
params := DomainParamsFromSpec(spec, diskPath, cloudInitISO, m.cfg.Network.Name, natMAC)
if strings.TrimSpace(spec.Machine) == "" {
    params.Machine = detectMachineType(m.conn)
}
```
Apply to all three call sites. Bonus: extend the regression test in `internal/vm` to exercise a `LibvirtManager` codepath where `detectMachineType` is stubbed (or refactor so the auto-detect happens inside `DomainParamsFromSpec` via an injected callback, so the unit tests cover the whole path).

### Nits (non-blocking)

1. **`pkg/types/vm.go` `regexpMustCompile` wrapper** — the file-level docstring justifies the wrapper as "keep the regex literal next to the consumer without polluting init()", but a top-level `var machineTypeRe = regexp.MustCompile(...)` already does that without the wrapper. Recommend dropping `regexpMustCompile` and calling `regexp.MustCompile` directly — one less indirection.
2. **VMSpec docstring is misleading on PATCH semantics** — `pkg/types/vm.go:139` says of `DiskBus`: "Baked at create time; see VMUpdateSpec for the mutability contract." But `VMUpdateSpec` doesn't actually mention these fields — sending them on a PATCH body is silently ignored (they JSON-unmarshal into nothing). Unlike `os_type` / `os_variant`, there's no 400 rejection. Either add `*string` rejection sentinels to `VMUpdateSpec` (matching the `os_type_immutable` contract) or update the docstring to say "ignored on PATCH; resend on a new create to change."

### Tests run locally

- `gofmt`: clean
- `go vet ./...`: clean
- `go test -count=1 ./pkg/types/... ./internal/vm/... ./internal/api/... ./internal/cli/...`: all green (including the 16 new resolver tests, 10 domain-layer tests, 5 API integration tests, 1 CLI test)

### Alignment

- Architecture: ✅ five new optional fields on `VMSpec` are stored verbatim, validated at the API boundary, resolved through pure helpers — same pattern as the existing `os_type` / `clock_offset` work in #333.
- Roadmap: ✅ closes 5.6.15 as a natural follow-up to #327 (Windows guests) + #333 (OS-type immutability + clock offset).
- The bug above means the roadmap claim "Per-VM device overrides ✅ Done" is incomplete for the `machine` axis specifically — the other four overrides work end-to-end.

Once `params.Machine` is fixed (plus a regression test that doesn't go through the mock manager), this is ready to merge.

---

chinmay28 (commented):
LGTM after the `resolveMachine` fix — closes roadmap 5.6.15. (Can't self-approve; treating this as approval.)

### Re-verifying the fix from `9e40b98`

The earlier review on `ba3e476` flagged that `params.Machine` was being unconditionally overwritten by `detectMachineType(m.conn)` at all three lifecycle entry points (Create / Clone / Update), silently dropping the operator's `--machine` override. The follow-up commit routes all three sites through a shared `resolveMachine(specMachine, fallback)` helper that only falls back when the spec is blank/whitespace-only:

```go
// internal/vm/lifecycle.go:231-232, :374-375, :900-902
params := DomainParamsFromSpec(...)
params.Machine = resolveMachine(spec.Machine, func() string { return detectMachineType(m.conn) })
```

The new `TestResolveMachine` (`internal/vm/domain_overrides_test.go`) pins this with four cases (empty / whitespace-only / verbatim override / trimmed override), and crucially asserts `called` so future regressions can't silently re-overwrite the override. The two nits (drop the `regexpMustCompile` wrapper, fix the `DiskBus` docstring) are also addressed.

### Verified
- **CI**: both checks green on `9e40b98`.
- **Tests locally** (after `apt install libvirt-dev`): `./pkg/types/...`, `./internal/vm/...`, `./internal/api/...`, `./internal/cli/...` all green — includes `TestResolveMachine`, `TestCreateVM_DeviceOverrides_*`, `TestDomainParamsFromSpec_DeviceOverrides`, `TestDomainParamsFromSpec_NICModelOverrideAppliedToEveryInterface`, `TestVMSpecResolved{DiskBus,NICModel,FirmwareAttr}`, `TestIsValidMachineType`, `TestCLI_VMCreate_DeviceOverrides`.
- **Vet + gofmt**: clean.
- **Merges cleanly** into current `main`.

### Architectural alignment
- Five new optional fields on `VMSpec` stored verbatim, validated at the API boundary via `validateDeviceOverrides`, resolved through pure helpers (`ResolvedDiskBus` / `ResolvedNICModel` / `ResolvedFirmwareAttr` / `IsValidMachineType`) — same pattern as the existing `os_type` / `clock_offset` work.
- OS-family defaults preserved when overrides are empty (Linux→virtio/no-firmware-attr; Windows→sata/e1000e); disk-target letter (`vda`/`sda`) and cloud-init cdrom slot flip together so the libvirt XML stays consistent under disk-bus override.
- NIC model applied to **every** `<interface>` (primary NAT + extras), not just the first — covered by the dedicated regression test.
- Per-VM `virtio_win_iso` override falls back through the daemon config and then the `/usr/share/virtio-win/virtio-win.iso` auto-probe; missing override path warns + falls back rather than failing the create, matching the existing config-level behaviour.
- CLI flags lowercased + whitespace-trimmed before forwarding (matches `--os`/`--clock-offset` convention).
- OpenAPI + `schema.d.ts` + GUI "Device Tuning" subsection in the Advanced tab + mock-server validation parity all consistent.

Ship it. (Note: PR #335 also touches `CreateVMModal`'s `emptyForm` — whichever of the two lands second will need a small rebase to merge the two field sets.)

---

PR #348, opened 2026-06-02
feat(schedules): duration range filter on run history (5.4.64)

chinmay28:
## Summary

Extends `GET /api/v1/schedules/{id}/runs` with `?min_duration_ms=<n>` / `?max_duration_ms=<n>` — inclusive non-negative integer bounds on each run's `finished_at - started_at` duration in milliseconds. The symmetric range counterpart to the `duration` sort axis added in 5.4.63: the same axis can now be **sorted** *and* **filtered** to bound the visible cohort, the way `?since=`/`?until=` complement the `started_at` sort axis.

Runs with a nil `finished_at` (still-running runs have no known duration) are filtered OUT whenever either bound is set, mirroring the `finished_since`/`finished_until` nil-handling so the filter family stays consistent.

Closes the SRE triage operator queries:
- *"show me every run that took ≥ 5 minutes during yesterday's window"*
- *"every run that completed in under a second (probably a no-op)"*

…that the categorical `?status=` filter cannot answer (a 5s success and a 2h success share `status=success`) and the time-range filters cannot answer either (they bound *when* a run ran, not *how long*).

## Vertical slice

| Layer | Change |
|---|---|
| **API** | `internal/api/handlers_schedule.go::ListScheduleRuns` parses both bounds via the existing `parseCountRangeParam` helper (400 `invalid_min_duration_ms` / `invalid_max_duration_ms` on garbage or negative) and applies the predicate right after the `finished_at` time-range check via `countInRange`. Duration computed as `int(run.FinishedAt.Sub(run.StartedAt) / time.Millisecond)`, clamped to ≥0 so out-of-order timestamps don't leak negative durations. |
| **CLI** | `vmsmith schedule runs <id> --min-duration-ms --max-duration-ms` (Int flags, client-side rejection of negatives, using `cmd.Flags().Changed(...)` so an explicit `--min-duration-ms=0` is forwarded while the unset default is not). |
| **GUI** | Two `<input type="number" min="0" step="1">` "Min duration ms / Max duration ms" controls in the Schedules-page recent-runs expander alongside the existing finished-at inputs, plus a "Clear duration" button. Client-side `parsePositiveInt` mirrors the API contract. |
| **OpenAPI** | `docs/openapi.yaml` documents both params on `GET /schedules/{scheduleID}/runs`; regenerated `web/src/api/generated/schema.d.ts` and `web/src/api/client.ts`. |
| **Mock-server** | `tests/web/mock-server.js` mirrors the integer parse (400 path + filter + nil-finished_at exclusion). |
| **Docs** | `CLAUDE.md` REST quick-reference + CLI map extended; `docs/ROADMAP.md` entry 5.4.64 added with ✅. |

## Test plan

- [x] **API integration**: `TestListScheduleRuns_FilterByDuration` — baseline / min lower bound / max upper bound / both-bounds window / inclusive boundary / wide-open excludes still-running / empty-no-op / whitespace-trim / compose with status / compose with `since` / invalid 400 paths incl. negative / paginated total-count
- [x] **CLI**: `TestCLI_ScheduleRuns_ForwardsDurationRange`, `_DurationRangeZeroExplicit` (explicit `--min-duration-ms=0` IS forwarded), `_DurationRangeUnsetOmitsParams` (unset flag does NOT send the param), `_RejectsNegativeDuration`
- [x] **Playwright** (mock-server): `duration range filter narrows recent runs by finished_at - started_at duration` exercises min-lower-bound / both-bounds collapse-to-single / clear / max-upper-bound
- [x] Local: `go test -race ./internal/api/ ./internal/cli/ ./pkg/types/` passes; `make build-go` passes; `npm run build` passes; mock-server end-to-end sanity check confirms `min=5000` → run-4+run-2, `5000..5000` → run-2, `max=4000` → run-1+run-3, negative/garbage → 400.

🔗 Roadmap: 5.4.64 ✅

---
_Generated by [Claude Code](https://claude.ai/code/session_0116CrFKZhPahv3PvHrQQFzH)_


chinmay28 (commented):
LGTM — clean vertical slice for roadmap 5.4.64. (Cannot formally approve since I'm the author; leaving a review comment instead.)

**What I checked:**
- API handler in `internal/api/handlers_schedule.go` parses both bounds with the existing `parseCountRangeParam` helper, applies `countInRange` right after the `finished_at` time-range check and before `?search=`, and clamps negative durations to 0 to handle out-of-order timestamps. Returns the documented `invalid_min_duration_ms` / `invalid_max_duration_ms` 400 codes. Nil-`finished_at` exclusion mirrors the `finished_since` / `finished_until` family — filter symmetry preserved.
- CLI in `internal/cli/schedule.go` uses `cmd.Flags().Changed()` so an explicit `--min-duration-ms=0` is forwarded while the unset default is not — verified by the dedicated `_DurationRangeZeroExplicit` / `_DurationRangeUnsetOmitsParams` tests.
- GUI `parsePositiveInt` mirrors the API contract; blank/negative/non-integer values silently disable the bound rather than firing a 400, so the polling refresh stays well-behaved.
- OpenAPI, generated TS schema/client, mock-server (`Math.max(0, ...)` mirrors the Go clamp), Playwright spec, CLAUDE.md REST quick-reference, and ROADMAP entry are all consistent with each other.
- CI is green on both Backend build/tests and Frontend build/mock GUI tests.

Tests are thorough (boundary, both-bounds window, wide-open-excludes-still-running, whitespace-trim, compose with status/since, 400 paths including negative, paginated total-count). Nothing to flag.

---

PR #349, opened 2026-06-03
feat(schedules): skip_reason filter on run history (5.4.65)

chinmay28:
## Summary

Extends `GET /api/v1/schedules/{id}/runs` with `?skip_reason=<vm_not_found|vm_already_stopped|vm_already_running|concurrent_run|catch_up_skipped|queue_full>`, a case-insensitive exact-match on each run's `skip_reason` field, mirrored through `vmsmith schedule runs <id> --skip-reason` and a "Filter skipped runs by skip reason" dropdown in the Schedules-page recent-runs expander.

The symmetric categorical sub-axis to `?status=skipped`: `?status=` narrows the visible cohort to the four enum buckets but cannot answer *which* of the six engine-produced reasons caused the skip. Closes the SRE triage queries:

- *"show me every run skipped because the queue was full"*
- *"every catch-up skip from yesterday's daemon restart"*
- *"which schedules are getting hit by `concurrent_run` skips"*

…that the existing `?status=` / `?vm_id=` / `?since=` / `?until=` filters can narrow to the skipped cohort but cannot slice by reason.

Runs with an empty `skip_reason` — every non-skipped run, and any skipped run persisted without a reason — are filtered OUT whenever the filter is set, mirroring the nil-`finished_at` exclusion in the `finished_since` / `finished_until` family — same "only rows where the field is populated" semantics. Composes additively with every other run filter; applied right after `?status=`, so the post-filter `X-Total-Count` stays correct.

## Vertical slice

| Layer | Change |
|---|---|
| **Types** | `pkg/types/schedule_run.go` adds `IsValidScheduleRunSkipReason` (mirrors the six engine constants; rejects empty / mixed-case / typos / whitespace-padded inputs at the type layer). |
| **API** | `internal/api/handlers_schedule.go::ListScheduleRuns` parses `?skip_reason=` (lowercased + trimmed) right after `?status=` and rejects unknown values with 400 `invalid_skip_reason`; the predicate (`string(run.SkipReason) != skipReasonFilter`) drops every run whose `skip_reason` field is empty when the filter is set. |
| **CLI** | `vmsmith schedule runs <id> --skip-reason` with client-side validation (lowercased + trimmed; `--skip-reason garbage` exits before the daemon is contacted). |
| **GUI** | A 7-option `<select>` (All skip reasons + the six engine constants) alongside the status filter in the recent-runs expander. |
| **OpenAPI** | `?skip_reason=` documented on `GET /schedules/{scheduleID}/runs` with an `enum` constraint; `web/src/api/generated/schema.d.ts` regenerated. |
| **Mock-server** | Mirrors the validation contract (400 path + filter) and seeds two new runs (`run-5` skipped/vm_already_stopped, `run-6` skipped/queue_full) on `sch-1` so Playwright exercises the filter against realistic data. |
| **Docs** | `CLAUDE.md` REST quick reference + CLI map extended; `docs/ROADMAP.md` 5.4.65 entry added with ✅. |

## Test plan

- [x] **Unit** (`pkg/types`): `TestIsValidScheduleRunSkipReason` — whitelist over all six engine constants + rejection for empty / mixed-case / typo / whitespace-padded
- [x] **API integration** (`internal/api`): `TestListScheduleRuns_FilterBySkipReason` — baseline / exact match on three different reasons / case-insensitive / whitespace-trim / empty no-op / recognized-but-unused reason returns 0 / compose with `?status=skipped` / compose with `?status=success` (yields 0 because success runs have empty skip_reason) / compose with `?since=` / invalid value returns 400 `invalid_skip_reason` / filtered pagination preserves post-filter total
- [x] **CLI** (`internal/cli`): `TestCLI_ScheduleRuns_ForwardsSkipReason` (lowercase + trim), `_EmptySkipReasonOmitsParam`, `_RejectsInvalidSkipReason`
- [x] **Playwright** (`tests/web/gui.spec.js`): `skip_reason filter narrows recent runs to one skipped reason at a time` — exercises queue_full narrowing / switch to vm_already_stopped / reset to "All skip reasons" / recognized-but-unused reason produces empty cohort
- [x] Local: `CGO_ENABLED=1 go test -race ./pkg/types/ ./internal/api/... ./internal/cli/... ./internal/scheduler/` passes; `make build-go` passes; `cd web && npm run build` passes; mock-server smoke test confirms `skip_reason=queue_full` → run-6, `skip_reason=vm_already_stopped` → run-5, `skip_reason=garbage` → 400, empty `skip_reason=` → all 6 runs.

🔗 Roadmap: 5.4.65 ✅

---
_Generated by [Claude Code](https://claude.ai/code/session_01TretxxNVtA3oBuanZtJcYw)_


chinmay28 (commented):
LGTM — clean vertical slice that mirrors the existing 5.4.58 / 5.4.62 / 5.4.64 patterns. (Approval skipped because GitHub disallows approving one's own PR — treating this as a +1.)

**Correctness verified:**
- Handler predicate `if skipReasonFilter != "" && string(run.SkipReason) != skipReasonFilter { continue }` correctly excludes empty-`skip_reason` runs when the filter is set, matching the documented nil-`finished_at` semantics. Confirmed by the `status=success + skip_reason=queue_full → 0 runs` integration test.
- `IsValidScheduleRunSkipReason` covers all six engine constants from `internal/scheduler/actions.go` / `engine.go` (`vm_not_found`, `vm_already_stopped`, `vm_already_running`, `concurrent_run`, `catch_up_skipped`, `queue_full`). Rejection set in the unit test (empty / mixed-case / typo / whitespace-padded) is thorough.
- CLI lowercases + trims before validation so invalid input exits before contacting the daemon — verified by `_RejectsInvalidSkipReason`.
- Both CI checks green (Backend build + Frontend mock GUI tests). Locally verified `TestIsValidScheduleRunSkipReason` passes; the rest of the Go suite is blocked here by missing libvirt headers, but CI covers it.

**One minor nit — non-blocking:**
- `docs/openapi.yaml` places `skip_reason` (line 2337) **after** `until` rather than immediately after `status` (line 2296) where the handler filter loop and the CLAUDE.md REST quick-reference both describe it. Parameter order in OpenAPI is cosmetic, but moving the block to sit right after `status` would match the documented application order and the symmetric `?status=skipped` framing in the roadmap entry. Worth fixing in a follow-up; not blocking this PR.

---

PR #351, opened 2026-06-03
feat(vms): os_variant filter on the VM list (5.4.66)

chinmay28:
## Summary

Extends `GET /api/v1/vms` with `?os_variant=<windows-10|windows-11|windows-server-2019|windows-server-2022|windows-server-2025>`, the sub-axis of `?os_type=windows` that slices the Windows cohort by edition. Mirrored through `vmsmith vm list --os-variant <v>` and a 6-option "All variants" dropdown on the VMList page (URL round-trip via `?os_variant=`).

Closes the fleet-audit operator queries *"show me every windows-11 VM in the lab"* / *"every windows-server-2022 host in production"* / *"how many windows-server-2025 VMs are we running"* that the categorical `?os_type=windows` filter (5.6.8) can narrow to the OS family but cannot slice by edition.

The symmetric sub-axis of `?os_type=windows`: `?os_type=` narrows to the OS family, `?os_variant=` slices the Windows cohort by edition exactly the way `?skip_reason=` complements `?status=skipped` on the schedule run history (5.4.65).

## Semantics

- Case-insensitive exact-match on `spec.os_variant`; whitespace-trimmed; empty disables.
- Validated against `types.IsKnownWindowsVariant` so typos return 400 `invalid_os_variant` at the API boundary instead of silently matching nothing.
- Unlike `?os_type=linux` (which matches empty-stored VMs via the documented linux default) the `?os_variant=` filter has NO documented default — an empty stored `spec.os_variant` means "operator did not specify an edition" and is filtered OUT whenever the filter is set, mirroring the webhook `?event_type=` membership semantics.
- Applied right after `?os_type=` and before `?network=`, so it composes additively with every other VM filter; `X-Total-Count` reflects the post-filter population.

## Vertical slice

| Layer | Change |
|---|---|
| **API** | `internal/api/parse_os_variant.go` adds `parseOSVariantFilter`. `internal/api/handlers_vm.go::ListVMs` parses `?os_variant=` right after `?os_type=` and applies a case-insensitive exact-match predicate. |
| **CLI** | `internal/cli/parse_os_variant.go` mirrors the API contract. `internal/cli/vm.go` adds `vm list --os-variant <v>` with client-side validation. |
| **GUI** | `web/src/pages/VMList.jsx` adds a 6-option `<select>` (All variants + 5 known editions) next to the existing os-type dropdown, URL round-trip via `?os_variant=`, page reset on change. |
[...]

PR #352, opened 2026-06-03
feat(templates): os_variant filter on the template list (5.4.67)

chinmay28:
## Summary

Extends `GET /api/v1/templates` with `?os_variant=<windows-10|windows-11|windows-server-2019|windows-server-2022|windows-server-2025>`, the symmetric counterpart to the VM list filter (5.4.66) on the template cohort. Mirrored through `vmsmith template list --os-variant <v>` and a 6-option "All variants" dropdown on the TemplateList page (URL round-trip via `?os_variant=`).

Closes the fleet-audit operator queries *"show me every windows-11 template in the lab"* / *"every windows-server-2022 template in prod"* that the categorical `?os_type=windows` filter (5.6.8) can narrow to the OS family but cannot slice by edition — `?os_type=windows` collapses Windows 10/11/Server-2019/2022/2025 into one cohort.

> [!NOTE]
> This PR is **stacked on #351** (5.4.66 — VM list `?os_variant=`). It reuses the shared `parseOSVariantFilter` / `parseCLIOSVariant` helpers introduced there. The PR base is `claude/eager-cerf-3lWCR` so CI sees the helpers as already present; merge #351 first and this auto-rebases onto main with the helpers in place.

## Semantics

- Case-insensitive exact-match on the template's `os_variant`; whitespace-trimmed; empty disables.
- Validated against `types.IsKnownWindowsVariant` so typos return 400 `invalid_os_variant`.
- Unlike `?os_type=linux` (which matches empty-stored via the linux default) `?os_variant=` has NO documented default — an empty stored `os_variant` means "operator did not specify an edition" and is filtered OUT whenever the filter is set, mirroring the VM list 5.4.66 / webhook `?event_type=` membership semantics.
- Applied right after `?os_type=` and before `?network=`, so it composes additively with every other template filter; `X-Total-Count` reflects the post-filter population.

## Vertical slice

| Layer | Change |
|---|---|
| **API** | `internal/api/handlers_template.go::ListTemplates` parses `?os_variant=` via the shared `parseOSVariantFilter` helper (reused from 5.4.66) right after `?os_type=` and applies a case-insensitive exact-match predicate against `tpl.OSVariant`. |
| **CLI** | `internal/cli/template.go` adds `template list --os-variant <v>` using the shared `parseCLIOSVariant` helper with client-side validation. |
| **GUI** | `web/src/pages/TemplateList.jsx` adds a 6-option `<select>` (All variants + 5 known editions) next to the os-type dropdown, URL round-trip via `?os_variant=`, page reset on change. |
| **Client** | `web/src/api/client.ts` extends `templates.list({ osVariant })` with the five-value type union. |
| **OpenAPI** | `docs/openapi.yaml` adds the `OSVariantFilter` `$ref` to `/templates` and generalises the component description so it documents both VM and template usage; `web/src/api/generated/schema.d.ts` regenerated. |
| **Mock-server** | `tests/web/mock-server.js` mirrors validation (400 path + filter + no-empty-match contract) and adds a fourth seed template (`windows-11-desktop`) so the GUI test exercises variant slicing. |
| **Docs** | `CLAUDE.md` REST quick reference + CLI map extended; `docs/ROADMAP.md` 5.4.67 entry added with ✅. |

## Test plan

- [x] **API integration** (`internal/api/api_test.go`): 7 cases (`TestListTemplates_FilterByOSVariant_ExactMatch`, `_IsCaseInsensitive`, `_TrimsWhitespace`, `_EmptyIsNoOp`, `_ExcludesEmptyStored`, `_InvalidValueReturns400`, `_ComposesWithOSType`, `_TotalCountReflectsFiltered`).
- [x] **CLI** (`internal/cli/commands_test.go`): 6 cases (`TestCLI_TemplateList_FilterByOSVariant_ExactMatch`, `_IsCaseInsensitive`, `_EmptyOmitsFilter`, `_ExcludesEmptyStored`, `_RejectsInvalidValue`, `_ComposesWithOSType`).
- [x] **Playwright** (`tests/web/gui.spec.js`): `os-variant filter narrows the template list to a single Windows edition and round-trips through the URL` — exercises matching-variant narrowing (`windows-server-2022` → only windows-2022) / recognised-but-unmatched empty cohort (`windows-10` → empty) / switch to `windows-11` → only windows-11-desktop / reset to "All variants" / URL round-trip.
- [x] Local: `CGO_ENABLED=1 go test -race ./internal/api/... ./internal/cli/... ./pkg/types/...` passes; `go build ./...` clean; `npm run build` clean; `golangci-lint run` clean; mock-server smoke-test via `curl` confirms the filter contract end-to-end.

🔗 Roadmap: 5.4.67 ✅

---
_Generated by [Claude Code](https://claude.ai/code/session_01EagerCerfBaS1R)_


chinmay28 (commented):
LGTM — would approve but the API doesn't allow self-approval.

Reviewed locally (after checking out on top of #351's head `4ca7ecf`) and against CI. Both CI jobs are green, and the local suite passes:

- `CGO_ENABLED=1 go vet ./...` — clean
- `make fmt-check` — clean
- `CGO_ENABLED=1 go test -race -count=1 ./internal/api/... ./internal/cli/... ./pkg/types/...` — pass, including the 8 new `TestListTemplates_FilterByOSVariant_*` cases and 6 new `TestCLI_TemplateList_FilterByOSVariant_*` cases.

This is a textbook stacked PR on #351 — the diff against `4ca7ecf` is just the template-specific surface (handler + CLI + GUI + Playwright + mock-server + OpenAPI generalisation + roadmap entry), reusing the shared `parseOSVariantFilter` / `parseCLIOSVariant` helpers from #351 verbatim. No drift between the two endpoints.

Nice touches I noticed:
- The `OSVariantFilter` OpenAPI parameter description gets generalised in this PR ("spec.os_variant on /vms, os_variant on /templates") rather than duplicating the parameter — the right call given both endpoints share identical semantics.
- The mock-server gains a fourth seed template (`windows-11-desktop`, `os_variant=windows-11`) so the Playwright test can exercise switching between two matching variants rather than only matching/unmatching one. The dependent test updates in `tests/web/gui.spec.js` that bump `toHaveCount(4)` → `toHaveCount(5)` for the unrelated VM-create template dropdown are correct and necessary (the seed now has 4 templates instead of 3).
- The `?os_variant=` filter is slotted right after `?os_type=` and before `?network=`, matching the position in the VM handler. The roadmap commentary about composes-additively-with-everything is borne out by the `_ComposesWithOSType` test.
- The "unusual but allowed" linux-typed template carrying an `os_variant` string is explicitly tested in `_ComposesWithOSType` — the right thing to verify since `os_variant` is a free-form string at the storage layer.

Note: this PR's base is `claude/eager-cerf-3lWCR` (#351's branch), so merge #351 first and this auto-rebases onto main.

---

PR #355, opened 2026-06-04
feat(vms): firmware filter on the VM list (5.4.68)

chinmay28:
## Summary

Extends `GET /api/v1/vms` with `?firmware=<bios|uefi|ovmf>`, a case-insensitive exact-match on each VM's `spec.firmware`, mirrored through `vmsmith vm list --firmware <v>` and a 4-option "All firmware / BIOS / UEFI / OVMF" dropdown on the VMList page (URL round-trip via `?firmware=`).

Closes the fleet-audit operator queries *"which VMs boot via OVMF / which can boot Windows 11"* / *"which VMs still use SeaBIOS"* / *"how many lab machines are on UEFI"* that the categorical `?os_type=windows` filter (5.6.8) can narrow to the OS family but tells you nothing about firmware choice — a Windows guest can be on BIOS, UEFI, or OVMF; a Linux guest can equally be on UEFI for a coreboot-compatible host.

## Semantics

- Case-insensitive exact-match on `spec.firmware`; whitespace-trimmed; empty disables.
- Validates the value against the existing three-value `FirmwareBIOS` / `FirmwareUEFI` / `FirmwareOVMF` vocabulary so typos return 400 `invalid_firmware` at the API boundary, matching the create-path validation contract on `POST /vms`.
- `?firmware=bios` matches both stored `"bios"` AND VMs with no firmware override — the SeaBIOS default, mirroring the `?os_type=linux` empty-means-linux contract.
- `?firmware=uefi` and `?firmware=ovmf` strict-match the stored value because uefi and ovmf are stored separately even though they map to the same libvirt `firmware='efi'` attribute at render time — preserves the operator's chosen alias on the filter round-trip.
- Applied right after `?os_variant=` and before `?network=`, so it composes additively with every other VM filter; `X-Total-Count` reflects the post-filter population.

## Vertical slice

| Layer | Change |
|---|---|
| **API** | `internal/api/parse_firmware.go` adds `parseFirmwareFilter` + `vmMatchesFirmwareFilter`. `internal/api/handlers_vm.go::ListVMs` parses `?firmware=` right after `?os_variant=` and applies the predicate. |
| **CLI** | `internal/cli/parse_firmware.go` mirrors the API contract. `internal/cli/vm.go` adds `vm list --firmware <v>` with client-side validation. |
| **GUI** | `web/src/pages/VMList.jsx` adds a 4-option `<select>` (All firmware + BIOS / UEFI / OVMF) right next to the existing os-variant dropdown, URL round-trip via `?firmware=`, page reset on change. |
| **Client** | `web/src/api/client.ts` extends `vms.list({ firmware })` with the three-value type union. |
| **OpenAPI** | `docs/openapi.yaml` adds the `FirmwareFilter` `$ref` to `GET /vms`; `web/src/api/generated/schema.d.ts` regenerated. |
| **Mock-server** | `tests/web/mock-server.js` mirrors the validation (400 path + filter + empty-means-bios contract) and pins the existing `win-app` Windows seed VM to `spec.firmware=uefi` so the Playwright test can exercise UEFI strict-match vs the implicit-BIOS Linux cohort. |
| **Docs** | `CLAUDE.md` REST quick reference + CLI map extended; `docs/ROADMAP.md` 5.4.68 entry added with ✅. |

## Test plan

- [x] **API integration** (`internal/api/api_test.go`): 9 cases (`TestListVMs_FilterByFirmware_ExactMatchUEFI`, `_ExactMatchOVMF`, `_BIOSMatchesEmptyStored`, `_IsCaseInsensitive`, `_TrimsWhitespace`, `_EmptyIsNoOp`, `_InvalidValueReturns400`, `_ComposesWithOSType`, `_TotalCountReflectsFiltered`).
- [x] **CLI** (`internal/cli/commands_test.go`): 6 cases (`TestCLI_VMList_FilterByFirmware_ExactMatchUEFI`, `_BIOSMatchesEmptyStored`, `_IsCaseInsensitive`, `_EmptyOmitsFilter`, `_RejectsInvalidValue`, `_ComposesWithOSType`).
- [x] **Playwright** (`tests/web/gui.spec.js`): `firmware filter narrows the VM list and round-trips through the URL` — exercises UEFI strict-match → only win-app / OVMF empty cohort / BIOS matches the two empty-stored Linux VMs / reset to All firmware / URL round-trip.
- [x] Local: `CGO_ENABLED=1 go test -race ./internal/api/... ./internal/cli/... ./pkg/types/...` passes; `go build ./...` clean; `go vet ./...` clean; `npm run build` clean; mock-server smoke-test via `curl` confirms the filter contract end-to-end (`?firmware=uefi` → 1 vm, `?firmware=bios` → 2 vms, `?firmware=ovmf` → 0 vms, `?firmware=garbage` → 400).

🔗 Roadmap: 5.4.68 ✅

---
_Generated by [Claude Code](https://claude.ai/code/session_01EagerCerfJogWf)_


chinmay28 (commented):
## Review — would approve if I could (GitHub blocks self-approval)

The `?firmware=` filter is a faithful mirror of the established `?os_type=` / `?os_variant=` pattern, the contract claims hold under inspection, and CI is green on both backend and frontend jobs.

### Strengths

- **Pattern consistency.** Two-file split (`internal/api/parse_firmware.go` + `internal/cli/parse_firmware.go`) matches the precedent set by `parse_os_type.go` / `parse_os_variant.go`. Filter chain placement in `handlers_vm.go::ListVMs` (after `os_variant`, before `network`) is correct.
- **Empty-means-BIOS predicate is implemented exactly right** — the contract claim *"`?firmware=bios` matches stored `bios` AND empty"* lines up with `pkg/types/vm.go::ResolvedFirmwareAttr()` where both empty and `"bios"` render to no firmware attribute:

  ```go
  // internal/api/parse_firmware.go:49-55
  func vmMatchesFirmwareFilter(spec types.VMSpec, filter string) bool {
      stored := strings.ToLower(strings.TrimSpace(spec.Firmware))
      if filter == types.FirmwareBIOS {
          return stored == "" || stored == types.FirmwareBIOS
      }
      return stored == filter
  }
  ```

  And `uefi` / `ovmf` strict-match preserves the operator's chosen alias on the round-trip, which is the right call given both render to libvirt's `firmware='efi'` shorthand.

- **Validation parity with the create path.** Same three-value vocabulary, same `invalid_firmware` 400 code as `validation.go::validateDeviceOverrides`.
- **Test coverage matches the claim** — 9 API cases, 6 CLI cases, 1 Playwright case exercise every advertised contract corner (case-insensitive, trim, empty-disables, BIOS-matches-empty, UEFI/OVMF strict-distinct, invalid→400, composes with `os_type`, post-filter `X-Total-Count`).
- **Mock-server mirrors the validation contract**, so the Playwright assertion exercises the same 400 / filter / empty-means-bios behavior as the real server.
- CLAUDE.md and ROADMAP.md 5.4.68 entries accurately describe what shipped.

### Nits (non-blocking, follow-up worthy)

1. **Inconsistent error message wording.** `parseFirmwareFilter` returns `"firmware must be one of: bios, uefi, ovmf"` while the create-path `validateDeviceOverrides` returns `"firmware must be one of %q, %q, %q"`. Both share `invalid_firmware` so machine clients are unaffected — but a unified human message would be nicer. The existing `parseOSVariantFilter` already uses the colon style, so the new code is internally consistent with its peer; harmonizing the create-path message could be a separate cleanup.

2. **Slightly weaker baseline assertion.** `TestListVMs_FilterByFirmware_TotalCountReflectsFiltered` asserts the post-filter `X-Total-Count=3` but doesn't first assert the pre-filter total to lock in the seed-count invariant. The analogous `os_variant` test does. Optional.

3. **Duplication opportunity.** `vmMatchesFirmwareFilter` and `cliVMMatchesFirmwareFilter` are byte-identical. Same observation applies to the existing `os_type` / `os_variant` filter helpers, so deferring this lift into `pkg/types` is consistent with current codebase style — but worth keeping in mind if a fourth `Resolved*` filter ever lands.

4. **Docs micro-precision.** The PR description says *"applied right after `?os_variant=` and before `?network=`"*, which is technically true but elides that `autoStart` and `locked` sit between firmware and network in the loop. Cosmetic only.

### Test run note

Could not exercise `CGO_ENABLED=1 go test ./internal/api/...` locally — sandbox lacks libvirt-dev system headers. This is environmental, not a PR issue. `pkg/types/...` tests passed, and the CI run on this PR covers the cgo-enabled tests successfully.

Nice incremental fleet-audit slice. 👍


---

PR #361, opened 2026-06-05
feat(api): add console websocket proxy

march26bot:
## Summary
- add the backend websocket proxy for `GET /api/v1/vms/{id}/console`
- proxy raw VNC bytes over the `binary` websocket subprotocol and track active sessions per VM
- cover happy-path echo, ticket reuse/expiry, dial failure, and TLS mismatch in API tests

## Testing
- `git diff --check`
- not run: `go test` (Go toolchain unavailable in this environment)


chinmay28 (changes requested):
The core handler is solid and the documented roadmap item 5.1.4 is correctly addressed — single-use ticket enforcement, expiry, dial failure, and TLS-mismatch paths are correctly implemented and tested. But there are several issues worth resolving before merge:

## Blocking

1. **The branch is severely out of date with `origin/main`.** `git merge-base pr/361 origin/main` returns no common ancestor; the branch is 287 commits long while `origin/main` has only 50 commits, and the comparison surfaces a 90-file / 13,807-deletion diff that is purely history-mismatch noise (the PR's actual commit `721f65b` only touches 7 files). Please rebase onto current `main` so reviewers see only the 558-line websocket-proxy diff and so CI can run. The PR description says "not run: `go test`" and the GitHub Checks tab shows zero check runs against this PR head — that's because the branch is too divergent for CI to wire up.

2. **`docs/openapi.yaml` is not updated.** The new `GET /api/v1/vms/{vmID}/console` route is missing from the OpenAPI spec. The ticket-issuance endpoint (5.1.2) is documented at `/vms/{vmID}/console/ticket`, but the proxy endpoint (this PR) needs a sibling entry with the `ticket` query parameter, the websocket subprotocol negotiation note, and the 401 / 403 / 502 / 503 / 429 error responses.

3. **`CLAUDE.md` REST API quick-reference is not updated.** The table currently only lists `POST /vms/{id}/console/ticket`; please add the `GET /vms/{id}/console` row alongside it so the operator-facing docs reflect the wired route.

## Non-blocking, but worth addressing

4. **`CheckOrigin` returns `true` unconditionally** (`handlers_console_ws.go:30`). The single-use ticket query parameter is the sole CSRF defense — that's acceptable since the ticket is an unguessable bearer, but it should be documented (a comment on the upgrader + a note in `WINDOWS_GUESTS.md` / the console section of `ARCHITECTURE.md`) so the threat model is explicit.

5. **TLS detection doesn't fall back to `X-Forwarded-Proto`** (`handlers_console_ws.go`, the `s.tlsEnabled && r.TLS == nil` check). Behind a TLS-terminating reverse proxy the daemon sees `r.TLS == nil` even though the client used HTTPS, so a legitimate `wss://` upgrade would be rejected with 403 `mixed_content_blocked`. If reverse-proxy deployments are in scope (the install docs mention them), the check should consult `X-Forwarded-Proto` when present.

6. **`apiKey` from `ConsumeTicket` is discarded** (`_ = apiKey` at line 72). The bearer token associated with the ticket-issuance call is dropped on the floor — if console sessions should be attributed to the issuing API key (for the audit-log / event-emission paths in roadmap 5.1.5), this is the spot to thread it through.

7. **Test gaps:**
   - No coverage for the `MaxConcurrentSessions` 429 `console_session_limit_reached` path.
   - No coverage for the `MaxSessionSeconds` timer-triggered close.
   - No coverage for `consoleStore == nil` → 503.
   - `TestProxyConsole_ClosesOnServerShutdownSignal` admits in its own comment that it doesn't actually exercise the shutdown channel — it tests backend-close unwinding instead. Worth tightening into two separate tests.

8. **Control flow in `proxyConsoleTCPToWebSocket`** is a little murky when `n > 0 && err != nil && !errors.Is(err, io.EOF)`: neither the `n == 0` early-returns nor the EOF clean-exit triggers, and the loop continues with the residual data already written. Not a leak, but worth tightening with an explicit return on non-EOF deadline errors after writing the partial.

9. `consoleWriteWait = 10 * time.Second` is hard-coded — consider promoting to `daemon.console.write_timeout_seconds` alongside the existing `idle_timeout_seconds` / `max_session_seconds` knobs from 5.1.6 so the deadline can be tuned per deployment.

Happy to re-review once the rebase + docs are in.

---

chinmay28 (commented):
Reviewed locally — well-scoped change, mostly good. The proxy itself works as advertised. A few items to discuss before merge.

## Build / tests
- `CGO_ENABLED=1 go build ./...` clean.
- All 6 new `ProxyConsole*` tests pass.
- `internal/console` tests pass.
- One **unrelated** failure on this branch: `TestTestWebhook_EndToEndFailure` (handlers_webhook_test.go:588, `LastStatus = 500, want 0`). The branch is 6 commits behind main, and this test pre-dates this PR — rebasing should clear it. The CI checks haven't run on this PR yet (likely needs approval since the author is an external bot).

## Good
- Uses `gorilla/websocket v1.5.3`, properly added to `go.mod` / `go.sum`.
- Honors the single-use ticket contract: `consoleStore.ConsumeTicket(ticket, vmID)` is called before upgrade; reuse test confirms 401 on second dial.
- Adds `Hijack()` to `responseRecorder` in `middleware.go:92-98` so the upgrade survives the logging middleware — good catch.
- `redactSensitiveRawQuery` already redacts `ticket=` in access logs (`middleware.go:170-173`), so tickets don't leak through the request log.
- Uses `Subprotocols: []string{"binary"}` and the test asserts the negotiated subprotocol.
- Per-VM session tracking via `s.consoleSessions` + a global `MaxConcurrentSessions` cap.

## Discussion items

### 1. `r.Context().Done()` is dead after a websocket upgrade — `internal/api/handlers_console_ws.go:122-127`
```go
select {
case <-r.Context().Done():
case <-s.shutdownNotify:
case <-maxSession:
case <-errCh:
}
```
After `Upgrade`, the Go HTTP server detaches the request and its context is no longer cancelled on client disconnect. Disconnect is still detected via `errCh` (the read loop returns), so functionally the proxy exits, but this arm is dead and misleading. Either drop it or replace with `wsConn.UnderlyingConn().Closed`-style detection.

### 2. Ticket consumed before backend dial — `internal/api/handlers_console_ws.go:67-95`
If `GetConsoleEndpoint` or `net.DialTimeout` fails (transient VNC-startup race after a `vm start`), the single-use ticket is already burned and the client must request a new one. Consider deferring `ConsumeTicket` until after a successful dial, or surfacing a `retry-after` hint on the 502 so the GUI can auto-reissue.

### 3. `apiKeyAuth` still applies to `/console` — likely defeats the ticket dance for browsers
When `daemon.api_key` is configured, browsers connecting to `wss://.../console?ticket=...` will still need to attach the API key in the URL (WebSocket can't set `Authorization` headers). That re-introduces the exact problem the single-use ticket flow exists to solve. Either exempt `/console` from `apiKeyAuth` (the ticket is the cap) or document that the GUI must pass the API key alongside the ticket and accept the trade-off.

### 4. Branch staleness — please rebase onto main
The branch is 6 commits behind. Once rebased, the unrelated webhook test will pass and CI can run cleanly.

### Minor
- No `SetWriteDeadline` on the initial Upgrade write path — minor, the proxy goroutines already deadline writes.
- `CheckOrigin: return true` is permissive; OK while the ticket + API key gate the upgrade, but worth a comment noting this is intentional.

Happy to approve once the rebase lands and items 1–3 are either addressed or explicitly waived.

---

chinmay28 (changes requested):
Second-pass review on `e8b9143`. The follow-up commit addresses most of the prior review feedback — but one mechanical issue will block CI.

## What got addressed in `e8b9143`

| Prior concern | Status |
|---|---|
| Blocking #1 — rebase onto current `main` | ✅ Done — branch is now 2 commits ahead of `main` with no divergence |
| Blocking #2 — `docs/openapi.yaml` `GET /vms/{vmID}/console` entry | ✅ Added (40 lines) |
| Blocking #3 — `CLAUDE.md` REST quick reference | ✅ Added the `GET /vms/{id}/console` row |
| Non-blocking #4 — `CheckOrigin = true` documentation | ✅ Comment on the upgrader + `docs/ARCHITECTURE.md` note on the open-origin policy |
| Non-blocking #5 — TLS detection falls back to `X-Forwarded-Proto` | ✅ New `consoleRequestUsedTLS()` helper consults `r.TLS` then `X-Forwarded-Proto: https`; covered by `TestProxyConsole_HonorsForwardedProtoForTLSDeployments` |
| Non-blocking #6 — `apiKey` discarded after `ConsumeTicket` | ✅ Now stored on `activeConsoleSession` for the 5.1.5 audit / session-termination follow-up |
| Non-blocking #7 — test gaps (`MaxConcurrentSessions` 429, `MaxSessionSeconds`, `consoleStore == nil` → 503) | ✅ All three added: `TestProxyConsole_RejectsWhenSessionLimitReached`, `TestProxyConsole_ClosesWhenMaxSessionDeadlineExpires`, `TestProxyConsole_NoStoreReturns503` |
| Non-blocking #8 — `proxyConsoleTCPToWebSocket` control flow when `n > 0 && err != nil && !errors.Is(err, io.EOF)` | ✅ Simplified — always returns on read error after writing the partial |

## Local verification on `e8b9143`

- `CGO_ENABLED=1 go test -count=1 -race ./internal/api/...` — all green, including all 10 `TestProxyConsole_*` cases
- `CGO_ENABLED=1 go vet ./...` — clean
- The previously-flagged unrelated webhook test (`TestTestWebhook_EndToEndFailure`) now passes after the rebase

## Blocker — gofmt violation

`internal/api/handlers_console_ws.go` is not gofmt-formatted under the pinned Go 1.22.5 toolchain. The CI workflow runs `make fmt-check` (`.github/workflows/ci.yml`) which will reject the PR until this is fixed.

```
$ GOTOOLCHAIN=go1.22.5 gofmt -l .
internal/api/handlers_console_ws.go
```

The diff:

```diff
@@ -113,8 +113,12 @@
 	wsConn.SetReadLimit(1 << 20)
 
 	errCh := make(chan error, 2)
-	go func() { errCh <- proxyConsoleWebSocketToTCP(wsConn, targetConn, time.Duration(s.consoleConfig.IdleTimeoutSeconds)*time.Second) }()
-	go func() { errCh <- proxyConsoleTCPToWebSocket(targetConn, wsConn, time.Duration(s.consoleConfig.IdleTimeoutSeconds)*time.Second) }()
+	go func() {
+		errCh <- proxyConsoleWebSocketToTCP(wsConn, targetConn, time.Duration(s.consoleConfig.IdleTimeoutSeconds)*time.Second)
+	}()
+	go func() {
+		errCh <- proxyConsoleTCPToWebSocket(targetConn, wsConn, time.Duration(s.consoleConfig.IdleTimeoutSeconds)*time.Second)
+	}()
```

Run `make fmt` (or `gofmt -w internal/api/handlers_console_ws.go`) and push.

## Carried-over discussion items (non-blocking, still applies)

- **`r.Context().Done()` arm is dead after upgrade** (`handlers_console_ws.go:122`). After `Upgrader.Upgrade` hijacks the connection, the request context is no longer cancelled on client disconnect — the `errCh` arm catches that path instead. Either drop the dead arm or replace with a `wsConn`-based disconnect signal. Functionally harmless, but misleading.
- **Ticket consumed before backend dial** (`handlers_console_ws.go:71-92`). If `GetConsoleEndpoint` or `net.DialTimeout` fails (transient VNC-startup race after a `vm start`), the single-use ticket is burned and the client must request a new one. Consider deferring `ConsumeTicket` until after a successful dial, or returning a `Retry-After`-style hint on the 502 so the GUI can auto-reissue.
- **`apiKeyAuth` still gates `/console` from browsers** — partially mitigated by `extractAPIKey` accepting `?api_key=` as a URL fallback (`auth_middleware.go:50`), so browsers connecting to `wss://.../console?ticket=...&api_key=...` will work. Worth mentioning in the operator-facing console docs (the `WINDOWS_GUESTS.md` console section or the README) so this isn't a deployment surprise.
- **`consoleWriteWait = 10 * time.Second` is hardcoded** — consider promoting to `daemon.console.write_timeout_seconds` alongside the existing `idle_timeout_seconds` / `max_session_seconds` knobs from 5.1.6.

Happy to re-review and approve once `gofmt` is fixed. The substance of the change is solid.

---

chinmay28 (approved):
## Review summary

LGTM — solid implementation of roadmap 5.1.4. The ticket → upgrade → bidirectional proxy flow is well-structured, the security boundary (single-use ticket, query-string redaction in the request logger, TLS-protocol gate, `X-Forwarded-Proto` honoring for reverse-proxy deployments) is clearly thought through, and the `Hijack()` plumbing through `responseRecorder` is the right fix for the upgrade-under-middleware case.

### Verification performed
- ✅ `CGO_ENABLED=1 go test -count=1 ./internal/api/... ./internal/console/... ./internal/vm/... ./pkg/types/...` passes locally on `32915c8`
- ✅ All 10 new `TestProxyConsole_*` cases pass with verbose output; ticket is shown as `query=ticket=REDACTED` in the request log (existing middleware redaction continues to work)
- ✅ `gofmt -l .` clean
- ✅ `CGO_ENABLED=1 go vet ./...` clean
- ✅ TLS gate: 403 `mixed_content_blocked` when `cfg.Daemon.TLS.{CertFile,KeyFile}` set and request is plain `ws://` without `X-Forwarded-Proto: https`
- ✅ Session cap (429) and max-session deadline (1 s test waits 3 s and confirms close) both wired and exercised
- ✅ Shutdown path: `BeginShutdown()` closes the websocket via `s.shutdownNotify`
- ✅ Architecture doc clearly justifies the open `CheckOrigin` (security boundary is the ticket, not browser-origin matching) and warns downstream loggers to treat `ticket=` as a secret — this is the right contract for a VNC-over-WS proxy
- ✅ Pre-existing data race in `TestJanitorRemovesExpiredTicketsAutomatically` (5.1.1 console store test) reproduces on `main` too — **not introduced by this PR**, but worth fixing in a follow-up so `make test` (which runs with `-race`) stays green

### Minor observations (non-blocking)

1. **`proxyConsoleTCPToWebSocket` has a dead branch** at lines 183–188 (`internal/api/handlers_console_ws.go`):
   ```go
   if err != nil {
       if errors.Is(err, io.EOF) {
           return err
       }
       return err
   }
   ```
   Both arms return `err`. Either drop the `errors.Is` branch or differentiate the return (e.g., `return nil` on EOF if you want clean close semantics). Cosmetic.

2. **`sanitizeManagerError(err)` is called twice** in the `GetConsoleEndpoint` error path (lines 81 and 91). Hoist to a local — purely a micro-cleanup, not a correctness issue.

3. **Session-count check has a small TOCTOU race** — `consoleSessionCount()` and `registerConsoleSession()` are separate critical sections, so N concurrent dials at the cap boundary can all pass the check before registering. The realistic console-open rate makes this benign; flagging for future hardening (token-bucket or count-and-register-under-one-lock).

4. **`activeConsoleSession.apiKey` is currently unused** — it's read off the consumed ticket and stashed but nothing reads it back. That's expected (5.1.5 will surface it for force-close attribution), but worth a brief code comment so a future reader doesn't delete the field as dead code.

5. **No test for `IdleTimeoutSeconds`** — the max-session deadline test covers wall-clock close, but the idle-deadline path (no traffic for `idle_timeout_seconds`) isn't exercised. Optional follow-up.

Roadmap update for 5.1.4 in this PR is accurate. Approving.

---

chinmay28 (commented):
## Confirming green on `32915c8` (post-approval status check)

The PR is already approved (review id 4443308239). Re-ran the verification locally as part of the open-PR sweep:

- ✅ `CGO_ENABLED=1 go test -count=1 ./internal/api/... ./internal/cli/... ./pkg/types/...` passes
- ✅ All 10 `TestProxyConsole_*` cases pass on this head
- ✅ `gofmt -l .` clean
- ✅ `CGO_ENABLED=1 go vet ./...` clean
- ✅ CI green on this head (Backend + Frontend mock GUI)
- ✅ Pre-existing race in `TestJanitorRemovesExpiredTicketsAutomatically` reproduces on `origin/main` as well — confirmed not introduced by this PR, worth a separate follow-up so `make test` (`-race`) stays green

The three prior-review non-blocking items (dead `err`-discriminator branch in `proxyConsoleTCPToWebSocket`, double `sanitizeManagerError` call, TOCTOU on session-count check) are still present on this head but were explicitly waived as non-blocking. No further changes needed before merge.

Ready to merge from my side.

---

PR #362, opened 2026-06-05
feat(ports): ?guest_ip= filter on the port-forward list (5.4.73)

chinmay28:
## Summary

Extends `GET /api/v1/vms/{id}/ports` with `?guest_ip=<addr>`, a case-insensitive exact-match on the rule's `guest_ip`, mirrored through `vmsmith port list <vm-id> --guest-ip <addr>` and a 250 ms-debounced text input on the VMDetail port-forward card (URL round-trip via `?port_guest_ip=`).

Closes the multi-NIC audit operator queries:

- *"show me every forward landing on 192.168.100.50"*
- *"which rules route to the data-net interface (10.0.0.7) vs the storage-net interface (10.0.0.8)"*

That the existing `?search=` filter can only fuzzy-match — searching for `192.168.100.50` also surfaces `192.168.100.5` because `search` does substring matching, so the operator can't disambiguate the two on a busy jumpbox VM.

## Semantics

- **Free-form value** (unlike the closed-vocabulary peers on this surface — `?protocol=` / `?nic_model=` / `?disk_bus=`): no validation rejection; garbage simply matches no rules. `guest_ip` is a value operators paste verbatim (IPv4 dotted-quad or IPv6 literal).
- **Case-insensitive**, whitespace-trimmed; empty disables. Case-insensitivity is meaningful for IPv6 literals (operators routinely paste either `fe80::ABCD` or `FE80::abcd`) and a no-op for IPv4 dotted-quads.
- Applied right after the guest-port range and before `?search=` so it composes additively with every other port-forward filter (`?tag=` / `?protocol=` / `?min_host_port=` / `?max_host_port=` / `?min_guest_port=` / `?max_guest_port=` / `?search=`); `X-Total-Count` reflects the post-filter population.

## Why a new filter when `?search=` already covers it

`?search=` is a substring filter against the whole port-forward record (description, protocol, host_port, guest_port, guest_ip, tags). For multi-NIC operators that's exactly the wrong shape — querying for `192.168.100.50` is supposed to mean "the rules landing on 192.168.100.50", but it also returns rules on `192.168.100.5`, `192.168.100.500`, etc. The `?guest_ip=` filter pins the operator's intent to exact-match.

## Vertical slice

| Layer | Change |
|---|---|
| **API** | `internal/api/parse_guest_ip.go` adds `parseGuestIPFilter` + `portMatchesGuestIPFilter`; `handlers_network.go::ListPorts` parses `?guest_ip=` right after the guest-port range. |
| **CLI** | `internal/cli/parse_guest_ip.go` mirrors the API contract; `network.go` adds `port list --guest-ip <addr>`. |
| **GUI** | `web/src/pages/VMDetail.jsx` adds a 250 ms-debounced text input (`data-testid="port-guest-ip-filter"`) + "Clear guest IP" button below the guest-port range, URL round-trip via `?port_guest_ip=`, page reset on change. |
| **Client** | `web/src/api/client.ts` extends `ports.list({ guestIp })`. |
| **OpenAPI** | `docs/openapi.yaml` adds the `guest_ip` query parameter to `GET /vms/{vmID}/ports`; `web/src/api/generated/schema.d.ts` regenerated. |
| **Mock server + Playwright** | `tests/web/mock-server.js` mirrors the filter (whitespace-trim + lowercase compare); `tests/web/run-gui-tests.js` adds a route-interception fixture with a multi-NIC layout (two rules on `192.168.100.10`, one on `10.0.0.7`) so the filter has a meaningful cohort to slice; matching `gui.spec.js` case for the Playwright test framework. |
| **Docs** | `CLAUDE.md` REST quick reference + CLI map extended; `docs/ROADMAP.md` adds the 5.4.73 entry with ✅. |

## Test plan

- [x] **API integration** (`internal/api/api_test.go`): 9 cases — `TestListPorts_FilterByGuestIP_ExactMatch`, `_OtherCohort`, `_CaseInsensitive`, `_TrimsWhitespace`, `_EmptyIsNoOp`, `_NoMatch`, `_ComposesWithProtocol`, `_ComposesWithGuestPortRange`, `_TotalCountReflectsFiltered`.
- [x] **CLI** (`internal/cli/commands_test.go`): 6 cases — `TestCLI_PortList_FilterByGuestIP_ExactMatch`, `_OtherCohort`, `_CaseInsensitive`, `_EmptyOmitsFilter`, `_NoMatch`, `_ComposesWithProtocol`.
- [x] **Playwright** (mock-runner `tests/web/run-gui-tests.js` + `tests/web/gui.spec.js`): `guest_ip filter narrows the port-forward list (5.4.73)` — exercises 192.168.100.10 cohort matches 2 primary-NIC rules / switch to 10.0.0.7 surfaces only the data-net rule / clear restores all three rules / URL round-trip via `?port_guest_ip=`. Total mock GUI count: 66 → 67.
- [x] Local: `CGO_ENABLED=1 go test -race -count=1 ./internal/api/... ./internal/cli/... ./pkg/types/... ./internal/vm/...` passes; `go vet ./...` clean; `gofmt -l .` clean; `npm run build` clean; `node tests/web/run-gui-tests.js` passes (67/67).

🔗 Roadmap: 5.4.73 ✅

---

---
_Generated by [Claude Code](https://claude.ai/code/session_01CpjC4vCvXzsCgGydQyRXkS)_


chinmay28 (commented):
LGTM (can't formally approve as PR author). Roadmap 5.4.73 — `?guest_ip=` on the port-forward list:

- `parse_guest_ip.go` (API + CLI) is intentionally simpler than the closed-vocabulary peers on this surface (no validation, garbage simply matches no rules) — the right call for a free-form value operators paste verbatim. The reasoning in the doc comment about why this differs from `?protocol=` is clear.
- Case-insensitive matching is meaningful for IPv6 literals (`fe80::ABCD` vs `FE80::abcd`) and a no-op for IPv4 — documented in the comment, which avoids the "why this filter is case-insensitive when `vm_id` isn't" question down the line.
- The "why `?guest_ip=` rather than just `?search=`" framing in the PR description nails the issue: `?search=` for `192.168.100.50` also surfaces `192.168.100.5` because of substring semantics, so the new filter pins operator intent to exact-match.
- Filter slot: right after the guest-port range and before `?search=`, matching documented ordering. Filters compose additively; `X-Total-Count` reflects post-filter cohort.
- Coverage: 9 API cases, 6 CLI cases, 1 Playwright case (multi-NIC fixture with two rules on `192.168.100.10` and one on `10.0.0.7` — exactly the cohort that exposes the substring-vs-exact-match difference). CI green.

Comparison to alternatives: I briefly considered whether the filter should accept CIDR ranges (e.g. `?guest_ip=192.168.100.0/24`), but exact-match is the right scope here — CIDR would be a separate `?guest_ip_cidr=` axis if it shows up as an operator need.

---

PR #363, opened 2026-06-05
feat(schedules): last_fired time-range filter on the schedule list (5.4.74)

chinmay28:
## Summary

Extends `GET /api/v1/schedules` with `?last_fired_since=<rfc3339>` / `?last_fired_until=<rfc3339>`, inclusive bounds on each schedule's `last_fired_at`, mirrored through `vmsmith schedule list --last-fired-since --last-fired-until` and two `<input type="datetime-local">` "Last fired since / until" controls on the Schedules page (URL round-trip via `?last_fired_since=` / `?last_fired_until=`).

Closes the SRE triage operator queries:

- *"which schedules fired during yesterday's maintenance window"*
- *"which schedules haven't fired since the last daemon restart"*
- *"which nightly-snapshot schedules went silent over the weekend"*

That the categorical `?enabled=` filter (which records whether the schedule is *administratively* enabled, not whether it actually fired) and the `?next_fire_since=` / `?next_fire_until=` filter (5.4.60) on the *next* fire — which orders by what *will* happen — cannot answer. The symmetric backward-looking counterpart to the existing forward-looking `next_fire_since` / `next_fire_until` range, completing the past-and-future time-range filter pair on the schedule list exactly the way `?finished_since=` / `?finished_until=` (5.4.62) complements `?since=` / `?until=` on the run-history surface.

## Semantics

- Reuses the same `parseTimeRangeParam` helper as `?since=` / `?until=` and the same `snapshotInTimeRange` predicate so the contract matches exactly: whitespace-trimmed; empty disables the bound; invalid values return 400 `invalid_last_fired_since` / `invalid_last_fired_until`.
- **Schedules with a nil `last_fired_at` (never-fired) are filtered OUT whenever any bound is set**, mirroring the nil-handling for the schedule `?next_fire_since=` / `?next_fire_until=` filter (5.4.60) and the webhook `?last_delivery_since=` / `?last_delivery_until=` filter (5.4.61).
- Applied right after the `next_fire_at` range and before `?search=`, so it composes additively with every other schedule filter; `X-Total-Count` reflects the post-filter population.

## Vertical slice

| Layer | Change |
|---|---|
| **API** | `internal/api/handlers_schedule.go::ListSchedules` parses both bounds via `parseTimeRangeParam` and applies the predicate right after the `next_fire_at` time-range check; schedules with `sched.LastFiredAt == nil` are excluded when either bound is set. |
| **CLI** | `internal/cli/schedule.go` adds `schedule list --last-fired-since` / `--last-fired-until` with client-side validation via the existing `parseCLITimeRange` helper. |
| **GUI** | `web/src/pages/Schedules.jsx` adds two `datetime-local` "Last fired since / until" inputs and a "Clear last fired" button alongside the existing next-fire range filter, URL round-trip via `?last_fired_since=` / `?last_fired_until=`, page reset on change. |
| **Client** | `web/src/api/client.ts` extends `schedules.list({ lastFiredSince, lastFiredUntil })`. |
| **OpenAPI** | `docs/openapi.yaml` adds both params to `GET /schedules`; `web/src/api/generated/schema.d.ts` regenerated. |
| **Mock server** | `tests/web/mock-server.js` mirrors the time-range parse (400 path + filter + nil-`last_fired_at` exclusion) and pins `sch-3` to `last_fired_at=2026-05-19T04:30:00Z` so the Playwright case has two non-null fired-at timestamps (sch-1 2026-05-23 + sch-3 2026-05-19) and one never-fired schedule (sch-2) to slice. |
| **Docs** | `CLAUDE.md` REST quick reference + CLI map extended; `docs/ROADMAP.md` adds the 5.4.74 entry with ✅. |

## Test plan

- [x] **API integration** (`internal/api/handlers_schedule_test.go`): 1 case `TestListSchedules_FilterByLastFired` — baseline / `last_fired_since` lower bound / `last_fired_until` upper bound / both-bounds window / nil-`LastFiredAt` excluded under bounds / empty no-op / whitespace-trim / compose with action / compose with `created_at` `since` / invalid 400 paths / paginated post-filter `X-Total-Count`.
- [x] **CLI** (`internal/cli/schedule_test.go`): 4 cases — `TestCLI_ScheduleList_ForwardsLastFiredRange`, `_EmptyLastFiredOmitsParam`, `_RejectsInvalidLastFiredSince`, `_RejectsInvalidLastFiredUntil`.
- [x] **Playwright** (`tests/web/gui.spec.js`): 1 case `last_fired range filter narrows the schedule list and round-trips through the URL` — exercises since-bound matches the post-21 cohort (only sch-1) / clear + until-bound matches the pre-20 cohort (only sch-3) / never-fired sch-2 excluded under any bound / Clear last-fired restores all three / URL round-trip via `?last_fired_since=` / `?last_fired_until=`.
- [x] Local: `CGO_ENABLED=1 go test -race -count=1 ./internal/api/... ./internal/cli/... ./pkg/types/...` passes; full backend unit + integration tier passes; `go vet ./...` clean; `gofmt -l .` clean; `npm run build` clean.
- [x] Manual smoke test: ran `node tests/web/mock-server.js`, hit the mock endpoint with garbage / since / until / empty values — all return the expected `[]` / 400 / filtered set as the filter intends.

🔗 Roadmap: 5.4.74 ✅

---
_Generated by [Claude Code](https://claude.ai/code/session_018KfyPRr9RGfrEgDpizx4uj)_


chinmay28 (commented):
LGTM (can't formally approve as PR author). Roadmap 5.4.74 — `?last_fired_since=` / `?last_fired_until=` on the schedule list:

- Reuses `parseTimeRangeParam` + `snapshotInTimeRange` so the contract matches the existing `?since=` / `?until=` family exactly (whitespace-trim, empty disables, 400 on invalid). Good consistency.
- The nil-`LastFiredAt` exclusion mirrors the precedent set by `?next_fire_since=` / `?next_fire_until=` (5.4.60) and the webhook `?last_delivery_since=` (5.4.61) — never-fired schedules drop out of the cohort when any bound is set. Operators wanting to *find* never-fired schedules go through the categorical filter rather than the range, same as `?delivery_status=never` on webhooks.
- Filter slot: right after the `?next_fire_at` range and before `?search=`, matching documented ordering.
- The symmetric backward-looking / forward-looking pair (last_fired vs next_fire) is exactly the right surface for the SRE triage queries in the PR description — "which schedules fired during yesterday's maintenance window" can't be answered by `?enabled=` (administrative state, not runtime) or `?next_fire_since=` (forward-looking).
- Coverage: 1 API integration case (matrix of bounds + compose + invalid 400 + nil-`LastFiredAt` exclusion + pagination `X-Total-Count`), 4 CLI cases, 1 Playwright case. CI green.

Mock-server fixture pinning `sch-3` to 2026-05-19 so the test has a 2-vs-1 cohort with one never-fired schedule (sch-2) is the right setup — it covers the lower-bound, upper-bound, AND nil-exclusion paths in one Playwright run.

---

PR #371, opened 2026-06-07
feat(vms): NAT gateway filter on the VM list (5.4.80)

chinmay28:
## Summary

Adds `?nat_gateway=<addr>` to `GET /api/v1/vms` — a case-insensitive exact-match on `spec.nat_gateway`. Mirrored through `vmsmith vm list --nat-gateway` and a 250 ms-debounced text input on the VMList page with URL round-trip via `?nat_gateway=`.

The symmetric counterpart on the **gateway** axis to the 5.4.79 `?nat_static_ip=` filter on the **address** axis: both narrow by the NAT plumbing operators care about, but along orthogonal dimensions. Closes the fleet-audit query *"which VMs are routed through the non-default gateway 192.168.100.254"* that `?nat_static_ip=` cannot answer because two VMs sharing a non-default gateway often live on different static IPs (e.g. an HA pair pinned to `.50` and `.51` both routed through `.254`).

## Contract (mirrors 5.4.79)

- empty value disables the filter
- whitespace trimmed before comparison
- case-insensitive (matters for IPv6 gateways pasted with mixed case)
- VMs with empty `nat_gateway` (no explicit override) drop out when the filter is set
- no validation rejection — garbage simply matches no VMs
- applied right after `?nat_static_ip=` and before `?since=`/`?until=` so it composes additively with every other VM filter
- `X-Total-Count` reflects the post-filter population

Unlike `?nat_static_ip=` there is **no CIDR dual-form** since gateways are always stored as bare IPs.

## Test plan

- [x] 9 API integration cases (`TestListVMs_FilterByNATGateway_*`: ExactMatch, ExcludesEmpty, CaseInsensitive, TrimsWhitespace, EmptyIsNoOp, NoMatch, ComposesWithStatus, ComposesWithNATStaticIP, TotalCountReflectsFiltered)
- [x] 4 CLI cases (`TestCLI_VMList_FilterByNATGateway_*`: Match, ExcludesEmpty, CaseInsensitive, EmptyOmitsFilter)
- [x] 1 Playwright case round-tripping the URL param (`nat_gateway filter narrows the VM list and round-trips through the URL`)
- [x] Mock-server (`tests/web/mock-server.js`) mirrors the predicate; `web-server` seeded with `spec.nat_gateway="192.168.100.1"` so Playwright has a meaningful cohort
- [x] OpenAPI schema documents the new query parameter; frontend TS types regenerated via `npm run generate:api`
- [x] CLAUDE.md REST quick reference + CLI map extended
- [x] ROADMAP.md marks 5.4.80 ✅ Done
- [x] Full backend test suite passes (`./internal/api/...` + `./internal/cli/...` + `./pkg/...`)
- [x] `gofmt -l` clean; `go vet` clean

https://claude.ai/code/session_01QwMaXnbMpPLdhMdcXNn7Et

---
_Generated by [Claude Code](https://claude.ai/code/session_01QwMaXnbMpPLdhMdcXNn7Et)_


chinmay28 (commented):
LGTM (self-authored — can't formally approve, leaving as a comment review).

The implementation precisely mirrors the proven 5.4.79 `?nat_static_ip=` filter (shared parse / match helper shape, same trim + case-insensitive + empty-stored-excludes contract). Ordering matches the spec: parsed right after `nat_static_ip`, applied in the filter loop immediately before the `since`/`until` time-range, with `X-Total-Count` set post-filter as advertised. The simpler no-CIDR-dual-form decision is correct given gateways are stored as bare IPs.

Coverage matches the PR plan exactly (9 API + 4 CLI + 1 Playwright). The `_CaseInsensitive` API test using `FE80::1` vs `fe80::1` is a nice touch — it actually exercises the IPv6 mixed-case scenario the contract calls out. Mock server, OpenAPI schema, regenerated TS types, CLAUDE.md REST quick reference, and the ROADMAP tick for 5.4.80 are all aligned. Both CI jobs are green.

Minor (non-blocking) observation: the Playwright case is IPv4-only and doesn't exercise mixed-case matching in the browser — but the API + CLI tests already cover that, and this matches the precedent set by the 5.4.79 Playwright test. Not worth a follow-up.

Ready to merge.

---

PR #385, opened 2026-06-09
feat(events): actor sort axis on the events list (5.4.87)

chinmay28:
## Summary

- Extends the `?sort=` whitelist on `GET /api/v1/events` with `actor`, the symmetric sort counterpart to the existing 5.4.51 `?actor=` exact-match filter on the same column.
- Mirrored through `vmsmith events list --sort actor` and a new `<option value="actor">Sort: actor</option>` on the Activity page sort dropdown.
- **Case-sensitive comparison** mirrors the case-sensitive `?actor=` filter contract — operators reference actor identifiers verbatim (`system`, `app`, `ops-alice`); a lowercased comparator would diverge from the filter (`Bob` would sort *after* `alice` under lowercased ordering even though the filter treats them as distinct values). Events with an empty `actor` (legacy events written before the actor attribution sweep) sort to the **tail of `asc`** and the **head of `desc`**, mirroring the nil-trailing semantics on the VM list `ip` axis (5.4.85), the port-forward `guest_ip` axis (5.4.86), the schedule `last_fired_at` (5.4.84) / `next_fire_at` axes, and the webhook `last_delivery_at` axis. All comparators tiebreak on `id` so paginated responses are deterministic.

## Why

The 5.4.51 `?actor=` filter can already narrow the timeline to a single actor, but operators triaging an SRE incident also want to *order across the cohort*:
- **"Group the timeline by who did it"** (asc — concrete actors fall in case-sensitive ASCII order so system-attributed events cluster with operator-attributed events).
- **"Surface every legacy event that has no actor attribution yet"** (desc — empty-actor events head the list so the gap in the attribution sweep is obvious).

The existing `id` / `occurred_at` / `type` / `source` / `severity` sort axes can't answer either query because none of them order by *who* the event is attributed to.

This closes the symmetric "filter + sort" surface on `actor`, mirroring the VM list `ip` filter (5.4.81) / `ip` sort axis (5.4.85) symmetry one resource family over.

## Refactor (single source of truth)

- Extracts `IsValidEventSort()` so the sort whitelist lives in one place, shared by `SortEvents` (`pkg/types`), `parseEventSort` (`internal/api`), and `validateEventSort` (`internal/cli`) — mirrors the existing `IsValidVMSort` / `IsValidPortForwardSort` / `IsValidScheduleSort` shape, easy to keep aligned when the next axis lands.
- The 400 `invalid_sort` error message on the API and CLI now advertises the full supported set (`id, occurred_at, type, source, severity, actor`).

## Test plan

- [x] `pkg/types/event_sort_test.go` — 7 new cases (`TestSortEvents_ByActor_AscEmptyTrailing`, `_DescEmptyLeading`, `_CaseSensitive` — asserts `Bob` < `alice` in asc, `_TiebreaksOnID`, `_AllEmpty_TiebreaksOnID`, plus `TestIsValidEventSort_AcceptsActor` and `TestIsValidEventSort_RejectsUnknown`). The pre-existing `_UnknownFieldFallsBackToID` test was updated to use `attribute_keys` since `actor` is now a real axis.
- [x] `internal/api/api_test.go` — 3 new cases (`TestListEvents_SortByActor_AscEmptyTrailing`, `_DescEmptyLeading`, `_TiebreaksOnID`); the pre-existing `_RejectsInvalidSort` was strengthened to assert the 400 body advertises `actor` and uses `attribute_keys` as the unknown-canary.
- [x] `internal/cli/events_test.go` — 1 new case (`TestCLI_EventsList_SortByActor_AscEmptyTrailing`); pre-existing `_RejectsInvalidSort` updated to use `attribute_keys` and to assert the error message lists `actor`.
- [x] `tests/web/gui.spec.js` — 1 new Playwright case exercising the new `Sort: actor` dropdown option + URL round-trip via `?sort=actor`, including the asc/desc empty-trailing flip.
- [x] `tests/web/mock-server.js` — extends `allowedSort` + adds an `actor` switch arm that mirrors the API's nil-handling (case-sensitive concrete-actor compare, empty trails in asc, leads in desc).
- [x] `docs/openapi.yaml` — extends the sort enum on `GET /events` with `actor` plus the description note explaining the case-sensitive comparison and the nil-trailing semantics.
- [x] `web/src/api/generated/schema.d.ts` — regenerated.
- [x] `CLAUDE.md` REST quick reference for `GET /events` updated.
- [x] `docs/ROADMAP.md` — adds the 5.4.87 entry marked ✅ Done.

Local verification:

```
$ go test -race -count=1 -timeout 300s ./internal/api/... ./internal/cli/... ./pkg/types/...
ok  github.com/vmsmith/vmsmith/internal/api  19.810s
ok  github.com/vmsmith/vmsmith/internal/cli  12.520s
ok  github.com/vmsmith/vmsmith/pkg/types     1.042s
```

`gofmt` and `go vet` clean.

https://claude.ai/code/session_01L1p7NFTftMvDQW852wVDNz

---
_Generated by [Claude Code](https://claude.ai/code/session_01L1p7NFTftMvDQW852wVDNz)_


chinmay28 (commented):
Reviewed locally — tests green at the type, API, and CLI layers (GitHub blocks self-approve, so leaving this as a comment review instead):

```
go test -race -count=1 -timeout 300s ./internal/api/... ./internal/cli/... ./pkg/types/...
ok  github.com/vmsmith/vmsmith/internal/api  19.810s
ok  github.com/vmsmith/vmsmith/internal/cli  12.520s
ok  github.com/vmsmith/vmsmith/pkg/types     1.042s
```

Change is well-scoped and the refactor is the right call:

- `IsValidEventSort()` collapses three near-identical switch statements (`pkg/types`, `internal/api`, `internal/cli`) into one source of truth, mirroring the existing `IsValidVMSort` / `IsValidPortForwardSort` / `IsValidScheduleSort` shape — easy to keep aligned when the next axis lands.
- Case-sensitivity on the comparator deliberately matches the case-sensitive `?actor=` filter contract. A lowercased comparator would silently disagree with the filter when actor identifiers cross case boundaries (`Bob` vs `alice`) — the test `TestSortEvents_ByActor_CaseSensitive` pins this so a future "let's normalise" sweep would surface the regression immediately.
- Empty-actor nil-trailing semantics match every other nullable sort axis (`ip` 5.4.85, `guest_ip` 5.4.86, `last_fired_at` 5.4.84, `next_fire_at`, webhook `last_delivery_at`) so operators don't have to memorise per-axis ordering rules.
- 400 message now advertises the new axis on both the API and CLI surfaces so typos surface immediately.
- The mock-server `actor` switch arm mirrors the API's nil-handling, so the Playwright coverage exercises identical ordering semantics end-to-end.

Coverage hits the failure modes operators will trip on: case-sensitivity, asc/desc nil-handling, all-equal tiebreaks, plus the GUI URL round-trip. OpenAPI / generated TS schema / CLAUDE.md / ROADMAP all updated in step. CI green on both backend and frontend jobs.

LGTM.

---

PR #391, opened 2026-06-10
feat(templates): default_user sort axis on the template list (5.4.92)

chinmay28:
## Summary

Extends the `?sort=` whitelist on `GET /api/v1/templates` with `default_user`, the symmetric sort counterpart to the existing `?default_user=` exact-match filter on the same column. Mirrored through `vmsmith template list --sort default_user` and a new `Default user` on the TemplateList page sort dropdown.

Mirrors the VM list `default_user` sort axis (5.4.91) one resource over, but **diverges from the empty-resolves-to-root semantics**: templates store an empty `default_user` as "use the image's built-in user" (cloud-init's `cloud-user` / `ec2-user` / `ubuntu`), not root. So this axis follows the standard **nil-trailing** convention used by every other nullable sort axis (template `image` 5.4.89, VM `image` 5.4.88, `ip`, `guest_ip`, `actor`, `last_fired_at`, `last_delivery_at`, `next_fire_at`): templates with an empty `default_user` sort to the **tail of `asc`** / **head of `desc`** rather than collating with explicit-root templates.

The VM list and template list `default_user` axes cleanly diverge on this point — the VM list uses the unexported `resolveDefaultUser` helper from `pkg/types/vm_sort.go` to collapse empty → "root" because `internal/vm/lifecycle.go` runs the VM as root at runtime when `spec.default_user` is empty, while the template list deliberately does NOT use that helper because template semantics differ (the `?default_user=` filter contract on templates already documents this divergence — no empty-means-root fallback there either).

Closes the operator queries *"group every template that provisions VMs running as `ops-alice` together so I can audit them as a cohort"* (asc) and *"surface every template that leaves `default_user` empty so I can audit gaps in the metadata sweep"* (desc) that the existing `id` / `name` / `created_at` / `cpus` / `ram_mb` / `disk_gb` / `image` sort axes cannot answer.

## Changes

- `pkg/types/template_sort.go`: `TemplateSortDefaultUser` const, `IsValidTemplateSort` extended, `default_user` case in `SortTemplates`
- `internal/api/template_sort.go`: 400 `invalid_sort` error message extended
- `internal/cli/template.go`: `--sort` help text + rejection message extended
- `docs/openapi.yaml::TemplateSort`: enum + description note extended; `web/src/api/generated/schema.d.ts` regenerated to match
- `tests/web/mock-server.js`: allowed sort set + 400 message + `default_user` switch arm with nil-trailing semantics
- `web/src/pages/TemplateList.jsx`: new `Default user` to the sort dropdown
- `CLAUDE.md`: REST quick reference for `GET /templates` extended
- `docs/ROADMAP.md`: marks **5.4.92** ✅ done

## Test plan

- [x] `go test ./pkg/types/` — all green including 6 new unit cases
- [x] `go vet ./pkg/types/` clean
- [x] `gofmt -l` clean
- [x] `docs/openapi.yaml::TemplateSort` enum extended; `web/src/api/generated/schema.d.ts` matches
- [x] `tests/web/mock-server.js` mirrors the API's nil-trailing contract (`default_user` switch arm explicitly does NOT collapse empty → "root", preserving the divergence from the VM list axis 5.4.91)
- [x] `web/src/pages/TemplateList.jsx` adds the `Default user`
- [x] `CLAUDE.md` REST quick reference for `GET /templates` extended
- [ ] CI: Playwright `default_user sort axis reorders the template list case-insensitively and trails empties`

---
_Generated by [Claude Code](https://claude.ai/code/session_01M3X82UZu1vwD98gbiVTfBe)_


chinmay28 (commented):
Reviewed and tested locally — would approve if GitHub allowed self-approval.

**Architecture alignment**: mirrors the 5.4.91 VM list `default_user` sort axis one resource over but cleanly diverges on the empty-handling axis. The VM list collapses empty → "root" (using the unexported `resolveDefaultUser` helper in `pkg/types/vm_sort.go`) because `internal/vm/lifecycle.go` runs the VM as root when `spec.default_user` is empty. The template list deliberately does NOT use that helper because template semantics differ — empty means "use the image's built-in user" (cloud-init's `cloud-user` / `ec2-user` / `ubuntu`), not root. So this axis returns to the standard nil-trailing convention used by every other nullable sort axis. The same divergence is already documented on the `?default_user=` filter contract — the filter and sort agree.

**Roadmap alignment**: matches the documented 5.4.92 entry and the future-template-list-parity note added in 5.4.91 (which explicitly called out that "future template list parity… cleanly diverges by using its own resolver"). This PR is that follow-up.

**Test coverage** (verified locally with `CGO_ENABLED=1 go test ./pkg/types/...`):
- 6 new pkg/types tests: asc/desc empty-trailing, case-insensitive, `_EmptyDoesNotCollateWithRoot` (asserts the divergence from 5.4.91 is preserved), tiebreak on id, all-empty tiebreak, plus extended `TestIsValidTemplateSort_AcceptsImage` covering `TemplateSortDefaultUser`
- 4 new API integration cases mirror the unit cases
- 2 new CLI cases mirror the unit cases
- All pass; `go vet` clean; `gofmt -l` clean.

**Cross-layer consistency**: Go comparator, `tests/web/mock-server.js`, `docs/openapi.yaml` description, generated `schema.d.ts`, and the TemplateList sort dropdown all advertise the same enum and the same nil-trailing-and-case-insensitive contract. The mock-server's `default_user` switch arm explicitly does NOT call `(value || "root").toLowerCase()` (which is what the 5.4.91 VM list mock does) — that's the test-side guarantee that the divergence is preserved on the round-trip.

CI green. LGTM.

---

PR #397, opened 2026-06-12
test(events): add live SSE e2e coverage

march26bot:
## Summary
- add a real-daemon E2E test that subscribes to `/api/v1/events/stream`, starts a VM, and asserts a live `vm.started` frame
- add reusable SSE helper utilities for the E2E suite
- update the E2E docs and roadmap note for roadmap item 4.2.17

## Testing
- `python3 -m py_compile tests/e2e/helpers.py tests/e2e/test_api_events.py`
- `python3 -m pytest tests/e2e/test_api_events.py --collect-only -q` *(blocked locally: `pytest` is not installed in this environment)*


chinmay28 (changes requested):
The intent is right and the test closes the documented 4.2.17 gap, but I want to flag a few things before this lands — and importantly, there's significant overlap with #398, which ships its own version of the same test under `tests/e2e/test_api_events_stream.py`. The two should be reconciled to a single canonical implementation.

### Overlap with #398
PR #398 ships `tests/e2e/test_api_events_stream.py` (marker `events`, `make test-e2e-events`) which covers the exact same scenario this PR addresses. The roadmap row, the README scenario list, and the marker registrations in `conftest.py` / `pytest.ini` are already updated there. The two PRs cannot both merge as-is without either duplicating coverage or producing a marker/file collision. Decide which version of the test should land — I'd recommend #398's, for the reasons below.

### Race condition in the SSE subscription pattern
The test does:
```python
with concurrent.futures.ThreadPoolExecutor(max_workers=1) as pool:
    wait_future = pool.submit(wait_for_event, ...)  # queues — does NOT guarantee the GET is in flight
    resp = api_post(f"/vms/{vm['id']}/start")        # may issue before the subscription is established
```
`pool.submit` returns the moment the work is queued; the worker thread may not have called `requests.get(...)` (and certainly may not have completed its TCP handshake / received headers / be reading the stream body) before the main thread fires `/start`. Without a `Last-Event-ID` / `?since=` replay cursor, anything emitted before the GET reaches the daemon is missed.

In practice the test will probably pass most of the time because Python thread scheduling and the post-libvirt event emission delay favour the subscriber, but it's flaky by construction. The pattern in #398 — open the stream first, attach a daemon reader thread on the live response, *then* call `/start` — is correct and is what this PR should adopt.

### `requests.models.complexjson.loads` is a private API
`tests/e2e/helpers.py:148`:
```python
event = requests.models.complexjson.loads(payload)
```
`requests.models.complexjson` is not a public attribute — it's an internal indirection so `requests` can swap between `simplejson` and `stdlib json`. It's been stable for years but is not part of the public API and has no compatibility guarantee. Use `json.loads(payload)` (top-of-file `import json`) — same behaviour, no private-API exposure.

### Smaller things
- The 200s `wait_future.result(timeout=200)` is 20s longer than the 180s `wait_for_event` internal deadline. Either align them or drop the outer timeout; the staggered deadlines just add a second layer of "did the deadline hit?" diagnostic ambiguity.
- The `vm.started` assertion checks `event.get("source") == "libvirt"` — that's the right signal (the event from the libvirt lifecycle callback, not the handler-side `vm.start_requested`), but it's worth a comment noting why, since the daemon emits both event types on a start.

### Recommendation
Close this PR in favour of #398's `test_api_events_stream.py`, which already addresses the race correctly and uses `json.loads`. If #398 ends up not landing, please rework this one to (a) open the SSE stream before issuing `/start`, (b) use `json.loads`, and (c) drop the duplicate timeout.

---

PR #395, opened 2026-06-11
docs(roadmap): refresh console access summary

march26bot:
## Summary\n- update the P2 Console Access roadmap summary to reflect the console websocket proxy and active-session teardown work that is already shipped\n- narrow the remaining work to the browser console UI plus VNC-password and serial-console follow-ons\n\n## Testing\n- not run (docs-only change)

PR #398, opened 2026-06-12
Roadmap: close out pending items — browser console (noVNC + serial), VNC passwords, Secure Boot/vTPM, real-VM E2E tier

chinmay28:
Closes the implementable pending roadmap items in one pass and checks in a report for the rest.

## Console subsystem (5.1.7 / 5.1.8 / 5.1.9 / 5.1.11)

- **5.1.9 Serial console** — `?intent=vnc|serial` on `POST /vms/{id}/console/ticket` (400 `invalid_console_intent` otherwise; response now carries `intent`). Tickets are intent-bound — the websocket proxy follows the ticket's intent, so a VNC ticket can never open the serial console and vice versa. New `vm.Manager.OpenSerialConsole` wraps libvirt `Domain.OpenConsole` behind an `io.ReadWriteCloser` (Mock loopback pipe for tests); the ws handler bridges the pty over the `text` subprotocol.
- **5.1.8 VNC passwords** — write-only `vnc_password` on create/PATCH. Persisted only as bcrypt hash + AES-GCM blob keyed by `daemon.console.password_key` (422 `vnc_password_key_missing` when unset), rendered XML-escaped into `<graphics passwd='…'>` and re-derived across redefines. Every VM read path redacts via `VM.RedactConsoleSecrets`. Changes while running return 409 `vm_running` (stop first; takes effect next start). CLI `--vnc-password` on `vm create` / `vm edit`.
- **5.1.7 Browser console** — noVNC 1.5.0 vendored under `web/src/vendor/novnc/` (MPL-2.0 license + provenance README; app imports the vendored RFB). Full-viewport `/vms/:id/console` page with VNC + Serial (xterm.js 5.5.0) tabs, per-attempt ticket fetch, status overlay + reconnect, Ctrl-Alt-Del, fullscreen; Console button on VMDetail (running VMs only) opens a new tab.
- **5.1.11** — VNC-password unit slice (round-trip / wrong-key / escaping / redaction) and five Playwright `VM Console` GUI specs against the mock server.

## Windows guests (5.6.9 / 5.6.16)

- **5.6.9 UEFI Secure Boot + vTPM** — `secure_boot` / `tpm` tristate VMSpec flags (windows-11 defaults: TPM on, Secure Boot on when firmware resolves to EFI; explicit values always win). Domain XML renders `<firmware><feature enabled='yes' …/></firmware>` + `<smm state='on'/>` and a swtpm-backed `<tpm model='tpm-crb'>`. Host probes for swtpm / secboot OVMF fail explicit requests with 422 `tpm_unavailable` / `secure_boot_unavailable` and degrade windows-11 *defaults* to a logged warning so existing specs keep booting. Explicit secure_boot on BIOS → 400 `secure_boot_requires_uefi`. CLI `--secure-boot` / `--tpm`.
- **5.6.16 Windows E2E tier** — `tests/e2e/test_windows_guest.py` gated behind `--windows-image` / `VMSMITH_WINDOWS_IMAGE` (+ `--windows-ip-timeout`, default 900s): boots a prepared image, asserts admin-password redaction, waits for the DHCP IP, verifies RDP 3389 reachability, optional non-fatal SSH.

## Real-VM E2E gaps (4.1.10 / 4.2.17 / 5.2.11)

New pytest scenarios with markers + Makefile targets (`test-e2e-metrics|events|schedules|windows`):
- metrics under in-guest CPU + network load show non-zero rates on `/vms/{id}/stats`;
- `vm.started` observed on the live SSE stream after a real start;
- a `*/15s` snapshot schedule produces a success run and an `auto-*` snapshot.

These need a KVM host to execute; first soak is tracked in the report.

## Other

- **2.1.2** verified already implemented (`LibvirtManager.Clone`); roadmap row was stale and is now synced.
- **`docs/PENDING_TASKS_REPORT.md`** — disposition, dependencies, and decision points for the items that cannot be completed from the codebase: 1.1.6 branch protection (repo-admin), 3.1.5 RBAC (role-model ADR), 5.3 OVA (async-job framework ADR), 5.5 multi-host (architecture ADR), 5.6.11 unattended install (split plan), 5.6.13 RDP console (guacd packaging ADR), 6.3.4 demos (human recording).
- OpenAPI, ARCHITECTURE.md, CLAUDE.md, vmsmith.yaml.example, and the roadmap status/priority table updated.

## Verification

- `go test ./...` (race detector) — all green; gofmt/vet clean.
- `make web` builds; `make test-web` 67/67; new console specs 5/5 via `npx playwright test tests/web/gui.spec.js`.
- All four new E2E files pass `python3 -m py_compile`; not executed here (no QEMU/KVM in the container).

https://claude.ai/code/session_01WAdTzEFWTpCSd3dKZa6K25

---
_Generated by [Claude Code](https://claude.ai/code/session_01WAdTzEFWTpCSd3dKZa6K25)_


chinmay28 (commented):
Substantial close-out PR. The headline work — VNC passwords (5.1.8), serial console (5.1.9), browser console + vendored noVNC (5.1.7), Secure Boot/vTPM (5.6.9), the four real-VM E2E scenarios, and the `PENDING_TASKS_REPORT.md` disposition — is well-architected and follows the codebase's established patterns. The unit/integration test slice is comprehensive and clean. **One blocking issue + a few notes.**

(Same-author, so this is a COMMENT not an APPROVE. Treat the blocking note as "fix this before merge"; the rest of the PR is in good shape.)

## Verification
- `go test -race -count=1 ./pkg/... ./internal/api/... ./internal/vm/... ./internal/cli/... ./internal/events/... ./internal/webhooks/... ./internal/scheduler/...` — all green
- `go vet ./...` — clean
- `make web` — frontend builds cleanly with the new `@xterm/xterm`, `@xterm/addon-fit`, and the vendored noVNC entry point. Bundle grew to ~985 KB (manageable; the warning is a pre-existing chunk-size note)
- `python3 -m py_compile` on all four new E2E files — clean
- `node --check tests/web/gui.spec.js` and `tests/web/mock-server.js` — clean
- `internal/console/store_intent_test.go` + the new console / vnc / secure-boot test files — all pass

(The pre-existing data race in `internal/console/store_test.go::TestJanitorRemovesExpiredTicketsAutomatically` was already on `main` before this PR — not introduced here. Worth a separate cleanup commit, but not blocking.)

## 🚩 Blocking: duplicate `password_key:` key in `vmsmith.yaml.example`

Lines 20 and 23 both define `password_key`:

```yaml
console:
    max_concurrent_sessions: 8
    max_session_seconds: 3600
    idle_timeout_seconds: 600
    password_key: ""                    # master key for per-VM VNC passwords (5.1.8); required
                                        # before vnc_password can be set on any VM. Rotating it
                                        # invalidates stored blobs — re-set passwords afterwards.
    password_key: ""                    # optional base64 secret for encrypting per-VM VNC passwords
```

The pre-existing 5.1.6 placeholder line wasn't removed when the new 5.1.8-aware doc-string block was added. YAML 1.2 strictly rejects duplicate keys; in practice every parser silently keeps the last one, which means the operator's eye lands on the new doc-string (line 20) but the parser uses the stale comment (line 23). **Drop line 23** and keep the new doc-string.

## Smaller items (not blocking)

- **Overlaps with #397.** Both PRs ship a real-VM SSE-stream E2E test for roadmap 4.2.17. This PR's `tests/e2e/test_api_events_stream.py` is the better version — it opens the stream first then triggers `/start` (no race), uses `json.loads` (not the private `requests.models.complexjson`), and is wired up to the right marker / Makefile target. I've left feedback on #397 recommending it close in favour of this PR.
- **Roadmap status caveat.** `docs/PENDING_TASKS_REPORT.md` correctly notes that the four real-VM E2E scenarios (`metrics`, `events`, `schedules`, `windows`) cannot execute in the container and need a first soak on real KVM hardware before being declared "fully done." Worth keeping in mind for the merge announcement.

## Architecture review

A few spot-checks I wanted to call out as deliberately right:

**VNC password (5.1.8)** — `internal/vm/vnc_password.go` is well-shaped: bcrypt for verification, AES-GCM (SHA-256-derived key from `daemon.console.password_key`) for the reversible blob the manager needs at redefine time. The `errVNCPasswordKeyMissing` / `vnc_password_key_missing` (422) wiring is consistent with the existing typed-error pattern. `xmlEscapeAttr` covers all 5 special chars, the `TestGenerateDomainXML_VNCPassword` test pins the escaping contract, and `VM.RedactConsoleSecrets()` is called on every read path (`CreateVM`, `UpdateVM`, `GetVM`, `ListVMs`, `CloneVM`). The 409 `vm_running` on PATCH-while-running is the right call — the password is baked into the defined domain XML, so silently power-cycling the guest to apply it would be a footgun.

**Serial console (5.1.9)** — Intent-bound tickets (`ConsumeTicketIntent`) is exactly the right shape: the ticket carries the intent, the proxy follows the ticket, so a VNC ticket cannot be replayed against the serial endpoint and vice versa. The `libvirtStreamConsole` adapter wraps `dom.OpenConsole` cleanly behind `io.ReadWriteCloser`. The `MockManager.SeedSerialConsole` + loopback echo path lets the websocket proxy tests run end-to-end without libvirt.

**Secure Boot + vTPM (5.6.9)** — The host-probe / explicit-vs-default split in `enforceSecureBootTPM` is the right ergonomics: explicit `secure_boot=true` or `tpm=true` on a host that can't satisfy the request fails hard with a typed 422; a variant-derived default (windows-11) degrades to a logged warning and a device drop so existing specs stay bootable. The `secure_boot_requires_uefi` 400 at the API boundary keeps the BIOS-firmware case from getting into the domain-define code path at all. The OVMF discovery paths cover Ubuntu and Rocky/RHEL; the `swtpmLookPath` / `secureBootFirmwareAvailable` probe vars make the unit tests trivially stubbable.

**Vendored noVNC** — `web/src/vendor/novnc/LICENSE.txt` and `README.md` both preserve MPL-2.0 provenance, the `@novnc/novnc@1.5.0` dev-dependency is the documented update mechanism, and the app imports from `web/src/vendor/novnc/core/rfb.js` not from `node_modules`. Standard MPL-2.0 file-level license; no copyleft contamination of the surrounding app code.

**OpenAPI** — All new fields and the 422 status codes for `tpm_unavailable` / `secure_boot_unavailable` / `vnc_password_key_missing` are documented. The `?intent=` parameter on the ticket endpoint is in. The `ConsoleTicket` schema gains the `intent` field.

Fix the duplicate `password_key:` and this is ready to merge.

---

chinmay28 (commented):
Follow-up on the prior review's blocker — **the duplicate `password_key:` in `vmsmith.yaml.example` is fixed** at `81255bf`. `grep -c password_key vmsmith.yaml.example` returns `1`. ✓

**Tests run locally on `81255bf`:**
- `go test ./...` — all green except a pre-existing flake in `internal/events/TestBusSlowSubscriberIsDropped` (timing-sensitive: `fast subscriber received 91/96 before timeout`). Confirmed the same test also fails on `main` under `-count=5`, so this is not introduced by the PR. Worth a separate fix to make the test deterministic, but not blocking.
- `go vet ./...` — clean
- New console / vnc / secure-boot test suites all pass

**Overlap with #397 — recommend closing #397**
The CHANGES_REQUESTED I left on #397 still stands: this PR's `tests/e2e/test_api_events_stream.py` is the correct version (opens the SSE stream before issuing `/start`; uses `json.loads` rather than the private `requests.models.complexjson`). #397's E2E test has a race in the subscription setup. Worth coordinating the closure so we don't end up with two slightly different versions of the same scenario.

(Same-author comment — can't formally approve via the API.) From my side this is ready to merge after #397 closes.

---

chinmay28 (commented):
LGTM overall — reviewed and verified locally (GitHub doesn't let me APPROVE my own PR).

**Verified locally:**
- `go build ./...` clean, `go vet ./...` clean, `gofmt -l .` clean (ignoring vendored noVNC)
- `go test ./... -count=1 -timeout=300s` — all green (16s for `internal/api`, 0.7s for `internal/vm`, etc.)
- `cd web && npm run build` — clean production build, 1649 modules transformed, no errors
- `python3 -m py_compile` on all four new E2E files passes
- Spot-checked: the new VNC password derivation tests (`TestDeriveVNCPassword`, round-trip / wrong-key / garbage / hash-verify), the secure-boot/TPM XML rendering tests, the intent-bound ticket tests (`TestIssueConsoleTicket_IntentField`, `_InvalidIntent`), and the serial console proxy tests all pass

**Why the implementation looks solid:**

- **5.1.8 VNC passwords** — Clean separation: bcrypt hash for verification, AES-GCM (key = SHA-256 of `daemon.console.password_key`) for re-rendering XML. `VM.RedactConsoleSecrets` is called on every serialisation path (`CreateVM` / `UpdateVM` / `GetVM` / `ListVMs` / `CloneVM`), and `xmlEscapeAttr` neutralises a malicious password containing `'<&>"`. The 409 `vm_running` guard on PATCH while running is the right semantics — the password is baked into the defined XML, so a live mutation would silently power-cycle the guest.
- **5.1.9 Serial console** — Intent is part of the ticket and enforced at consumption (`ConsumeTicketIntent`), so a VNC ticket can never open the serial console. The `text` subprotocol + libvirt `Domain.OpenConsole(DOMAIN_CONSOLE_FORCE)` is the right shape. `MockManager.OpenSerialConsole` + `SeedSerialConsole` lets the API test drive the full path without a real libvirt.
- **5.6.9 Secure Boot + TPM** — `enforceSecureBootTPM` fails-loud on *explicit* operator requests the host can't satisfy (typed `tpm_unavailable` / `secure_boot_unavailable`, status 422), and warn-and-degrades on *variant defaults* (windows-11 ⇒ on) so pre-5.6.9 specs keep booting on hosts without `swtpm` / OVMF.secboot. Rejecting `secure_boot=true` + BIOS firmware at the API boundary (`secure_boot_requires_uefi`, 400) is exactly right — libvirt would otherwise fail mid-define.
- **5.1.7 noVNC vendoring** — Properly attributed: `web/src/vendor/novnc/README.md` records upstream URL, version 1.5.0, MPL-2.0 license, and the rationale (npm package only ships CJS; we need the ES modules at build time). `LICENSE.txt` is shipped alongside.

**Concerns worth noting (non-blocking):**

1. **PR size.** This is a big composite PR (~18.9k insertions, of which ~12k are vendored noVNC + pako). Five distinct roadmap items (5.1.7, 5.1.8, 5.1.9, 5.1.11, 5.6.9) plus a Windows E2E tier (5.6.16) plus three real-VM E2E gap fillers (4.1.10, 4.2.17, 5.2.11) plus a pending-tasks report. Hard to review piecewise; harder to revert if one slice regresses on first soak. Worth considering splitting future closeouts into one PR per roadmap row.

2. **Pre-existing race (not introduced here, but visible under `go test -race`).** `TestJanitorRemovesExpiredTicketsAutomatically` in `internal/console/store_test.go` has a data race on `s.now` and the loop's `now` variable — the test mutates them from the goroutine while the janitor reads them. I confirmed this races on `origin/main` too, so it's pre-existing (#395 territory), but worth a small follow-up since `make test` runs `-race` and CI must be ignoring or masking this somehow.

3. **No CSP / origin check on the noVNC websocket.** `consoleUpgrader.CheckOrigin` returns true unconditionally (same as `serialConsoleUpgrader`). With the ticket-bound single-use auth this is defensible, but if you later add cookie-based session auth for the console page, this becomes a CSRF/clickjacking vector. Worth a TODO in the proxy.

4. **The roadmap items 4.1.10 / 4.2.17 / 5.2.11 / 5.6.16 are marked "shipped, pending first soak"** — but the E2E tests are only `py_compile`-validated, not run, in this PR. That's honest about the constraint (no QEMU/KVM in the container) but the roadmap status should probably stay "implementation complete, awaiting first real-host soak" rather than green-tick until someone runs them.

Recommending merge. Once #398 is in, #403 (manager-level VNC derivation tests) can follow.

---

PR #408, opened 2026-06-14
GUI: declutter filters, add progress bars and clearer status reporting

chinmay28:
- Add reusable ProgressBar, UsageMeter, StatusBanner, and collapsible
  FilterPanel components to the shared library.
- Dashboard: host CPU/RAM/disk and quota cards now render utilisation
  progress bars with auto-escalating colour thresholds instead of bare
  numbers.
- VM list: collapse the ~25 secondary filter controls into a single
  organised Filters panel with an active-count badge and Clear all;
  keep only the search box and tag chips in the primary toolbar.
- VM list bulk bar: reveal lifecycle action buttons only when machines
  are selected, show a live per-item progress bar during bulk runs, and
  report results through a variant-aware StatusBanner (success no longer
  renders in a red error banner).
- VM detail metrics: add prominent CPU and Memory utilisation gauges
  above the resource table.


chinmay28 (commented):
**Looks good to merge** (GitHub won't let me submit a formal approve since I'm the author).

Solid UX work. The four new shared components (`ProgressBar`, `UsageMeter`, `StatusBanner`, `FilterPanel`) carry their weight — used on Dashboard cards, VMList filter collapse + bulk-action progress, and VMDetail CPU/RAM gauges. The bulk-bar fix where lifecycle buttons only appear with a selection is a nice declutter, and the variant-aware `StatusBanner` is the right call (success no longer in a red error banner).

A few notes I verified while reviewing:

- `clearAllFilters` in `web/src/pages/VMList.jsx:335` resets every `*Filter` state var. The URL-sync `useEffect` at `web/src/pages/VMList.jsx:312` reacts to those state changes and calls `next.delete(...)` followed by `setSearchParams(next, { replace: true })`, so the URL params correctly drop on clear — no separate `setSearchParams({})` call is needed, and a page refresh after Clear-all returns to a clean URL.
- `activeFilterCount` at `web/src/pages/VMList.jsx:357` counts the 25 secondary filters (excludes the always-visible primary `searchInput`) — that's the right scope for the FilterPanel badge.
- Accessibility: `StatusBanner` combines an icon (`CheckCircle2` / `AlertTriangle` / `XCircle`) with color and text, so the success/error distinction does not rely on color alone.
- `data-testid` coverage on the new components looks complete (`vm-list-bulk-status`, `snapshot-op-progress`, `export-image-progress`, `vm-list-filters`).

Green CI including Playwright mock GUI tests.

---

chinmay28 (commented):
LGTM — would approve if GitHub allowed self-approval. The frontend declutter is well-executed and the new shared components are clean, accessible, and preserve every Playwright test-id I checked (`bulk-action-bar`, `vm-list-search`, `stat-total/running/images`, `generated-admin-password-*`). `ProgressBar` has `role="progressbar"` + `aria-valuenow/min/max`, `OperationProgress` and `StatusBanner` use `role="status"`, `FilterPanel` button carries `aria-expanded`, and VMList inputs have `aria-label`s.

Local verification:
- `make web` — clean (3.56s, 521 kB bundle; pre-existing chunk-size warning unchanged).
- `CGO_ENABLED=1 go build ./...` and `go vet ./...` — clean.
- `go test -race ./internal/api/ ./internal/vm/ ./internal/storage/` — all green; the new `OperationProgress*` broker tests cover publish, replay, done-clears-last, per-VM isolation, cancel-closes-channel, and the percent throttle.
- `make test-web` — couldn't run in my sandbox (Playwright wants chromium v1217, only v1194 available + outbound download blocked). Relying on CI for that.

A few non-blocking follow-ups worth folding in:

1. **De-duplicate `qemuProgressRE` and `scanLinesCR`.** They're copy-pasted verbatim into both `internal/storage/image.go:94-143` and `internal/vm/lifecycle.go:1590-1638` — two identical ~50-line blocks. A small `internal/storage/qemuprogress.go` (or similar) exporting `RunConvertWithProgress(cmd, fn)` would remove the duplication.

2. **`FilterPanel` default state.** `web/src/components/Shared.jsx:262` defaults `defaultOpen=true` with no caller override, so the panel is *expanded* on first paint — the description says "collapsed into a single Filters panel". Either flip `defaultOpen` to `false` for the list pages that have many filters, or sharpen the PR copy. Persisting per-page open/closed state in `localStorage` would also be nice since the current `useState` resets on every navigation.

3. **PR description undersells the backend.** Most of the new Go surface (`internal/api/operation_progress.go` + tests, the `GET /api/v1/vms/{vmID}/operations/progress` SSE route, `qemu-img convert -p` progress streaming, OpenAPI + schema.d.ts additions) isn't mentioned in the description. Worth a callout — reviewers shouldn't have to discover that from the diff.

Minor (no need to address): consider hoisting the ProgressBar 75/90 auto-color thresholds (`Shared.jsx:163`) into named constants since they're used in three places (Dashboard host/quota cards, VMDetail gauges, bulk bar) and are effectively a documented contract.

---

PR #409, opened 2026-06-14
Add quickstart.sh: one-command end-to-end installer with systemd

chinmay28:
## Summary

Adds a comprehensive bash installer script (`scripts/quickstart.sh`) that automates the entire VMSmith setup process on Ubuntu/Debian hosts, including dependency installation, source checkout, build, configuration, and systemd service management. The script supports non-disruptive in-place upgrades with automatic rollback on failure.

## Key Changes

- **New `scripts/quickstart.sh`**: 470-line installer that:
  - Performs preflight checks (root, systemd, OS detection, KVM availability)
  - Installs system dependencies via apt (qemu-kvm, libvirt, Go 1.22.5, Node.js 22)
  - Clones or updates VMSmith source from a configurable git ref
  - Builds the binary from source
  - Creates production data directories (`/var/lib/vmsmith`, `/etc/vmsmith`, `/var/log/vmsmith`)
  - Generates a default `config.yaml` with sensible defaults (libvirt URI, NAT network, quotas, schedules)
  - Snapshots the metadata database and previous binary before upgrade
  - Installs the binary and systemd unit file
  - Performs health checks with automatic rollback on failure
  - Prunes old backups (keeps 10 by default)

- **Configurable via environment variables**:
  - `VMSMITH_REPO`, `VMSMITH_REF`, `VMSMITH_PREFIX`, `VMSMITH_DATA_DIR`, `VMSMITH_CONFIG_DIR`, `VMSMITH_BIN_DIR`, `VMSMITH_LISTEN`
  - `INSTALL_DEPS` (auto/never), `BACKUP_KEEP`

- **Updated `README.md`**:
  - Reorganized installation instructions into three options: A (quickstart), B (binary only), C (manual build)
  - Added prominent quickstart section with curl one-liner and upgrade instructions
  - Clarified that quickstart provides managed systemd service vs. binary-only install

## Notable Implementation Details

- **Non-disruptive upgrades**: New binary is built *before* stopping the old service; DB is snapshotted with service stopped; health check (30s timeout) triggers automatic rollback if new version fails
- **Rollback mechanism**: Preserves previous binary and DB snapshots; restores both on health check failure; retains up to 10 snapshots by default
- **Local checkout detection**: If run from within an existing VMSmith repo, builds in place rather than cloning
- **Health check adaptation**: Converts wildcard listen addresses (0.0.0.0, ::) to 127.0.0.1 for localhost health checks
- **Colored output**: Terminal-aware formatting (blue steps, green success, yellow warnings, red errors)
- **Graceful dependency handling**: Detects existing Go/Node installations; skips reinstall if versions are sufficient; provides manual fallback instructions for non-Debian systems

https://claude.ai/code/session_01BveDmo3sdQ9joMzEpoNDTs

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,9 +42,41 @@ Source of truth: [`docs/ROADMAP.md`](docs/ROADMAP.md). The block above is regene
 
 Want a disposable containerized setup for local testing instead of a host install? See [docs/CONTAINER.md](docs/CONTAINER.md).
 
-### 1. Install the latest release binary
+### Option A — One-command install (managed systemd service)
 
-For Linux x86_64 hosts, you can install the latest published release directly to `/usr/local/bin/vmsmith`:
+On an Ubuntu/Debian host, the quickstart script does everything end to end: installs
+dependencies (qemu/kvm, libvirt, Go, Node), builds VMSmith from source, writes a
+production config, creates `/var/lib/vmsmith`, and runs the daemon under systemd.
+
+```bash
+curl -fsSL https://raw.githubusercontent.com/chinmay28/VMSmith/main/scripts/quickstart.sh | sudo bash
+```
+
+It defaults to the `main` branch. Pin a specific release tag (or any branch/commit)
+with `VMSMITH_REF`:
+
+```bash
+curl -fsSL https://raw.githubusercontent.com/chinmay28/VMSmith/main/scripts/quickstart.sh | sudo VMSMITH_REF=v1.0.0 bash
+```
+
+**Re-run it any time to upgrade in place.** Upgrades are non-disruptive: the new
+binary is built *before* the running service is stopped, the metadata DB is
+snapshotted first, and a failed post-upgrade health check automatically rolls back
+to the previous binary and database. VMs, images, and metadata are never touched.
+
+Manage the service afterwards:
+
+```bash
+systemctl status vmsmith
+journalctl -u vmsmith -f
+```
+
+Other knobs (all optional env vars): `VMSMITH_LISTEN`, `VMSMITH_DATA_DIR`,
+`VMSMITH_CONFIG_DIR`, `VMSMITH_PREFIX`, `INSTALL_DEPS=never`, `BACKUP_KEEP`.
+
+### Option B — Install just the release binary
+
+For Linux x86_64 hosts, you can install the latest published release directly to `/usr/local/bin/vmsmith` (no systemd setup):
 
 ```bash
 curl -fsSL https://raw.githubusercontent.com/chinmay28/VMSmith/main/scripts/install.sh | sh
@@ -56,9 +88,11 @@ Pin a specific Git tag by setting `VERSION` before piping to `sh`:
 curl -fsSL https://raw.githubusercontent.com/chinmay28/VMSmith/main/scripts/install.sh | VERSION=v1.0.0 sh
 ```
 
-If you want to build from source instead, follow the steps below.
+### Option C — Build from source manually
 
-### 2. Install dependencies
+If you want to drive each step yourself, follow the steps below.
+
+#### 1. Install dependencies
 
 ```bash
 # Ubuntu / Debian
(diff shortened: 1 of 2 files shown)


chinmay28 (commented):
**Looks good to merge** (GitHub won't let me submit a formal approve since I'm the author).

Well-structured installer. The non-disruptive upgrade sequencing is the right shape — build the new binary *before* stopping the old service, snapshot the bbolt DB (+ `-wal` / `-shm` sidecars) with the service stopped, and roll back to the previous binary + DB on a failed health check.

A few small things worth noting (none blocking):

- The systemd unit at `/etc/systemd/system/vmsmith.service` is overwritten on every run. Acceptable for an opinionated quickstart, but operators who edit the unit (e.g., to add `Environment=HTTPS_PROXY=...`) will lose customisations on upgrade. A "skip if exists" branch parallel to the existing config-file guard would be friendlier; or document the overwrite explicitly.
- `daemon.auth.enabled: false` in the generated config is correct for a localhost-only quickstart and is loudly warned about in `summary()` — good.
- `node_ok` regex accepts v22–v99 but rejects v20/v21 — matches the documented Node ≥22 requirement.
- The Go tarball download to `/tmp/goX.Y.Z.linux-amd64.tar.gz` has no checksum verification. Matches the existing `scripts/install-deps-ubuntu.sh` pattern, so consistent with the project's status quo.
- `prune_backups` uses `ls -1t ... | tail -n +N` which is technically fragile if filenames contain newlines, but generated filenames are timestamp-only — fine in practice.

Health check hits `/api/version`, which sits outside the auth gate per `CLAUDE.md` — correct probe.

Green CI.

---

chinmay28 (commented):
Nice additive script — generated config matches `internal/config/config.go`, embedded systemd unit is byte-identical to the canonical `vmsmith.service` shipped by the deb/rpm packaging, and the install/upgrade flow correctly builds before stopping the running service. Comments below are quality/correctness nits, not blockers.

**Real bugs**

- `scripts/quickstart.sh:339-340` — `[ -f X ] && cp -p X Y || true` silently swallows `cp` failures, so an upgrade can proceed past `snapshot_state` with an incomplete WAL/SHM snapshot. Prefer `if [ -f X ]; then cp -p X Y; fi` so a failed sidecar copy aborts the upgrade before the binary swap. (The mirror at L423-424 in `rollback()` is more defensible because it falls through to `rm -f`.)
- L151-153 — libvirtd `enable --now` failure is downgraded to a warning, but the health check that follows only hits `/api/version`, which doesn't touch libvirt. A fresh install on a host where libvirtd refused to start will report success and then fail at the first VM create. Suggest also probing `/api/v1/host/interfaces` or `/api/v1/host/stats` before declaring success.

**Cosmetic / consistency**

- L282 — `pid_file: "/run/vmsmith.pid"` diverges from `vmsmith.yaml.example:6` and `config.DefaultConfig()` (both use `/var/run/vmsmith.pid`). Functionally equivalent on systemd hosts but inconsistent.
- L442 — `summary()` prints `http://${HEALTH_HOST}:${PORT}`, which becomes `http://127.0.0.1:8080` even when the user explicitly listens on `[::]:8080`. Maybe surface the original `VMSMITH_LISTEN` (or the host's hostname) here so v6-only operators aren't pointed at v4.
- L264 — `chmod 0750` on `${VMSMITH_DATA_DIR}/state` is actually correct (the DB is only read by the daemon running as root, not by libvirt-qemu), but it visually contradicts the CLAUDE.md "storage dirs must be 755" guidance — worth a brief inline comment.

**Shellcheck (all `info`-severity, none breaking)**

- L173 SC2016, L209/339/340/423/424 SC2015, L348/351 SC2012, L440 SC2059 — the SC2015 at L339-340 is the only one with real-world impact (see above).

**Architecture**

- Genuinely additive vs `scripts/install.sh` (release-binary, no service), `scripts/install-deps-ubuntu.sh` (deps only), and the deb/rpm builders. Doesn't duplicate any of them.
- Data layout (`/var/lib/vmsmith/{vms,images,state}`, `/etc/vmsmith`, `/var/log/vmsmith`) matches `install-deps-ubuntu.sh` and the CLAUDE.md world-readable-storage requirement.
- Generated YAML keys map 1:1 to `config.Config` YAML tags; omitted sections fall back to `DefaultConfig()`. ✅

---

PR #390, opened 2026-06-10
feat(vms): default_user sort axis on the VM list (5.4.91)

chinmay28:
## Summary

Extends the `?sort=` whitelist on `GET /api/v1/vms` with `default_user`, the symmetric sort counterpart to the existing case-insensitive `?default_user=` exact-match filter (5.4.23) on the same column. Mirrored through `vmsmith vm list --sort default_user` and a new `<option value="default_user">Default user</option>` on the VMList page sort dropdown.

Closes the fleet-audit operator queries *"group every VM that runs as `ops-alice` together so I can audit them as a cohort"* (asc) and *"surface every VM running as root (including the empty-stored unset ones — they all run as root at runtime)"* that the existing `id` / `name` / `created_at` / `state` / `cpus` / `ram_mb` / `disk_gb` / `ip` / `image` sort axes cannot answer.

## Empty resolves to "root" (diverges from nil-trailing convention)

Unlike `ip` (5.4.85), `guest_ip` (5.4.86), `actor` (5.4.87), `image` (5.4.88), `last_fired_at` (5.4.84), and `next_fire_at` — which all push empty/null values to the tail of asc / head of desc — this column has a documented default:

- `internal/vm/lifecycle.go` runs the VM as `root` whenever `spec.default_user` is empty
- The `?default_user=root` filter contract is "empty matches root" (5.4.23)

The sort axis mirrors that semantics: empty VMs collate with explicit-root VMs in alphabetical order rather than sinking to the tail. A new unexported `resolveDefaultUser` helper at `pkg/types/vm_sort.go` locks the empty-means-root contract at one place so the filter and sort share a single source of truth.

## Changes

- `pkg/types/vm_sort.go`: `VMSortDefaultUser` const, `IsValidVMSort` extended, `resolveDefaultUser` helper, `default_user` case in `SortVMs`
- `internal/api/vm_sort.go`: 400 `invalid_sort` error message extended
- `internal/cli/vm.go`: `--sort` help text + rejection message extended
- `docs/openapi.yaml::VMSort`: enum extended, description note updated; `web/src/api/generated/schema.d.ts` regenerated
- `tests/web/mock-server.js`: `default_user` switch arm + 400 message + `vm1.spec.default_user = "ubuntu"` seed so the Playwright cohort has a meaningful split (web-server=ubuntu, db-server="", win-app="" — the latter two resolve to root and collate together)
- `web/src/pages/VMList.jsx`: new `<option value="default_user">Default user</option>` to the sort dropdown
[...]

PR #424, opened 2026-06-23
test(web): cover VM list default-user filter

march26bot:
## Summary
- add Playwright coverage for the VM list default-user filter
- create a VM with a non-empty `default_user` so the filter has a concrete target
- verify URL round-trip and case-insensitive matching for the debounced filter

## Testing
- `node --check tests/web/gui.spec.js`
- full GUI run blocked locally because the workspace does not have the frontend toolchain installed (`npm run build` fails: `vite: not found`)


chinmay28 (changes requested):
Nice idea to add Playwright coverage for the VM list `default_user` filter — closes a gap on the 5.4.23 contract. Backend CI is green and the first test (`default-user filter narrows the VM list and round-trips through the URL`) looks correct.

There's one concrete bug in the second test, plus a comment that's stale:

**Bug: `default-user filter is case-insensitive` will fail against current main.**

The test creates `ops-box` with `default_user="ubuntu"`, filters by `UBUNTU`, and then asserts that **only** `ops-box` is visible:

```js
await page.getByTestId("vm-list-default-user-filter").fill("UBUNTU");
await expect(page.getByTestId("vm-card-ops-box")).toBeVisible();
await expect(page.getByTestId("vm-card-web-server")).toHaveCount(0);  // <-- this will fail
await expect(page.getByTestId("vm-card-db-server")).toHaveCount(0);
```

But `tests/web/mock-server.js` now pins `web-server`'s `default_user` to `"ubuntu"` (from #390 / 5.4.91 — the seed is at `seed()` around line 108: `vm1.spec.default_user = "ubuntu";`). So a case-insensitive `UBUNTU` filter should match both `ops-box` AND `web-server`. The assertion `vm-card-web-server` count == 0 will fail.

The PR's `merge-base` with the commit that introduced that seed (`224cd51`) is `224cd51` itself, so this isn't a stale-branch issue — the bug is reachable on the PR's own branch the moment the mock seed is loaded. (Frontend CI on this PR is red on a number of unrelated 5.4.x tests that are baseline-flaky pre-#425, which is why this specific failure may not yet have surfaced in the CI log.)

Fix is straightforward: pick a `default_user` value that isn't already seeded — e.g. `"deploy"` (matches your first test) or `"ec2-user"`. Then the case-insensitive filter (`DEPLOY` / `EC2-USER`) will isolate `ops-box` cleanly.

**Stale comment.** The comment above the first test reads:
```js
// 5.4.x — default_user filter on the VM list. The seeded VMs have empty
// default_user values, so create one VM with a non-empty default user and
// verify the debounced text filter isolates it and round-trips via
// ?default_user=.
```

That premise is no longer true post-#390 — `web-server` is pinned to `ubuntu`. Worth updating the comment to reflect "one seeded VM (`web-server`) has `default_user="ubuntu"`; we create a second VM with a different `default_user` so the filter has a single-VM cohort". This also lets you tighten the roadmap reference from `5.4.x` to the concrete task ID (5.4.23 — the original `default_user` filter — or whichever task the GUI coverage is closing).

**Suggested follow-ups (non-blocking):**
- Rebase onto current `main` to pick up the #425 baseline fix; this should turn the rest of the GUI suite green and let your two new tests run in a stable harness.
- Optional third case: `default-user filter matches no VMs when query has no hits` — mirrors the existing `image filter matches no VMs when query has no hits` test from the previous block (line 556) and closes the empty-cohort branch.

Marking as request-changes for the assertion bug. The structure of both tests is good and the first one is ready to ship once the second is fixed.

---

chinmay28 (changes requested):
Reviewed at HEAD `8e6b255`. The test seed update (`ec2-user` instead of `ubuntu` to avoid collision with the seeded `web-server` VM whose `default_user` is `"ubuntu"`) is a good catch and the testids the spec reaches for (`vm-list-default-user-filter`, `input-vm-default-user`, `tab-advanced`, `vm-card-<name>`) all exist on `VMList.jsx`. The mock server's create-VM route also accepts `default_user` and the GET handler implements the case-insensitive `?default_user=` filter, so the wire-level logic the tests intend to verify is sound.

**However, both tests fail with `Test timeout of 30000ms exceeded` because `vm-list-default-user-filter` is not in the DOM when the test tries to fill it.** Reproduction:

```
$ PLAYWRIGHT_BROWSERS_PATH=/opt/pw-browsers npx playwright test \
    --grep "default-user filter narrows the VM list and round-trips" --reporter=list

  ✘  1 tests/web/gui.spec.js:581:3 › VM List › default-user filter narrows the VM list and round-trips through the URL (30.3s)

  1) tests/web/gui.spec.js:581:3 …
    Error: locator.fill: Test timeout of 30000ms exceeded.
    Call log:
      - waiting for getByTestId('vm-list-default-user-filter')

      594 |     await expect(page.getByTestId("vm-card-web-server")).toBeVisible();
      595 |
    > 595 |     await page.getByTestId("vm-list-default-user-filter").fill("deploy");
```

Root cause: PR #414 ("GUI: collapse filters by default…") switched `FilterPanel` to `defaultOpen = false`, and `VMList.jsx:450` wraps the default-user input in that collapsed panel:

```jsx
<FilterPanel activeCount={activeFilterCount} onClear={clearAllFilters} testId="vm-list-filters">
  …
  data-testid="vm-list-default-user-filter"
```

`FilterPanel`'s children are only mounted when `open === true`, so the input only appears after the `vm-list-filters-toggle` button is clicked. `run-gui-tests.js` already learned this lesson and ships an `openFilterPanel(page, "vm-list-filters")` helper for exactly this reason — the new Playwright spec needs the equivalent before each `.fill()` / `.click()` on a filter control.

Note: many existing `gui.spec.js` tests in the VM List / Templates / Webhooks suites suffer from the same regression after #414 (I confirmed `image filter narrows the VM list…` also times out for the same reason). The whole spec file appears to have drifted from the collapsed-panel reality, which is masked because CI only runs `run-gui-tests.js` (`make test-web`), not `npx playwright test`. So this PR isn't introducing new breakage — but it also isn't gaining the regression coverage advertised in the PR description.

Suggested fix:

```js
test("default-user filter narrows the VM list and round-trips through the URL", async ({ page }) => {
  await page.goto(BASE_URL);
  await page.getByTestId("nav-vms").click();
  // …create deploy-box…

  await page.getByTestId("vm-list-filters-toggle").click(); // <-- expand
  await page.getByTestId("vm-list-default-user-filter").fill("deploy");
  // …rest of assertions…
});
```

Apply the same toggle click to the second test (`default-user filter is case-insensitive`).

After this fix the tests should actually exercise the filter logic. Happy to re-approve once the toggle clicks are added and you've confirmed the two new tests pass with `npx playwright test --grep "default-user filter"`.

---

PR #429, opened 2026-06-23
fix: normalize template time-range filter input

march26bot:
## Summary
- normalize TemplateList datetime-local inputs to RFC3339 before calling the API
- keep the local datetime string in component state so the control stays valid and clearable
- add a focused GUI regression check for the template time-range filter

## Testing
- `node tests/web/run-gui-tests.js` *(fails in this environment due unrelated existing GUI runner issues: dashboard seeded stats/VM rows and VM image-filter test; the new template time-range regression path was added but the suite does not reach a clean pass yet)*


chinmay28 (approved):
Reviewed at HEAD `09871be`.

The bug is real: before this fix, the `onChange` handler stored `${e.target.value}:00Z` in the same state the input renders from. That format (`2026-05-10T00:00:00Z`) isn't a valid value for an `<input type="datetime-local">` element, so once any character was typed the input would reject the value and become uncontrolled / unclearable — and the `Z` suffix would mark the time as UTC even though the user picked a wall-clock local time, silently shifting the boundary.

The fix is correctly shaped:
1. **Split state.** `sinceInput` / `untilInput` hold the raw `datetime-local` string the input round-trips, so the field stays valid and clearable.
2. **`datetimeLocalToISO`** converts to RFC3339 (UTC) at API-call time via `new Date(value).toISOString()`. `new Date("2026-05-10T00:00")` interprets the string as local time, matching the convention the user sees in the picker. Invalid input is treated as empty (silent disable).
3. **Single derivation point.** `since` / `until` are now memoized derivations of `sinceInput` / `untilInput`, so every existing consumer (`useFetch` deps, URL sync, empty-state messages, "Clear filters" predicate) keeps working without changes.

State migration coverage looks complete — `clearAllFilters`, `activeFilterCount`, and the time-range clear button all flipped to `setSinceInput` / `setUntilInput` / `sinceInput` / `untilInput`. URL params still write the derived ISO string so `?since=` query params remain a stable, parseable contract for sharing links.

**Tests.** Cherry-picked onto origin/main and ran `node tests/web/run-gui-tests.js`:

- Baseline (origin/main):    `Results: 67 passed, 4 failed, 71 total`
- With this PR's change:    `Results: 68 passed, 4 failed, 72 total`

The 4 GPU-related failures are pre-existing on `main` (`gpu quota card on dashboard`, `gpu passthrough selection…`, `gpu sort axis…`, `gpu filter…`) and unrelated to this PR. The new `template time-range filter uses local datetime input and clears cleanly` regression check **passes** (delta: +1 test, +1 pass, +0 fail). The PR description's caveat about a partial local-run was a base-vs-main discrepancy — none of the listed failures are introduced by this change.

LGTM.

---

